PR #3, opened 2025-12-10
execution bugfixes

fmoletta:
Fix the following bugs:
* JALR: incorrect PC update (+offset should override usual +4)
* Store/Load: used register address instead of value at register as base
* JALR/JAL: if dst is zero, do not write next pc into it
* MultipleOpcodes: Use i32 for immediate
* Arithmetic operations with immediate: implement missing operations
* SP: set starting sp value according to current size constraints


vm/src/vm/instructions.rs
```diff
@@ -232,11 +232,15 @@ const SLTU_FUNC_IDENTIFIER: u32 = 0x3;
...
     let func3 = (instruction & FUNC3_MASK) >> 12;
-    let rs2 = (instruction & RS2_MASK) >> 20;
     let rs1 = (instruction & RS1_MASK) >> 15;
-    let mut imm = func7 | rs2;
+    let imm = ((instruction >> 20) & 0x7ff) as i32;
+    let mut imm: i32 = if (instruction & 0x8000_0000) != 0 {
+        imm - (1 << 11)
+    } else {
+        imm
+    };
```
gianbelinche: Lets add constants for those numbers

vm/src/vm/instructions.rs
```diff
@@ -354,6 +358,11 @@ fn parse_b_instruction(instruction: u32, opcode: Opcode) -> Instruction {
 fn parse_j_instruction(instruction: u32, opcode: Opcode) -> Instruction {
     let imm =
         instruction & 0xff000 | ((instruction & 0x100000) >> 9) | ((instruction >> 20) & 0x7fe);
+    let imm: i32 = if (instruction & 0x8000_0000) != 0 {
+        imm as i32 - (1 << 20)
+    } else {
+        imm as i32
+    };
```
gianbelinche: Same here

gianbelinche (approved):

MauroToscano (approved):

PR #5, opened 2025-12-11
Add workspace

gianbelinche:
This PR adds a workspace, moves the null-vm inside its own folder and adds a folder for the riscv-prover


fmoletta (approved):

MauroToscano:
Should be pointing to main

PR #22, opened 2025-12-15
feat: update crate structure

fmoletta:
Updates crate structure to the following:
- prover
- executor
- verifier


MauroToscano (changes requested):
Since the executor doesn't have GPU code, we can remove the CPU folder for the time being

gianbelinche (approved):

MauroToscano (approved):

PR #23, opened 2025-12-15
Add Jalr constraint

ColoCarletti:
- Add PC+4 constraint to JALR instruction


MauroToscano (approved):
Double check the constraint is correct

prover/src/air/constraints_templates.rs
```diff
@@ -418,3 +428,228 @@ pub fn new_sub_constraint(
...
+///
+/// Carry 0:
+/// lhs_0 = lhs[0] + 256 * lhs[1]
+/// rhs_0 = 4
+/// res_0 = res[0] + 256 * res[1]
+///
+/// carry_0 = (lhs_0 + rhs_0 - res_0) / 65536
+/// constraint: carry_0 * (carry_0 - 1) = 0
+///
+/// Carry 1:
```
diegokingston, line 433: I don't get why we are using 4 limbs when it is more efficient to use 2. Unless there is some reason to have this as some sort of template to prepare for u64

prover/src/air/constraints_templates.rs
```diff
@@ -418,3 +428,228 @@ pub fn new_sub_constraint(
...
+                let flag = self
+                    .flags_idx
+                    .iter()
+                    .fold(FieldElement::<Babybear31PrimeField>::zero(), |acc, &idx| {
+                        acc + step.get_main_evaluation_element(0, idx)
+                    });
+
+                let lhs_0 = step.get_main_evaluation_element(0, self.lhs_start_idx);
+                let rhs_0 = FieldElement::<Babybear31PrimeField>::from(4);
+                let res_0 = step.get_main_evaluation_element(0, self.res_start_idx)
```
diegokingston, line 525: I feel that at some point we could have a simpler way of defining constraints, this seems as bit convoluted

prover/src/air/constraints_templates.rs
ColoCarletti: Values are stored in the trace table as a 4-limb word, and then cast into a 2-limb word to compute the carries to be constrained.
I add some comments on https://github.com/yetanotherco/lambda_vm/pull/23/commits/e9f9ebf2b433c3278f68d24eefac86bff96e6c84

prover/src/air/constraints_templates.rs
```diff
@@ -418,3 +430,231 @@ pub fn new_sub_constraint(
...
+        self.constraint_idx
+    }
+
+    fn exemptions_period(&self) -> Option<usize> {
+        None
+    }
+
+    fn periodic_exemptions_offset(&self) -> Option<usize> {
+        None
+    }
```
MauroToscano: These are by default none, so you don't need to define them. For the purpose of the design of VM they are not useful so the default works

```suggestion
```

ColoCarletti: Solved in: https://github.com/yetanotherco/lambda_vm/pull/23/commits/5dc6cf7ed8c79c827c4044fd60292fb25cd22ead

prover/src/air/constraints_templates.rs
```diff
@@ -406,3 +416,212 @@ pub fn new_sub_constraint(
...
+                let lhs_0 = step.get_main_evaluation_element(0, self.lhs_start_idx);
+                let rhs_0 = FieldElement::<Babybear31PrimeField>::from(4);
+                let res_0 = step.get_main_evaluation_element(0, self.res_start_idx)
+                    + two_fifty_six * step.get_main_evaluation_element(0, self.res_start_idx + 1);
+
+                let one = FieldElement::<Babybear31PrimeField>::one();
+                let inverse = FieldElement::<Babybear31PrimeField>::from(INV_65536);
+                let carry_0 = (lhs_0 + rhs_0 - res_0) * inverse;
+
+                let bit_constraint: FieldElement<Babybear31PrimeField> = match self.carry_idx {
```
diegokingston, line 532: As a future optimization, the constraints for the carries should be computed simultaneously, in order to avoid recalculation of carry_0

diegokingston (approved):

prover/src/air/constraints_templates.rs
MauroToscano: Let's check this in the future

prover/src/air/constraints_templates.rs
MauroToscano: We will check this in the future

PR #24, opened 2025-12-15
feat(executor): improve logs

fmoletta:
Add the following information to `Log`:
- The current value of `pc` (aka value before executing the instruction)
- The value stored at the src1 and src2 registers (left as 0 if either is not used by the instruction)


vm/src/vm/logs.rs
```diff
@@ -1,8 +1,22 @@
 use crate::vm::instruction::decoding::Instruction;
 
 /// Log containing the executed instruction and the new value of the updated register
+/// Uses zero as default value if the ibstruction doesn't use either of src1, src2 or dst
```
MauroToscano:
```suggestion
/// Uses zero as default value if the instruction doesn't use either of src1, src2 or dst
```

MauroToscano (approved):

ColoCarletti (approved):

vm/src/vm/instruction/execution.rs
```diff
@@ -16,53 +16,73 @@ impl Instruction {
...
-            registers.0[updated_register as usize] = new_register_value;
-        }
-        Ok(Log {
-            instruction: self,
-            updated_register_value: new_register_value,
-            updated_pc: *pc,
-        })
+        let log = self.execute(*pc, registers, memory)?;
+        // Cleanup zero register in case it was written to
+        registers.0[0] = 0;
```
gianbelinche, line 22: Shouldn't we return an error here? It would be a bug

gianbelinche (approved):

vm/src/vm/instruction/execution.rs
fmoletta: Added TODO comment!

PR #50, opened 2025-12-17
Add basic CPU table.

ColoCarletti:
This PR adds the logic to populate the CPU table from the logs produced by the executor.

This is a first version where we only have a single table. Once we introduce multiple tables, this logic will need to be adapted accordingly.

For now, it runs the same Rust tests, except for the RLP test, since multiplication and division instructions are not implemented yet.


prover/src/utils.rs
```diff
@@ -0,0 +1,90 @@
...
+    [
+        x & 0xFF,
+        (x >> 8) & 0xFF,
+        (x >> 16) & 0xFF,
+        (x >> 24) & 0xFF,
+    ]
+    .map(FE::new)
+}
+
+// TODO: Revisar la logica de como guardar integers in la tabla.
```
gabrielbosio:
```suggestion
// TODO: Review logic that stores integers in the table.
```

Also, does that logic refer to an already implemented function in this repo?

gabrielbosio (dismissed):

prover/src/utils.rs
ColoCarletti: Lambdaworks does not support creating a field element directly from an integer. I checked the logic and updated this function accordingly.

prover/src/tables/cpu.rs
```diff
@@ -0,0 +1,355 @@
...
+                    }
+                    ArithOp::SetLessThanU => row.slt = FE::one(),
+                    _ => todo!(),
+                }
+            }
+
+            Instruction::JumpAndLink { dst, offset } => {
+                row.jalr = FE::one();
+                row.rd = FE::from(&dst);
+                // TODO: Check PC index = 255.
```
diegokingston: This will have to be enforced or perhaps this is already conditioned in the decoding, I have to check

prover/src/tables/cpu.rs
```diff
@@ -0,0 +1,355 @@
...
+                row.rs2 = FE::zero();
+                row.imm = u32_to_2_limbs(imm);
+                row.arg2 = u32_to_2_limbs(imm);
+                row.write_register = FE::one();
+            }
+
+            Instruction::AddUpperImmToPc { dst, imm } => {
+                row.add = FE::one();
+                row.rd = FE::from(&dst);
+                // TODO: Revisar.
```
diegokingston: review

MauroToscano (approved):

diegokingston (approved):

prover/src/tables/cpu.rs
ColoCarletti: Fix in: https://github.com/yetanotherco/lambda_vm/pull/50/commits/20499e776ee12c7e27be139d6a465c1cd5b24b67

prover/src/tables/cpu.rs
ColoCarletti: Done in: https://github.com/yetanotherco/lambda_vm/pull/50/commits/f077b9805ead6634d4bd2fe0f3d0a8287c7d9f2c

MauroToscano (approved):

PR #56, opened 2025-12-18
feat: add basic cli + bench program

fmoletta:
In order to be able to run a workflow that compares benchmarks between each PR and main (#48) we first need to have a basic cli binary we can build on the main branch. This PR takes the basic CLI and bench program from #48 so that we can fully test it out before merging


bin/cli/src/main.rs
```diff
@@ -0,0 +1,11 @@
+use executor::{elf::Elf, vm::execution::run_program};
+
+fn main() {
+    let mut args = std::env::args();
+    let elf_filename = args.nth(1).expect("No filename given");
+    let elf_data = std::fs::read(elf_filename).expect("Failed to read elf file");
+    let program = Elf::load(&elf_data).expect("Failed to load elf program");
+    let (_logs, _) =
+        run_program(program.image, program.entry_point).expect("Failed to run program");
+    // TODO: Prove program execution using logs
```
gianbelinche: Wouldn't the second result be the logs?

gianbelinche (commented):
Let's add a Readme with a basic example on how to run

bin/cli/src/main.rs
fmoletta: fixed

gianbelinche (approved):

MauroToscano (approved):

PR #66, opened 2025-12-22
feat: port `crypto` crate from lambdaworks

fmoletta:
Based on #65 
Replaces `lambdaworks-crypto` import with `crypto` crate, which is copied from the lambda works repository. Only superficial changes were made due to the newer rust version


MauroToscano (approved):
We can remove a lot of things from here later

gianbelinche (approved):

PR #70, opened 2025-12-22
feat: port `math` crate from lambdaworks

fmoletta:
Based on #66 
Replaces `lambdaworks-math` import with `math` crate, which is copied from the lambda works repository. It shall later be simplified and modified to accommodate lambda-vm's proof system needs.
The only changes were  the removal of the `winterfel-compatibility` feature which is not needed, and some minor refactors to comply with an updated clippy version


MauroToscano (approved):
We can also remove a lot of things from here later

gianbelinche (approved):

PR #83, opened 2025-12-29
spec: Fix some chip rendering pain points

RobinJadoul:
This fixes the following pain points:
- Assumptions and constraints requiring a `ref` for rendering to succeed
- The `desc` field of `arith` constraints not being rendered
- The `constraint` field of an `arith` constraint using eval in code mode
- Long tables (columns and constraints) didn't break across pages
- Template constraints did not have conditions rendered
- Constraint groups didn't get the proper prefix if specified
- The default branch of expression rendering has missing arguments

It also introduces a nice visual todo macro

### DIFF
--- a/spec/book.typ
+++ b/spec/book.typ
@@ -19,3 +19,4 @@
   *TODO #if name != none { [(#name)] }*: #body
 ]
 #let rj = todo.with(background: teal, name: "Robin")
+#let et = todo.with(background: rgb("d4aa3a"), name: "Erik")
--- a/spec/book.typ
+++ b/spec/book.typ
@@ -13,3 +13,9 @@
 // re-export page template
 #import "/templates/page.typ": project
 #let book-page = project
+
+#let todo(background: white, foreground: black, name: none, body) = block(fill: background, outset: 0.5em, radius: 20%, stroke: black)[
+  #set text(fill: foreground)
+  *TODO #if name != none { [(#name)] }*: #body
+]
+#let rj = todo.with(background: teal, name: "Robin")
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -43,9 +43,11 @@
   ), caption: [Column overview of #chip.name chip.])
 }
 
-#let cref(constraint) = {
+#let cref(constraint, body) = {
   if "ref" in constraint {
-    label(constraint.ref)
+    [#body#label(constraint.ref)]
+  } else {
+    body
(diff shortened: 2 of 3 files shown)


spec/chip.typ
```diff
@@ -43,9 +44,11 @@
...
 
-#let cref(constraint) = {
+#let cref(constraint, body) = {
   if "ref" in constraint {
-    label(constraint.ref)
+    [#body#label(constraint.ref)]
+  } else {
+    body
   }
 }
```
erik-3milabs: 1. Technically, the first argument is not a `constraint`, see line 75. Perhaps rename to `obj` or smth?
2. Could you elaborate on why you made this change?

spec/book.typ
```diff
@@ -13,3 +13,9 @@
 // re-export page template
 #import "/templates/page.typ": project
 #let book-page = project
+
+#let todo(background: white, foreground: black, name: none, body) = block(fill: background, outset: 0.5em, radius: 20%, stroke: black)[
+  #set text(fill: foreground)
+  *TODO #if name != none { [(#name)] }*: #body
+]
+#let rj = todo.with(background: teal, name: "Robin")
```
erik-3milabs, line 21:
```suggestion
#let rj = todo.with(background: teal, name: "Robin")
#let et = todo.with(background: rgb("d4aa3a"), name: "Erik")
```

spec/book.typ
erik-3milabs, line 1: > - Constraint groups didn't get the proper prefix if specified

Could you elaborate on how you achieved this here? I cannot find a trace of it anywhere.

spec/chip.typ
RobinJadoul: 1. Fair enough
2. Some things didn't want to compile when I didn't specify a `ref = `, this made it work. It was some weird error message about an empty sequence iirc

spec/book.typ
RobinJadoul: With the introduction of `lookup_group` and the corresponding call for the argument to `tag`. Before, I was trying to read the `prefix` attribute of a string.

spec/chip.typ
erik-3milabs: 2. Sorry, I meant: why did you change the function to include `body`?

RobinJadoul: This allows me to concat the label directly to the body, without the indirection of maybe returning a label

PR #84, opened 2025-12-30
spec: `SHIFT` chip

erik-3milabs:
This PR introduces the `SHIFT` chip specification.

### DIFF
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -26,27 +26,82 @@
 
 /// Generates a table listing `chip`'s columns.
 #let render_chip_column_table(chip, config) = {
+
+  // Render a definition's range
+  let render_def_range(idx, range) = {
+    if type(range) == array {
+      if range.len() == 1 {
+        [#raw(idx) `=` #range.at(0)]
+      } else if range.len() == 2 {
+        [#raw(idx) #sym.in `[`#range.at(0)`,`#range.at(1)`]`]
+      } else {
+        assert(false, message: "invalid range: " + repr(range) + repr(range.len()))
+      }
+    } else {
+      [#raw(idx) `=` #range]
+    }
+  }
+
+  // Render definition `def`
+  let render_definition(def, var_name) = {
+    if type(def) in (array, str) {
+      return (
+        [],
+        table.cell(align: right, emph[definition]), 
+        table.cell(colspan: 2, expr_to_math(def))
+      )
+    }
+
+    assert(type(def) == dictionary, message: "invalid definition: " + repr(def))
+
+    if "poly" in def {
+      (
+        [],
+        table.cell(align: right, emph[definition]), 
+        expr_to_math((":=", ("idx", var_name, def.idx), def.poly)),
+        render_def_range(def.idx, def.range)
+      )
+    } else if "polys" in def {
+      (
+        [],
+        table.cell(align: right, emph[definition]), 
+        table.cell(colspan: 2, expr_to_math(("idx", var_name, def.idx)))
+      )
+      for (i, poly) in def.polys.enumerate() {
+        (
+          [],
+          [],              
+          expr_to_math((":=", "  ", poly.poly)),
+          render_def_range(def.idx, poly.range), 
+        )
+      }
+    } else {
+      assert(false, message: "invalid definition: " + repr(def))
+    }
+  }
+
   // Group variables by category
   show figure: set block(breakable: true)
   figure(table(
-    columns: (auto, auto, 1fr),
+    columns: (auto, auto, 1fr, auto),
     inset: 6pt,
     align: left + top,
     stroke: none,
-    table.header([*Label*], [*Type*], [*Description*]),
+    table.header([*Label*], [*Type*], table.cell(colspan: 2, [*Description*])),
     table.hline(stroke: stroke(thickness: 2pt)),
     ..for (cat, vars) in chip.variables.pairs() {
-      ([#emph(cat)], [], [], table.hline(stroke: .6pt))
+      (table.cell(colspan: 4, emph(cat)), table.hline(stroke: .6pt))
       for var in vars {
-        ([#raw(var.name)], [#type_to_code(var.type)], [#eval(var.desc, mode: "markup")])
-        for (i, poly) in var.at("polys", default: ()).enumerate() {
-          (if i == 0 { emph[def] }, [], expr_to_math(("=", ("idx", var.name, i), poly)))
-        }
-        if "poly" in var {
-          (emph[def], [], expr_to_math(var.poly))
+        (
+          [#raw(var.name)], 
+          [#type_to_code(var.type)], 
+          table.cell(colspan: 2, [#eval(var.desc, mode: "markup")])
+        )
+        if "def" in var {
+          render_definition(var.def, var.name)
         }
       }
-      ([], [], [])
+      (table.cell(colspan: 4, []), )
     },
   ), caption: [Column overview of #chip.name chip.])
 }
@@ -143,7 +198,7 @@
 
   // Whether constraint has a "desc" field we need to render separately
   let has_extra_description(constraint) = {
-    constraint.kind == "arith" and "desc" in constraint
+    "desc" in constraint
   }
 
   // Rendering polynomial constraints
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -1,4 +1,4 @@
-#import "expr.typ": expr_to_code, expr_to_math
+#import "expr.typ": expr_to_code, expr_to_math, type_to_code
 
 /// Computes the total number of variables in a `chip`
 #let total_nr_variables(chip) = {
@@ -13,13 +13,21 @@
   .filter(pair => pair.at(0) in config.variables.categories.instantiated)
   .map(pair => pair.at(1))
   .flatten()
-  .map(var => config.variables.types.filter(type => type.label == var.type).at(0).subtypes.len())
+  .map(var => {
+    let (label, factor) = if type(var.type) == array {
+      (var.type.at(0), var.type.at(1))
+    } else {
+      (var.type, 1)
+    }
+    config.variables.types.filter(type => type.label == label).first().subtypes.len() * factor
+  })
   .sum()
 }
 
 /// Generates a table listing `chip`'s columns.
 #let render_chip_column_table(chip, config) = {
   // Group variables by category
+  show figure: set block(breakable: true)
   figure(table(
     columns: (auto, auto, 1fr),
     inset: 6pt,
@@ -30,7 +38,7 @@
     ..for (cat, vars) in chip.variables.pairs() {
       ([#emph(cat)], [], [], table.hline(stroke: .6pt))
       for var in vars {
-        ([#raw(var.name)], [#raw(var.type)], [#eval(var.desc, mode: "markup")])
+        ([#raw(var.name)], [#type_to_code(var.type)], [#eval(var.desc, mode: "markup")])
         for (i, poly) in var.at("polys", default: ()).enumerate() {
           (if i == 0 { emph[def] }, [], expr_to_math(("=", ("idx", var.name, i), poly)))
         }
@@ -43,9 +51,11 @@
   ), caption: [Column overview of #chip.name chip.])
 }
 
-#let cref(constraint) = {
-  if "ref" in constraint {
-    label(constraint.ref)
+#let cref(obj, body) = {
+  if "ref" in obj {
+    [#body#label(obj.ref)]
+  } else {
+    body
   }
 }
 
@@ -69,7 +79,7 @@
     let index = if "range" in assumption { "." + assumption.range.at(0) } else { "" }
     let lbl = [#chip.name\-A]
     show figure: (it) => align(left, block[#lbl#context it.counter.display()#index])
-    [#figure(kind: "assumption", numbering: (i) => [#lbl#i#index], supplement: [], [])#cref(assumption)]
+    cref(assumption)[#figure(kind: "assumption", numbering: (i) => [#lbl#i#index], supplement: [], [])]
   }
 
   figure(table(
@@ -96,13 +106,16 @@
   }
   assert(groups.all(group => group in all_groups), message: "unknown group")
 
+  // Find the group definition in the constraint_groups
+  let lookup_group(name) = chip.constraint_groups.filter((g) => g.name == name).at(0, default: (name: name))
+
   /// Render the contraint's tag.
   let tag(constraint, group) = {
     let index = if "range" in constraint { "." + constraint.range.at(0) } else { "" }
     let prefix = if "prefix" in group { group.prefix }
     let lbl = [#chip.name\-C#prefix]
     show figure: (it) => align(left, block[#lbl#context it.counter.display()#index])
-    [#figure(kind: "constraint", numbering: (i) => [#lbl#i#index], supplement: [], [])#cref(constraint)]
+    cref(constraint)[#figure(kind: "constraint", numbering: (i) => [#lbl#i#index], supplement: [], [])]
   }
 
   /// Generates a representation of `constraint`
@@ -112,17 +125,25 @@
     if kind == "interaction" {
       raw(constraint.tag) + `[` + args_interaction_like(constraint.input, constraint.at("output", default: none)) + `]`
     } else if kind == "arith" {
-      [#eval(constraint.constraint)]
+      [#eval(constraint.constraint, mode: "markup")]
     } else if kind == "template" {
-      raw(constraint.tag) + `<` + args_interaction_like(constraint.input, constraint.at("output", default: none)) + `>`
+      let cond = if "cond" in constraint {
+        $#expr_to_math(constraint.cond) arrow.r.double$ + " "
+      }
+      cond + raw(constraint.tag) + `<` + args_interaction_like(constraint.input, constraint.at("output", default: none)) + `>`
     } else {
       assert(false, message: "illegal constraint format: " + kind)
     }
   }
 
-  // Whether constraints has polynomial constraints
+  // Whether constraint has polynomial constraints
   let has_polynomial_constraints(constraint) = {
-    constraint.at("kind") == "arith" and ("poly" in constraint or "polys" in constraint)
+    constraint.kind == "arith" and ("poly" in constraint or "polys" in constraint)
+  }
+
+  // Whether constraint has a "desc" field we need to render separately
+  let has_extra_description(constraint) = {
+    constraint.kind == "arith" and "desc" in constraint
   }
 
   // Rendering polynomial constraints
@@ -139,6 +160,12 @@
     },)
   }
 
+  // Rendering the additional "desc" field for arith constraints
+  let render_extra_description(constraint) = {
+    ([_description_], [], eval(constraint.desc, mode: "markup"), [])
+  }
+
+  show figure: set block(breakable: true)
   figure(table(
     columns: (auto, auto, 1fr, auto),
     inset: 6pt,
@@ -149,11 +176,14 @@
     ..for group in groups {
       for constraint in chip.constraints.at(group) {
         (
-          [#tag(constraint, group)],
+          [#tag(constraint, lookup_group(group))],
           [#interval(constraint)],
           [#repr_constraint(constraint)],
           [#expr_to_math(constraint.at("multiplicity", default: ""))],
         )
+        if has_extra_description(constraint) {
+          render_extra_description(constraint)
+        }
         if has_polynomial_constraints(constraint) {
           render_polynomial_constraints(constraint)
         }
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -71,7 +71,7 @@
         (
           [],
           [],              
-          expr_to_math((":=", "  ", poly.poly)),
+          table.cell(inset: (left: 1.5em), expr_to_math((":=", "", poly.poly))),
           render_def_range(def.idx, poly.range), 
         )
       }
@@ -134,7 +134,7 @@
     let index = if "range" in assumption { "." + assumption.range.at(0) } else { "" }
     let lbl = [#chip.name\-A]
     show figure: (it) => align(left, block[#lbl#context it.counter.display()#index])
-    cref(assumption)[#figure(kind: "assumption", numbering: (i) => [#lbl#i#index], supplement: [], [])]
+    cref(assumption)[#figure(kind: chip.name + "assumption", numbering: (i) => [#lbl#i], supplement: [], [])]
   }
 
   figure(table(
@@ -160,6 +160,7 @@
     groups = (groups,)
   }
   assert(groups.all(group => group in all_groups), message: "unknown group")
+  let selected_constraints = groups.map(g => ((g): chip.constraints.at(g))).join()
 
   // Find the group definition in the constraint_groups
   let lookup_group(name) = chip.constraint_groups.filter((g) => g.name == name).at(0, default: (name: name))
@@ -170,7 +171,7 @@
     let prefix = if "prefix" in group { group.prefix }
     let lbl = [#chip.name\-C#prefix]
     show figure: (it) => align(left, block[#lbl#context it.counter.display()#index])
-    cref(constraint)[#figure(kind: "constraint", numbering: (i) => [#lbl#i#index], supplement: [], [])]
+    cref(constraint)[#figure(kind: chip.name + "constraint", numbering: (i) => [#lbl#i], supplement: [], [])]
   }
 
   /// Generates a representation of `constraint`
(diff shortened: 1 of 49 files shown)


spec/src/shift.toml
```diff
@@ -0,0 +1,348 @@
...
+kind = "template"
+tag = "IS_BIT"
+input = ["limb_shift_3"]
+ref = "shift:c:limb_shift_3_is_bit"
+
+[[constraints.defs]]
+kind = "template"
+tag = "IS_BIT"
+input = [["+", "limb_shift_0", "limb_shift_1", "limb_shift_2", "limb_shift_3"]]
+ref = "shift:c:limb_shift_sum_is_bit"
```
erik-3milabs: @RobinJadoul do we agree that this guy is superfluous?

RobinJadoul: I think so, though we should probably document the reasons _why_ that we discussed :)

erik-3milabs: I removed it and added an explanation in the .typ

spec/chip.typ
```diff
@@ -156,7 +156,7 @@
     }
 
     (..for poly in polys {
-      ([_polynomial constraint_], [], $#expr_to_math(poly) = 0$, [])
+      (table.cell(colspan: 2, align: right, emph("polynomial constraint")), $#expr_to_math(poly) = 0$, [])
```
erik-3milabs: this is a sneaky change that perhaps does not belong here...

RobinJadoul: True, but the resulting visual indent does look nicer

spec/src/shift.toml
```diff
@@ -0,0 +1,310 @@
...
+name = "extension"
+type = "Half"
+desc = "sign extension of `in`."
+poly = ["*", 65535, "is_negative"]
+
+[[variables.virtual]]
+name = "left"
+type = "Bit"
+desc = "Whether to perform a left-shift."
+poly = ["-", "μ", "direction"]
```
RobinJadoul: Probably worthwhile documenting this trick
(but most of the text writing for the chips is yet to come, I assume, in any case)

spec/src/shift.toml
```diff
@@ -0,0 +1,310 @@
...
+input = [["idx", "limb_shift", "i"]]
+range = ["i", 0, 4]
+ref = "shift:c:limb_shift_is_bit"
+
+[[constraint_groups]]
+name = "intra_limb_left_shifting"
+
+[[constraints.intra_limb_left_shifting]]
+kind = "interaction"
+tag = "HWSL"
```
RobinJadoul: IIRC, the reason we need separate `HWSL[C]` and `HWSR[C]` was that a right shift by `x` is equivalent to a left shift by `16 - x`, but this breaks down for `x = 0` (since we can't do a lookup for a shift by 16), right?

What if we introduce one extra column here (say `effective_bit_shift`) that we constraint to be `16 - bit_shift` when `bit_shift != 0` and `0` otherwise.
This costs one more prover column here, but saves on 2 prover (multiplicity) columns in the bitops, and some preprocessed columns too, as well as reducing the amount of somewhat.

One way to do it, would be to introduce another existing lookup (still reducing the total amount of lookups needed) as follows:
```
AND_BYTE[effective_bit_shift; 16 - bit_shift, 15]
```
Maybe there is a set of constraints that works for this, if we combine it with the fact that `effective_bit_shift` is naturally range checked by the shift lookup, but I haven't found a good one yet.

(add in multiplicity for the `AND_BYTE` and implication constraint to multiplex between `16 - bit_shift` and `bit_shift` as appropriate)

spec/src/shift.toml
```diff
@@ -0,0 +1,310 @@
...
+multiplicity = "right"
+ref = "shift:c:right_sign_extend"
+
+
+[[constraint_groups]]
+name = "limb_shifting"
+
+[[constraints.limb_shifting]]
+kind = "arith"
+constraint = "$#`limb_shift[0]` => #`shifted[i]` = #`bit_shifted[i]`$"
```
RobinJadoul: I expect that indeed, writing `shifted` as a combination of `bit_shifted[i] * limb_shift[j]` will simplify the presentation, and may even reduce the total number of constraints.
We already have degree three constraints otherwise, so nonlinearity will not cost us.
(There will probably end up being another factor for the direction involved, but we still stay in deg 3)

spec/expr.typ
```diff
@@ -129,7 +129,7 @@
   if type(typ) == array {
     check_array_type(typ)
     return raw(typ.at(0) + "[" + str(typ.at(1)) + "]")
-  } else if type(typ) == string {
+  } else if type(typ) == str {
```
RobinJadoul: This should already be on spec/main, so idk why github is showing a diff here

spec/src/shift.toml
erik-3milabs: > most of the text writing for the chips is yet to come, I assume, in any case

Indeed. I was not planning on leaving it as it is right now.

spec/src/shift.toml
erik-3milabs: > ... was that a right shift by x is equivalent to a left shift by 16 - x

To be precise: the _carry_ for `a >> x` can be computed as `a << (16-x)`, and vice versa.

> ... but this breaks down for x = 0 (since we can't do a lookup for a shift by 16), right?

and, yes, that is the problem indeed.

> What if we introduce one extra column here (say `effective_bit_shift`) that we constraint to be `16 - bit_shift` when `bit_shift != 0` and `0` otherwise.

I believe it should be `16` otherwise, since the carry should be `0` when `bit_shift = 0`.


Don't take this the wrong way: I believe you're making a very valid point. I'm just trying to make sure we both fully grasp the problem :)

spec/expr.typ
erik-3milabs: :man_shrugging:

spec/src/shift.toml
RobinJadoul: That's why I led with that question :)
I think we have a similar, and probably even nicer, situation for the uncarried version though, where the uncarried right shift follows my above reasoning, up to reordering of the limbs, which we would already tackle separately for left and right anyway

spec/src/shift.toml
erik-3milabs: You were absolutely right. This solution allows us to replace 11 constraints with this guy:

image

spec/src/shift.toml
```diff
@@ -0,0 +1,310 @@
...
+# Input
+
+[[variables.input]]
+name = "in"
+type = "DWordHL"
+desc = "The word being shifted"
+
+[[variables.input]]
+name = "shift"
+type = "BaseField"
```
RobinJadoul: Shouldn't this be `Byte` btw?
We don't need any assumptions on it, since we just need it to make the AND work, but it's maybe cleaner as documentation for what we expect.

spec/expr.typ
```diff
@@ -24,6 +24,7 @@
 // 3 : /
 // 4 : not (e.g., 5 => 1-5)
 // 5 : +
+// 5 : sum
```
RobinJadoul: Making `+` and `sum` the same precedence probably introduces ambiguity between `Σ(a + b)` and `(Σ a) + b`, which may be significant

spec/src/shift.toml
```diff
@@ -0,0 +1,276 @@
...
+ref = "shift:c:is_negative_if_unsigned"
+
+
+[[constraint_groups]]
+name = "bit_shift"
+
+[[constraints.bit_shift]]
+kind = "interaction"
+tag = "AND_BYTE"
+input = ["shift", "0x0F"]
```
RobinJadoul: Feels dangerous to me to confuse hex with a variable (nonsensically) named `0x0F`

spec/src/shift.toml
```diff
@@ -0,0 +1,276 @@
...
+ref = "shift:a:word_instr"
+
+# Constraints
+
+[[constraint_groups]]
+name = "left_flag"
+
+[[constraints.left_flag]]
+kind = "arith"
+desc = "enforces `left` is `Bit`; cheaper than `IS_BIT<left>`."
```
RobinJadoul: Not sure it's actually cheaper; looks like degree 2 in either case to me.
(but this was the original form we made the construction for `μ - x` with, so I'm still okay with the form)

spec/src/shift.toml
```diff
@@ -0,0 +1,276 @@
...
+name = "word_instr"
+type = "Bit"
+desc = "Whether this is a Word-instruction (1) or not (0)."
+
+
+# Output
+
+[[variables.output]]
+name = "shifted"
+type = "DWordHL"
```
RobinJadoul: With a bit of casting, I think we can make this a `DWordWL`

spec/src/shift.toml
```diff
@@ -0,0 +1,276 @@
...
+
+[[variables.auxiliary]]
+name = "limb_shift"
+type = ["Bit", 4]
+desc = "One-hot vector indicating whether $floor.l #`shift` / 16 floor.r equiv i mod 4$."
+
+# Virtual
+
+[[variables.virtual]]
+name = "extensions"
```
RobinJadoul:
```suggestion
name = "extension"
```

spec/src/shift.toml
```diff
@@ -0,0 +1,276 @@
...
+
+[[constraints.is_negative]]
+kind = "interaction"
+tag = "MSB16"
+input = [["idx", "in", 3]]
+output = "is_negative"
+multiplicity = "signed"
+ref = "shift:c:is_negative_if_signed"
+
+[[constraints.is_negative]]
```
RobinJadoul: I don't think this one is strictly needed, as we only use `extension` when multiplied by `right` anyway.
Keeping it and decoupling the summation of `extension * xxx` and `right * yyy` may give us some room for optimizations somewhere though (no concrete ideas where atm)

spec/shift.typ
```diff
@@ -0,0 +1,85 @@
...
+The #shift chip has the following interface:
+#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(240), 
+```
+// param in: the value being shifted
+// param shift: the number of bits to shift `in` by
+// param direction: whether to shift left (0) or right (1) 
+// param signed: whether to interpret `in` as a signed (1) or unsigned (0) integer
+// param word_instr: whether to execute the SLL/SR* (0) or SLLW/SR*W (1) instruction
+// out shifted: the resulting value
+SHIFT[shifted: DWord; in: DWord, shift: Byte, direction: Bit, signed: Bit, word_instr: Bit]
```
RobinJadoul, line 30: We should at some point discuss what the common elements are we want across chips. This feels like it's mostly repeating what's already in the input variables

spec/src/shift.toml
```diff
@@ -0,0 +1,276 @@
...
+type = ["Half", 5]
+desc = "scratch variable."
+
+[[variables.auxiliary]]
+name = "Y"
+type = ["Half", 4]
+desc = "scratch variable."
+
+[[variables.auxiliary]]
+name = "limb_shift"
```
RobinJadoul, line 66: Annoyingly, it's already used in a degree 3 constraint, otherwise, we could maybe make this into two bits (even adding a column for the output of `shift:c:limb_shift_lookup` would save one column then).

RobinJadoul (commented):
Maybe a few more things we can shave, but I think the constraints should work as they are

spec/src/shift.toml
erik-3milabs: Done

spec/src/shift.toml
erik-3milabs: I remove the second half of this comment.

spec/src/shift.toml
erik-3milabs: I replaced all occurrences of '0x0F` with just `0x0F` (which is then rendered as 15)

spec/expr.typ
erik-3milabs: Right now, it's
```
...
div = 4
sum = 5
not = 6
add = 7
...
```
wdyt?

spec/src/shift.toml
erik-3milabs: casting turned out to be rather annoying, especially because of the extension. I decided to make `shifted` virtual and assert that the new `out` would match the cast of shifted.

spec/src/shift.toml
erik-3milabs: I'm having a hard time thinking of something better here myself.

spec/src/shift.toml
erik-3milabs: I'm removing it for now.

spec/shift.typ
erik-3milabs: spun off as issue #112

spec/expr.typ
RobinJadoul: As long as the result is readable, it's fine; I _think_ my intuition would be `sum > add`, but I'm not actually entirely sure

spec/src/shift.toml
RobinJadoul: OK, I don't love having non-linear virtual columns, but at least we've saved the relevant columns :)

spec/src/shift.toml
```diff
@@ -0,0 +1,276 @@
...
+
+[[variables.auxiliary]]
+name = "is_negative"
+type = "Bit"
+desc = "Whether `in` is negative"
+
+[[variables.auxiliary]]
+name = "bit_shift"
+type = "Byte"
+desc = "Value by which to shift `in` to obtain `bit_shifted`"
```
RobinJadoul: `bit_shifted` dissolved into `X` and `Y`

spec/shift.typ
```diff
@@ -0,0 +1,147 @@
...
+== Explanation
+This chip has a rather complex design as a result of designing it to fit in as few columns possible.
+We briefly discuss the intricacies of the design, attempting to illustrate on its correctness.
+
+The chip's design revolves around a two-phase shifting process:
+1. shift `in` by $#`shift` mod 16$ bits, 
+2. shift that result by $`shift` mod 64$ (or $mod 32$ if $ #`word_instr` = 1$).
+The intermediate value representing the state between the two phases is stored in the scratch variables `X` and `Y`.
+The definition of `shifted` describes how one can combine the `X`, `Y` and `extension` variables to constructed the output value as described using `Half`-limbs.
+The output variable `out` is equivalent to `shifted`, but expressed using `Word`-limbs.
```
RobinJadoul, line 69: Maybe mention somewhere in this first part that it's logical-shift only at first.
The notation technically covers that, but it feels too subtle

spec/shift.typ
```diff
@@ -0,0 +1,147 @@
...
+== Assumptions
+#render_chip_assumptions(chip, config)
+
+== Explanation
+This chip has a rather complex design as a result of designing it to fit in as few columns possible.
+We briefly discuss the intricacies of the design, attempting to illustrate on its correctness.
+
+The chip's design revolves around a two-phase shifting process:
+1. shift `in` by $#`shift` mod 16$ bits, 
+2. shift that result by $`shift` mod 64$ (or $mod 32$ if $ #`word_instr` = 1$).
```
RobinJadoul: This isn't just `mod`, it's the `floor(shift / 16) mod {2, 4}` thing

spec/src/shift.toml
```diff
@@ -0,0 +1,276 @@
...
+
+[[variables.auxiliary]]
+name = "Y"
+type = ["Half", 4]
+desc = "scratch variable."
+
+[[variables.auxiliary]]
+name = "limb_shift"
+type = ["Bit", 4]
+desc = "One-hot vector indicating whether $floor.l #`shift` / 16 floor.r equiv i mod 4$."
```
RobinJadoul: I think it's mod 2 for `word_instr = 1`

spec/shift.typ
```diff
@@ -0,0 +1,147 @@
...
+The intermediate value representing the state between the two phases is stored in the scratch variables `X` and `Y`.
+The definition of `shifted` describes how one can combine the `X`, `Y` and `extension` variables to constructed the output value as described using `Half`-limbs.
+The output variable `out` is equivalent to `shifted`, but expressed using `Word`-limbs.
+
+=== First phase
+We zoom in on the first step.
+Here, we make use of the two lookup operations 
+- $#`HWSL[x: Half, y: B4]` := (#`x` #`<<` #`y`) mod 2^16$, and
```
RobinJadoul: Might be nice to have a mention what the names `HWSL` and `HWSLC` mean

spec/shift.typ
```diff
@@ -0,0 +1,147 @@
...
+- $#`HWSLC[x: Half, y: B4]` := #`x` #`>>` (16-#`y`)$
+Note here that one can use these two lookups to compute `out: Half[4] := in << y` as:
+$
+  #`out[`i#`]` = cases(
+    #`HWSL[in[`0#`], y]` &"if" i = 0,
+    #`HWSL[in[`i#`], y] || HWSLC[in[`i-1#`], y]` &"if" i > 0
```
RobinJadoul:
```suggestion
    #`HWSL[in[`i#`], y] | HWSLC[in[`i-1#`], y]` &"if" i > 0   
```

spec/shift.typ
```diff
@@ -0,0 +1,147 @@
...
+  )
+$
+as long as $#`y` < 16$.
+Observing that 
+$#`HWSL[x,` 16-#`y]` = (#`x` #`<<` (16-#`y`)) mod 2^16$, and
+$#`HWSLC[x,` 16-#`y]` = #`x` #`>>` #`y`$ for $#`y` in [1, 15]$,
+one can also use these lookups to compute `out := in >> y` as
+$
+  #`out[`i#`]` = cases(
+    #`HWSLC[in[`i#`],` 16-#`y] || HWSL[in[`i+1#`], y]` &"if" i < 3,
```
RobinJadoul:
```suggestion
    #`HWSLC[in[`i#`],` 16-#`y] | HWSL[in[`i+1#`], y]` &"if" i < 3,
```

spec/shift.typ
```diff
@@ -0,0 +1,147 @@
...
+    #`HWSL[in[`3#`],` 16-#`y]` &"if" i = 3
+  )
+$
+as long as $0 < #`y` < 16$.
+
+Observe now that the values being looked up are (almost) only independent from the direction of the shift: only the shift-amount varies slightly.
```
RobinJadoul:
```suggestion
Observe now that the values being looked up are (almost) independent from the direction of the shift: only the shift-amount varies slightly.
```

spec/shift.typ
```diff
@@ -0,0 +1,147 @@
...
+When we now define
+$
+  #`bit_shift` := cases(
+    #`shift` mod 16 & "when shifting left",
+    (16-#`shift`) mod 16 & "when shifting right"
+  ),  
+$
+it only takes some rearranging and combining of the values $#`X[`i#`] := HWSL[in[`i#`], bit_shift]`$ and $#`Y[`i#`] := HWSLC[in[`i#`], bit_shift]`$ to form the limbs of $#`in <</>> shift` mod 16$.
+In the exceptional case that `right = 1` and $#`shift` = 0 mod 16$, the limbs of $#`in <</>> shift` mod 16$ simply match those of `in`.
```
RobinJadoul:
```suggestion
In the remaining case that `right = 1` and $#`shift` = 0 mod 16$, the limbs of $#`in <</>> shift` mod 16$ simply match those of `in`.
```

spec/shift.typ
```diff
@@ -0,0 +1,147 @@
...
+
+=== Second phase
+Note that, since we're operating on 16-bit limbs, all the limbs in $#`in <</>> shift`$ must also occur somewhere in $#`in <</>> shift` mod 16$.
```
RobinJadoul:
```suggestion
Since we're operating on 16-bit limbs, all the limbs in $#`in <</>> shift`$ must also occur somewhere in $#`in <</>> shift` mod 16$.
```

spec/shift.typ
```diff
@@ -0,0 +1,147 @@
...
+Next, we shift the limbs of `in` left and right by the appropriate amount, storing the results in `X` and `Y` respectively.
+When `zbs = 1`, the output cannot be used to compose $#`in >>/>>> shift` mod 16$.
+To resolve this, we override `Y[i] := in[i]` and `X[i] := 0` in this case.
+
+The case of `left`-shifting and $#`bit_shift` = 0$ will be used for padding rows.
+To prevent unnecessary lookups in padding rows, we override $#`X[i]` := #`in[i]`$ and $#`Y[i]` := 0$ here.
+#render_constraint_table(chip, config, groups: "intra_limb_shift")
+
+=== Full-limb shifting
+Next, we only have to constrain that `limb_shift` is a proper unary encoding of the fifth (and sixth if `word_instr = 0`) bit of `shift`.
```
RobinJadoul: Maybe we should also cover why the weird linear combination works/is correct here

spec/test.typ
```diff
@@ -0,0 +1,24 @@
...
+#render_constraint_table(chip, config, groups: ("all", "other"))
+#render_constraint_table(chip, config, groups: ("other", "all"))
+
+// *Note*: @test:c:ref\.1 can technically be ignored, due to reflexivity. 
+
+#let x = ("a": "b")
+#let y = ("c": "d")
+#let z = (x, y)
+#let w = z.join()
+[#x, #y, #z, #w]
```
RobinJadoul: Probably to be removed

spec/src/test.toml
```diff
@@ -0,0 +1,55 @@
...
+[[constraint_groups]]
+name = "other"
+
+[[constraints.other]]
+kind = "arith"
+constraint = "$#`c[i]` => #`d[i]`$"
+poly = ["*", "c[i]", ["not", "d[i]"]]
+range=["i", 0, 1]
+multiplicity="1"
+ref = "test:c:ref"
```
RobinJadoul: To be removed?

RobinJadoul (commented):
Looks correct, makes sense and is readable with underlying knowledge of what's going on at the least

spec/shift.typ
erik-3milabs: I decided to rephrase it, though slightly differently from what you describe. Let me know if you agree with this.

RobinJadoul: Works for me :+1:

spec/shift.typ
```diff
@@ -0,0 +1,175 @@
...
+Of course, when $#`word_instr` = 1$ and, thus, only $#`shift` mod 32$ should be considered, the bit-mask for the lookup constraining `limb_shift` is adjusted appropriately (see @shift:c:limb_shift_lookup).
+
+=== Arithmetic right shift
+Lastly, we discuss the case of performing the _arithmetic_ right shift.
+Here, `extension` is constrained to contain a repetition of `in`'s most significant bit.
+Copies of this variable are used for any full limbs shifted in when $#`right` = #`signed` = 1$.
+Moreover, `X[4]` contains a copy of `extension` shifted over by the right number of bits, to allow the construction of $#`in >>> shift` mod 16$ as the appropriate intermediate.
+
+== Constraints
+First, we constraint `bit_shift` based on whether we are left or right-shifting.
```
RobinJadoul:
```suggestion
First, we constrain `bit_shift` based on whether we are left or right-shifting.
```

spec/src/shift.toml
```diff
@@ -0,0 +1,283 @@
...
+
+[[variables.auxiliary]]
+name = "Y"
+type = ["Half", 4]
+desc = "scratch variable."
+
+[[variables.auxiliary]]
+name = "limb_shift"
+type = ["Bit", 4]
+desc = "One-hot vector indicating whether $floor.l #`shift` / 16 floor.r equiv i mod s$, where $s = 2$ when $#`word_instr` = 1$ and $4$ otherwise.."
```
RobinJadoul:
```suggestion
desc = "One-hot vector indicating whether $floor.l #`shift` / 16 floor.r equiv i mod s$, where $s = 2$ when $#`word_instr` = 1$ and $4$ otherwise."
```

spec/shift.typ
```diff
@@ -0,0 +1,175 @@
...
+
+The `SHIFT` chip is comprised of #nr_variables variables that are expressed using #nr_columns columns:
+#render_chip_column_table(chip, config)
+
+== Assumptions
+#render_chip_assumptions(chip, config)
+
+== Explanation
+This chip has a rather complex design as a result of designing it to fit in as few columns possible.
+We briefly discuss the intricacies of the design, attempting to illustrate on its correctness.
```
RobinJadoul:
```suggestion
We briefly discuss the intricacies of the design, attempting to illustrate its correctness.
```

RobinJadoul (approved):
Minor language pass
Besides the possibility of removing the HWSLC bitop we're discussing, this looks ready to me

erik-3milabs:
> Besides the possibility of removing the `HWSLC` bitop we're discussing, this looks ready to me

I added issue #119 to track this.

PR #85, opened 2025-12-30
spec: support array-like types

erik-3milabs:
Support array-like variable types.

Typed as:
```toml
[[variables.auxiliary]]
name = "var"
type = ["Bit", 5]
desc = "five bits"
```

### DIFF
--- a/spec/expr.typ
+++ b/spec/expr.typ
@@ -117,3 +117,24 @@
   var: v => if v.len() == 1 { $#v$ } else { $#raw(v)$ },
   num: n => math.equation[#n],
 )
+
+// Check that a type expression is structurally valid, without validating against a set of known base types
+#let check_array_type(typ) = {
+  assert(type(typ.at(0)) == str, message: "Array types need to have a regular type as base")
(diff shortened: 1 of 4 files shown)


RobinJadoul:
Thoughts about moving `render_type` into `expr.typ` (and then probably follow a similar code/math specialization)?
That makes some sense as types are somewhat part of the expression language and its surroundings, and it paves the way to properly using it for casting.

Maybe even something like computing the size of a type would make sense to move over there.

spec/expr.typ
```diff
@@ -117,3 +117,24 @@
...
+  assert(type(typ.at(0)) == str, message: "Array types need to have a regular type as base")
+  assert(type(typ.at(1)) == int, message: "Array types need to have a constant dimension")
+}
+
+// Render a type to code
+#let type_to_code(typ) = {
+  if type(typ) == array {
+    check_array_type(typ)
+    return raw(typ.at(0) + "[" + str(typ.at(1)) + "]")
+  } else if type(typ) == string {
```
erik-3milabs, line 132: guess what: `string` does not exist 🥳

PR #87, opened 2025-12-30
spec: cast expressions

RobinJadoul:
We can now cast expressions to different types.
It is assumed that all casts allow a "natural" interpretation (such as transforming `Byte` limbs into `Halfword` limbs), and as such no formal semantics for casts are introduced.
If you want to use an ambiguous cast, prefer writing out the corresponding expression in full.


erik-3milabs (approved):
lgtm!

PR #88, opened 2025-12-31
spec: CPU chip for RV64IMC

RobinJadoul:
This currently still has two small commits to fix/improve rendering that we may want to spin out into a separate PR for cleanliness.

### DIFF
--- a/spec/cpu.typ
+++ b/spec/cpu.typ
@@ -69,7 +69,7 @@ The interactions with the wider system.
 
 === Input and output to the ALU
 
-We constraint `arg1`, `arg2` and `rvd` to correspond to the wanted values,
+We constrain `arg1`, `arg2` and `rvd` to correspond to the wanted values,
 including the appropriate sign/zero extension, depending on `word_instruction`.
 
 #render_constraint_table(chip, config, groups: "ext")
--- a/spec/cpu.typ
+++ b/spec/cpu.typ
@@ -1,0 +1,84 @@
+#import "/book.typ": book-page, rj
+#import "/src.typ": load_config, load_chip
+#import "/chip.typ": (
+  render_chip_assumptions,
+  render_chip_column_table,
+  total_nr_variables,
+  total_nr_instantiated_columns,
+  render_constraint_table,
+)
+
+#let config = load_config()
+#let chip = load_chip("src/cpu.toml", config)
+
+#show: book-page.with(title: "CPU chip")
+
+== Columns
+#let nr_variables = total_nr_variables(chip)
+#let nr_columns = total_nr_instantiated_columns(chip, config)
+
+The `CPU` chip is comprised of #nr_variables variables that are expressed using #nr_columns columns:
+#render_chip_column_table(chip, config)
+
+== Assumptions
+#render_chip_assumptions(chip, config)
+
+== Constraints
+First, we perform a decoding lookup for the current PC.
+
+#render_constraint_table(chip, config, groups: "decode")
+
+#rj[All casts for interactions will have to be reviewed once other chip interfaces stabilise]
+
+=== Range checks
+
+We constrain all columns to have the appropriate ranges.
+The flags and register indices looked up from the decoding need to be checked,
+as they are communicated through the interaction in a packed form.
+In contrast, we know ahead of time that decoding will ensure proper range checks for `pc` and `imm`.
+Similarly, since `next_pc` will propagate through the memory argument and be looked up
+in the instruction decoding on the next cycle, it is forced to be in the correct range.#rj[is this true, do we need this elsewhere for chip assumptions?]
+For the auxiliary columns, we need to check the limbs of `arg1`, `arg2`, and `res`.
+The ranges of the other auxiliary columns are enforced through later constraints.
+#rj[Make sure we argue for every column here]
+#rj[is `rvd` still sufficiently constrained? (can also be done through the memory argument like `pc`?)]
+
+#render_constraint_table(chip, config, groups: "range")
+
+=== ALU
+
+The ALU functionality is then obtained through judicious dispatching to the corresponding chips.
+
+#render_constraint_table(chip, config, groups: "alu")
+
+=== Memory
+
+The interactions with the memory, both for register loading and storing, as for `LOAD` and `STORE` instructions are handled.
+Note that since registers need no byte-addressing, we store them in the memory argument with `Word` limbs.
+The timestamps are ensured to be disjoint for disjoint memory locations.
+One consequence of that is that `next_pc` is written at `timestamp + 1`
+to ensure the access is disjoint with the `pc` read into `rv1` as part of the `AUIPC` instruction.
+
+#render_constraint_table(chip, config, groups: "mem")
+
+=== System
+
+The interactions with the wider system.
+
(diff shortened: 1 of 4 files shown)


spec/src/cpu.toml
```diff
@@ -0,0 +1,696 @@
...
+desc = "When `STORE + LOAD + BEQ + BLT = 0`, either `rs2 = 0` or `imm = 0` should be enforced by the decoding. This is needed for `arg2`."
+ref = "cpu:a:arg2-multiplex"
+
+[[constraint_groups]]
+name = "decode"
+
+[[constraints.decode]]
+kind = "interaction"
+tag = "DECODE"
+input = ["pc", "imm", "packed_decode"]
```
erik-3milabs, line 283: shouldn't packed_decode (and imm?) be an output here? (not that it matters much)

spec/src/cpu.toml
```diff
@@ -0,0 +1,696 @@
...
+name = "imm"
+type = "DWordWL"
+desc = "The fully extended 64-bit version of the immediate"
+
+[[variables.input]]
+name = "mp_selector"
+type = "Bit"
+desc = """Multi-purpose selector used by different ALU operations for different purposes. Currently, it is used
+    - by the `MUL` chip to select between `MUL`/`MULH` and `MULH[S]U`, and
+    - as flag for inverting the condition of conditional branches (see `branch_cond`)"""
```
erik-3milabs: this itemization does not mention SHIFT's `direction` (see CA36)

spec/src/cpu.toml
```diff
@@ -0,0 +1,696 @@
...
+tag = "SUB"
+cond = ["+", "SUB", "BEQ"]
+input = [["cast", "arg1", "DWordWL"], ["cast", "arg2", "DWordWL"]]
+output = ["cast", "res", "DWordWL"]
+ref = "cpu:c:sub"
+
+[[constraints.alu]]
+kind = "interaction"
+tag = "LT"
+input = [["cast", "arg1", "DWordHHW"], ["cast", "arg2", "DWordHHW"], "signed"]
```
erik-3milabs, line 496: `signed` is not listed as a variable.

spec/cpu.typ
```diff
@@ -0,0 +1,84 @@
...
+
+=== System
+
+The interactions with the wider system.
+
+#render_constraint_table(chip, config, groups: "sys")
+
+=== Input and output to the ALU
+
+We constraint `arg1`, `arg2` and `rvd` to correspond to the wanted values,
```
```suggestion
We constrain `arg1`, `arg2` and `rvd` to correspond to the wanted values,
```

spec/src/cpu.toml
```diff
@@ -0,0 +1,696 @@
...
+    - by the `MUL` chip to select between `MUL`/`MULH` and `MULH[S]U`, and
+    - as flag for inverting the condition of conditional branches (see `branch_cond`)"""
+
+[[variables.input]]
+name = "muldiv_selector"
+type = "Bit"
+desc = "Selects which output of `MUL` (lo/hi) or `DIV` (quo/rem) is wanted"
+
+[[variables.input]]
+name = "word_instruction"
```
erik-3milabs: this is a rather long name, especially when appearing in the multiplicity column. Perhaps you can find a shorter name?

spec/cpu.typ
```diff
@@ -0,0 +1,84 @@
...
+
+=== Range checks
+
+We constrain all columns to have the appropriate ranges.
+The flags and register indices looked up from the decoding need to be checked,
+as they are communicated through the interaction in a packed form.
+In contrast, we know ahead of time that decoding will ensure proper range checks for `pc` and `imm`.
+Similarly, since `next_pc` will propagate through the memory argument and be looked up
+in the instruction decoding on the next cycle, it is forced to be in the correct range.#rj[is this true, do we need this elsewhere for chip assumptions?]
+For the auxiliary columns, we need to check the limbs of `arg1`, `arg2`, and `res`.
```
erik-3milabs, line 41: I don't see these checks actually listed in the constraint tables. While I would be OK with assuming that `arg1` and `arg2` are properly extracted from memory, I would very much expect the limbs of `res` to be range checked.

erik-3milabs (commented):
Looks really solid already! I only found a couple minor points to look at.

spec/src/cpu.toml
RobinJadoul: They would indeed both be, but we don't support multiple outputs, so I figured this was likely nicer than choosing one over the other as canonical putput.

spec/cpu.typ
RobinJadoul: I'll double check. 
I would in fact not be happy with arg1 or arg2 either ;)

spec/src/cpu.toml
```diff
@@ -0,0 +1,696 @@
...
+output = ["idx", "res", "i"]
+multiplicity = "XOR"
+range = ["i", 0, 7]
+
+[[constraints.alu]]
+kind = "interaction"
+tag = "SHIFT"
+input = [["cast", "arg1", "DWordHL"], ["idx", "arg2", 0], "mp_selector", "signed"]
+output = ["cast", "res", "DWordHL"]
+multiplicity = "SHIFT"
```
erik-3milabs, line 535: > SLLW, SRLW, and SRAW are RV64I-only instructions that are analogously defined but operate on 32-bit
values and sign-extend their 32-bit results to 64 bits. The shift amount is given by rs2[4:0].

We have to discuss which component (CPU or SHIFT) is responsible for truncating `arg2` in this case.

spec/src/cpu.toml
erik-3milabs: ... or we support multiple outputs ;-)

spec/src/cpu.toml
RobinJadoul: Whoops, yep, missed that part of the truncation. 
I think the cleanest place to do this would be in the AND we have inside of SHIFT atm, extracting the top bits of the amount

spec/src/cpu.toml
RobinJadoul: Yes, but that didn't seem worth the added complexity to me, for this one usecase

spec/src/cpu.toml
RobinJadoul: Shortening to `word_instr` for now, it _could_ always still become `w_instr`

spec/src/cpu.toml
erik-3milabs: IIUC: you're proposing to add another input flag (e.g., `word_op: Bit`) to `SHIFT` and then have 
```
AND_BYTE[<lin. comb of limb_shift>; shift, 0x10 + 0x20 * (1-word_op)]
```
?

erik-3milabs:
lgtm for now.

spec/src/cpu.toml
RobinJadoul: I think that's it, yes; otherwise we need to start doing extra bit extraction things in the CPU, which would (I think) also introduce an extra column, in a table with higher utilisation

erik-3milabs: I'm on board with pulling this into SHIFT for now. Let's wait for an optimization round to see whether there are perhaps multiple chips that can be simplified when we have the CPU deal with the bit-extraction.

PR #90, opened 2025-12-31
spec: Introduce LT chip

RobinJadoul:
This should be a decent first version of the chip, ready to merge until we get to writing a bunch more things in the typst file.

### DIFF
--- a/spec/lt.typ
+++ b/spec/lt.typ
@@ -31,31 +31,34 @@ between two options, depending on the input flag `signed`.
 In the case of unsigned comparison, we simply need `unsigned_lt`, indicating
 that a wraparound (carry bit) modulo $2^64$ is needed to go from `rhs` to `lhs` via addition.
 For the case of signed comparison, we first need some case analysis.
-We can conclude that $a < b$ exactly when any of the following disjoint events happens.
 
-- $(a < 0) and (b >= 0)$
-- $(a < 0) and (b < 0) and (a < b)$
-- $(a >= 0) and (b >= 0) and (a < b)$
+We split $a < b$ into four disjoint cases, conditioned on the sign of $a$ and $b$.
+Recall that the sign of a number in two's complement can be read off from the MSB,
+being $1$ for a negative number and $0$ for a positive one.
+For this analysis, we denote the MSB of `lhs` as $A$ and the MSB of `rhs` as $B$.
 
-We represent can the comparisons of inputs to zero as the MSB or sign bits,
-which we shall denote as $A$ for `lhs` and $B$ for `rhs`.
-From this, we obtain the boolean formula $A dash(B) or A B C or dash(A) dash(B) C$,
-where we let $C$ also denote the indicator of $(a - b < 0) and (A == B)$.
-Since our cases were disjoint, this can be computed as the binary-valued polynomial
++ $dash(A) and B and (a < b)$
++ $A and dash(B) and (a < b)$
++ $A and B and (a < b)$
++ $dash(A) and dash(B) and (a < b)$
+
+The first case is evidently false, while the second case simplifies to $A and dash(B)$.
+For the third and fourth case, observe that when $A = B$, the $<$ relation is preserved
+by the modular correspondence between $[-2^(31), 2^(31))$ and $[0, 2^(64))$.
+Importantly, this modular correspondence is merely a reinterpretation of the
+bits or values of $a$ and $b$, due to the representation in two's complement.
+Hence, we can introduce the value $C = #`unsigned_lt`$, that accurately represents
+the relation $a < b$ when $A = B$.
+
+Combining our three remaining cases, we obtain the boolean formula $A dash(B) or A B C or dash(A) dash(B) C$.
+Since the cases are disjoint, this can be computed as the binary-valued polynomial
 $P(A, B, C) = A (1 - B) + A B C + (1 - A) (1 - B) C$.
 
-Observe that after modular reduction to the range $[0, 2^64)$,
-when $A == B$, the ordering $a < b$ is preserved, so $C$ can be expressed
----just as in the unsigned case--- by the overflow of the addition.
-
 The polynomial $P$ can be simplified to a total degree of two.
-We claim that the polynomial $Q(A, B, C) = A dot (1 - B) + A dot C + (1 - B) dot C$
+We claim that the polynomial $Q(A, B, C) = A (1 - B) + A C + (1 - B) C$
 is, for the purposes of this chip, equivalent to $P$.
-Through exhaustive checking, one can verify that the only binary input
-for which $P(A, B, C) != Q(A, B, C)$, is the triple $(A, B, C) = (1, 0, 1)$.
-This, however, corresponds to the case where $a > b$ in the reduction to $[0, 2^64)$,
-*and* an overflow is said to occur to go $b$ to $a$ by addition,
-which is a contradiction with the correctness of the addition.
+An exhaustive check shows that $P(A, B, C) != Q(A, B, C)$, only for the triple $(A, B, C) = (1, 0, 1)$.
+This is however impossible due to the correctness of `ADD`.
 In more detail, if we let $s$ be the (range-checked) difference $a - b$
 (so the equivalent of the #`lhs_sub_rhs` column),
 and $x'$ be the most significant word of $x$,
--- a/spec/lt.typ
+++ b/spec/lt.typ
@@ -1,0 +1,38 @@
+#import "/book.typ": book-page
+#import "/src.typ": load_config, load_chip
+#import "/chip.typ": (
+  render_chip_assumptions,
+  render_chip_column_table,
+  total_nr_variables,
+  total_nr_instantiated_columns,
+  render_constraint_table,
+)
+
+#let config = load_config()
+#let chip = load_chip("src/lt.toml", config)
+
+#show: book-page.with(title: "LT chip")
+
+== Columns
+#let nr_variables = total_nr_variables(chip)
+#let nr_columns = total_nr_instantiated_columns(chip, config)
+
+The `LT` chip is comprised of #nr_variables variables that are expressed using #nr_columns columns:
+#render_chip_column_table(chip, config)
+
+== Assumptions
+We assume the inputs `lhs`, `rhs` and `signed` are appropriately range checked.
+#render_chip_assumptions(chip, config)
+
+== Constraints
+We first constrain that all variables correspond to their definition.
+
+#render_constraint_table(chip, config, groups: "defs")
+
+And then we constrain the subtraction.
+
+#render_constraint_table(chip, config, groups: "sub")
+
+The chip contributes the following to the lookup argument.
+
+#render_constraint_table(chip, config, groups: "output")
--- a/spec/lt.typ
+++ b/spec/lt.typ
@@ -35,7 +35,8 @@ For the case of signed comparison, we first need some case analysis.
 We split $a < b$ into four disjoint cases, conditioned on the sign of $a$ and $b$.
 Recall that the sign of a number in two's complement can be read off from the MSB,
 being $1$ for a negative number and $0$ for a positive one.
-For this analysis, we denote the MSB of `lhs` as $A$ and the MSB of `rhs` as $B$.
+For this analysis, we denote the MSB of $a$ as $A$ and the MSB of $b$ as $B$.
+The four disjoint cases then become:
 
 + $dash(A) and B and (a < b)$
 + $A and dash(B) and (a < b)$
@@ -51,19 +52,19 @@ Hence, we can introduce the value $C = #`unsigned_lt`$, that accurately represen
 the relation $a < b$ when $A = B$.
 
 Combining our three remaining cases, we obtain the boolean formula $A dash(B) or A B C or dash(A) dash(B) C$.
-Since the cases are disjoint, this can be computed as the binary-valued polynomial
+Since the cases are disjoint, this can be computed with the binary-valued polynomial
 $P(A, B, C) = A (1 - B) + A B C + (1 - A) (1 - B) C$.
 
 The polynomial $P$ can be simplified to a total degree of two.
 We claim that the polynomial $Q(A, B, C) = A (1 - B) + A C + (1 - B) C$
 is, for the purposes of this chip, equivalent to $P$.
-An exhaustive check shows that $P(A, B, C) != Q(A, B, C)$, only for the triple $(A, B, C) = (1, 0, 1)$.
-This is however impossible due to the correctness of `ADD`.
+An exhaustive check shows that $P(A, B, C) != Q(A, B, C)$ only for the triple $(A, B, C) = (1, 0, 1)$.
+This is, however, impossible due to the correctness of `ADD`.
 In more detail, if we let $s$ be the (range-checked) difference $a - b$
 (so the equivalent of the #`lhs_sub_rhs` column),
-and $x'$ be the most significant word of $x$,
+and $x'$ denote the most significant word of a variable $x$,
 we need $c dot 2^32 + a' = b' + s' + #`carry[0]`$, by the definition of `carry`.
-However, the left hand side of this, is at least $3 dot 2^31$, as $(A, C) = (1, 1)$,
+However, the left hand side of this is at least $3 dot 2^31$, as $(A, C) = (1, 1)$,
 and the right hand side is at most $(2^31 - 1) + (2^32 - 1) + 1 = 3 dot 2^31 - 1$.
 Therefore, we can use $Q$ to constrain `lt` when `signed = 1`.
 
--- a/spec/lt.typ
+++ b/spec/lt.typ
@@ -26,7 +26,43 @@ We assume the inputs `lhs`, `rhs` and `signed` are appropriately range checked.
 
 == Constraints
 We first constrain that all variables correspond to their definition.
-#rj[Explain formulae properly, including sign bit logic and how overflow only matters if signs differ]
+For the defining constraint of `lt`, @lt:c:lt, observe that it is a choice
+between two options, depending on the input flag `signed`.
+In the case of unsigned comparison, we simply need `unsigned_lt`, indicating
+that a wraparound (carry bit) modulo $2^64$ is needed to go from `rhs` to `lhs` via addition.
+For the case of signed comparison, we first need some case analysis.
+We can conclude that $a < b$ exactly when any of the following disjoint events happens.
+
+- $(a < 0) and (b >= 0)$
+- $(a < 0) and (b < 0) and (a < b)$
+- $(a >= 0) and (b >= 0) and (a < b)$
+
+We represent can the comparisons of inputs to zero as the MSB or sign bits,
+which we shall denote as $A$ for `lhs` and $B$ for `rhs`.
+From this, we obtain the boolean formula $A dash(B) or A B C or dash(A) dash(B) C$,
+where we let $C$ also denote the indicator of $(a - b < 0) and (A == B)$.
+Since our cases were disjoint, this can be computed as the binary-valued polynomial
+$P(A, B, C) = A dot (1 - B) + A dot B dot C + (1 - A) dot (1 - B) dot C$.
+
+Observe that after modular reduction to the range $[0, 2^64)$,
+when $A == B$, the ordering $a < b$ is preserved, so $C$ can be expressed
+---just as in the unsigned case--- by the overflow of the addition.
+
+The polynomial $P$ can be further simplified to a total degree of two.
+We claim that the polynomial $Q(A, B, C) = A dot (1 - B) + A dot C + (1 - B) dot C$
+is, for the purposes of this chip, equivalent to $P$.
+Through exhaustive checking, one can verify that the only binary input
+for which $P(A, B, C) != Q(A, B, C)$, is the triple $(A, B, C) = (1, 0, 1)$.
+This however, corresponds to the case where $a > b$ in the reduction to $[0, 2^64)$,
+*and* an overflow is said to occur to go $b$ to $a$ by addition,
+which is a contradiction with the correctness of the addition.
+In more detail, if we let $s$ be the (range-checked) difference $a - b$
+(so the equivalent of the #`lhs_sub_rhs` column),
+and $x'$ be the most significant word of $x$,
+we need $c dot 2^32 + a' = b' + s' + #`carry[0]`$, by the definition of `carry`.
+However, the left hand side of this, is at least $3 dot 2^31$, as $(a, c) = (1, 1)$,
+and the right hand side is at most $(2^31 - 1) + (2^32 - 1) + 1 = 3 dot 2^31 - 1$.
+Therefore, we can use $Q$ to constrain `lt` when `signed = 1`.
 
 #render_constraint_table(chip, config, groups: "defs")
 
--- a/spec/lt.typ
+++ b/spec/lt.typ
@@ -1,4 +1,4 @@
-#import "/book.typ": book-page
+#import "/book.typ": book-page, rj
 #import "/src.typ": load_config, load_chip
 #import "/chip.typ": (
   render_chip_assumptions,
@@ -26,6 +26,7 @@ We assume the inputs `lhs`, `rhs` and `signed` are appropriately range checked.
 
 == Constraints
 We first constrain that all variables correspond to their definition.
+#rj[Explain formulae properly, including sign bit logic and how overflow only matters if signs differ]
 
 #render_constraint_table(chip, config, groups: "defs")
 
--- a/spec/lt.typ
+++ b/spec/lt.typ
@@ -42,25 +42,25 @@ which we shall denote as $A$ for `lhs` and $B$ for `rhs`.
 From this, we obtain the boolean formula $A dash(B) or A B C or dash(A) dash(B) C$,
 where we let $C$ also denote the indicator of $(a - b < 0) and (A == B)$.
 Since our cases were disjoint, this can be computed as the binary-valued polynomial
-$P(A, B, C) = A dot (1 - B) + A dot B dot C + (1 - A) dot (1 - B) dot C$.
+$P(A, B, C) = A (1 - B) + A B C + (1 - A) (1 - B) C$.
 
 Observe that after modular reduction to the range $[0, 2^64)$,
 when $A == B$, the ordering $a < b$ is preserved, so $C$ can be expressed
 ---just as in the unsigned case--- by the overflow of the addition.
 
-The polynomial $P$ can be further simplified to a total degree of two.
+The polynomial $P$ can be simplified to a total degree of two.
 We claim that the polynomial $Q(A, B, C) = A dot (1 - B) + A dot C + (1 - B) dot C$
 is, for the purposes of this chip, equivalent to $P$.
 Through exhaustive checking, one can verify that the only binary input
 for which $P(A, B, C) != Q(A, B, C)$, is the triple $(A, B, C) = (1, 0, 1)$.
-This however, corresponds to the case where $a > b$ in the reduction to $[0, 2^64)$,
+This, however, corresponds to the case where $a > b$ in the reduction to $[0, 2^64)$,
 *and* an overflow is said to occur to go $b$ to $a$ by addition,
 which is a contradiction with the correctness of the addition.
 In more detail, if we let $s$ be the (range-checked) difference $a - b$
 (so the equivalent of the #`lhs_sub_rhs` column),
 and $x'$ be the most significant word of $x$,
 we need $c dot 2^32 + a' = b' + s' + #`carry[0]`$, by the definition of `carry`.
-However, the left hand side of this, is at least $3 dot 2^31$, as $(a, c) = (1, 1)$,
+However, the left hand side of this, is at least $3 dot 2^31$, as $(A, C) = (1, 1)$,
 and the right hand side is at most $(2^31 - 1) + (2^32 - 1) + 1 = 3 dot 2^31 - 1$.
 Therefore, we can use $Q$ to constrain `lt` when `signed = 1`.
 
--- a/spec/src/lt.toml
+++ b/spec/src/lt.toml
@@ -1,0 +1,162 @@
+name = "LT"
+
+
+# Input
+
+[[variables.input]]
+name = "lhs"
+type = "DWordHHW"
+desc = "The left operand"
+
+[[variables.input]]
+name = "rhs"
+type = "DWordHHW"
+desc = "The right operand"
+
+[[variables.input]]
+name = "signed"
+type = "Bit"
+desc = "whether to interpret `lhs` and `rhs` as signed integers"
+
+# Output
+
+[[variables.output]]
+name = "lt"
+type = "Bit"
+desc = "Whether $#`lhs` < #`rhs`$, taking `signed` into account"
+
+
+# Auxiliary
+
+[[variables.auxiliary]]
+name = "signed_lt"
+type = "Bit"
+desc = "Whether $#`lhs` < #`rhs`$, as signed integers"
+
+[[variables.auxiliary]]
+name = "lhs_sub_rhs"
+type = "DWordHL"
+desc = "$#`lhs` - #`rhs`$"
+
+[[variables.auxiliary]]
+name = "lhs_msb"
+type = "Bit"
+desc = "The most significant bit of `lhs`"
+
+[[variables.auxiliary]]
+name = "rhs_msb"
+type = "Bit"
+desc = "The most significant bit of `rhs`"
+
+[[variables.auxiliary]]
+name = "lhs_sub_rhs_msb"
+type = "Bit"
+desc = "The most significant bit of $#`lhs` - #`rhs`$"
+
+
+# Virtual
+
+[[variables.virtual]]
+name = "carry"
+type = "Bit[3]"
+desc = "The carry of subtracting limbs of `rhs` from those of `lhs`"
+polys = [
+  ["/", ["-", ["+", ["idx", "rhs", 0], ["idx", "lhs_sub_rhs", 0]], ["idx", "lhs", 0]], ["^", 2, 32]],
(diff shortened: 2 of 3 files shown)


spec/src/lt.toml
```diff
@@ -0,0 +1,162 @@
...
+name = "lhs_sub_rhs_msb"
+type = "Bit"
+desc = "The most significant bit of $#`lhs` - #`rhs`$"
+
+
+# Virtual
+
+[[variables.virtual]]
+name = "carry"
+type = "Bit[3]"
```
erik-3milabs:
```suggestion
type = ["Bit", 3]
```

spec/src/lt.toml
```diff
@@ -0,0 +1,162 @@
...
+type = "Bit"
+desc = "Whether $#`lhs` < #`rhs`$, taking `signed` into account"
+
+
+# Auxiliary
+
+[[variables.auxiliary]]
+name = "signed_lt"
+type = "Bit"
+desc = "Whether $#`lhs` < #`rhs`$, as signed integers"
```
erik-3milabs:
```suggestion
desc = "Whether $#`lhs` < #`rhs`$, as signed integers (1) or not (0)"
```

spec/src/lt.toml
```diff
@@ -0,0 +1,162 @@
...
+
+[[variables.virtual]]
+name = "carry"
+type = "Bit[3]"
+desc = "The carry of subtracting limbs of `rhs` from those of `lhs`"
+polys = [
+  ["/", ["-", ["+", ["idx", "rhs", 0], ["idx", "lhs_sub_rhs", 0]], ["idx", "lhs", 0]], ["^", 2, 32]],
+  ["/", ["-", ["+", ["idx", "rhs", 1], ["idx", "lhs_sub_rhs", 1], ["idx", "carry", 0]], ["idx", "lhs", 1]], ["^", 2, 16]],
+  ["/", ["-", ["+", ["idx", "rhs", 2], ["idx", "lhs_sub_rhs", 2], ["idx", "carry", 1]], ["idx", "lhs", 2]], ["^", 2, 16]]
+]
```
erik-3milabs: I find the following easier to read when rendered:
```suggestion
polys = [
  ["*", ["^", 2, -32], ["-", ["+", ["idx", "rhs", 0], ["idx", "lhs_sub_rhs", 0]], ["idx", "lhs", 0]]],
  ["*", ["^", 2, -16], ["-", ["+", ["idx", "rhs", 1], ["idx", "lhs_sub_rhs", 1], ["idx", "carry", 0]], ["idx", "lhs", 1]]],
  ["*", ["^", 2, -16], ["-", ["+", ["idx", "rhs", 2], ["idx", "lhs_sub_rhs", 2], ["idx", "carry", 1]], ["idx", "lhs", 2]]]
]
```

If you disagree with this approach, perhaps we should consider using a different [rendering style](https://typst.app/docs/reference/math/frac/#parameters-style).

erik-3milabs: Also, indexing is wrong here/there is invalid type mixing: `lhs_sub_rhs` is a `DWordHL` according to its definition, while `lhs` and `rhs` are `DWordHHW`s

spec/src/lt.toml
```diff
@@ -0,0 +1,162 @@
...
+
+[[assumptions]]
+desc = "`IS_HALFWORD[lhs[i]]` and `IS_WORD[lhs[0]]`"
+range = ["i", 1, 2]
+ref = "lt:a:range_lhs"
+
+[[assumptions]]
+desc = "`IS_HALFWORD[rhs[i]]` and `IS_WORD[rhs[0]]`"
+range = ["i", 1, 2]
+ref = "lt:a:range_rhs"
```
erik-3milabs, line 80: why not unbatch these? The current notation suggests that the `IS_WORD[*hs[0]]` should be applied twice.

spec/src/lt.toml
```diff
@@ -0,0 +1,162 @@
...
+tag = "MSB16"
+input = [["idx", "lsb_sub_rhs", 2]]
+output = "lhs_sub_rhs_msb"
+multiplicity = "μ"
+ref = "lt:c:lhs_sub_rhs_msb"
+
+[[constraints.defs]]
+kind = "arith"
+constraint = "$#`lt` = #`signed` dot #`signed_lt` + (1 - #`signed`) dot #`unsigned_lt`$"
+desc = "where some shorthand means something"
```
erik-3milabs: ?
```suggestion
```

spec/src/lt.toml
```diff
@@ -0,0 +1,162 @@
...
+input = [["idx", "carry", "i"]]
+range = ["i", 0, 1]
+
+[[constraints.sub]]
+kind = "interaction"
+tag = "IS_HALFWORD"
+input = [["idx", "lhs_sub_rhs", "i"]]
+range = ["i", 0, 3]
+multiplicity = "μ"
+ref = "lt:c:lhs_sub_rhs_range"
```
erik-3milabs, line 131: This will need to be changed if you agree to change `lhs_sub_rhs`'s type to `DWordHHW`

spec/src/lt.toml
```diff
@@ -0,0 +1,162 @@
...
+
+[[variables.auxiliary]]
+name = "signed_lt"
+type = "Bit"
+desc = "Whether $#`lhs` < #`rhs`$, as signed integers"
+
+[[variables.auxiliary]]
+name = "lhs_sub_rhs"
+type = "DWordHL"
+desc = "$#`lhs` - #`rhs`$"
```
erik-3milabs, line 34:
```suggestion
[[variables.auxiliary]]
name = "lhs_sub_rhs"
type = "DWordHHW"
desc = "$#`lhs` - #`rhs`$"
```

saves a column + is required by the definition of `carry`

spec/src/lt.toml
```diff
@@ -0,0 +1,162 @@
...
+desc = "Whether $#`lhs` < #`rhs`$, as unsigned integers"
+poly = ["idx", "carry", 2]
+
+
+# Multiplicity
+
+[[variables.multiplicity]]
+name = "μ"
+type = "Bit"
+desc = ""
```
erik-3milabs, line 69: Why `μ` is a `Bit` here? Why not make it a `BaseField`?

spec/src/lt.toml
```diff
@@ -0,0 +1,162 @@
...
+
+[[constraint_groups]]
+name = "sub"
+desc = "Constrain the subtraction"
+
+[[constraints.sub]]
+kind = "template"
+tag = "IS_BIT"
+input = [["idx", "carry", "i"]]
+range = ["i", 0, 1]
```
erik-3milabs, line 123:
```suggestion
[[constraints.sub]]
kind = "template"
tag = "IS_BIT"
input = [["idx", "carry", "i"]]
range = ["i", 0, 2]
```

spec/src/lt.toml
erik-3milabs: Also, `signed_lt` is in no way constrained.

erik-3milabs (changes requested):
There are some points to look at before we can merge this.

spec/src/lt.toml
RobinJadoul: The reasoning was that they belong together conceptually, and given that assumptions are text anyway, there's no "evaluation" involved anyway, so the reader's understanding is the only thing that really matters.

spec/src/lt.toml
RobinJadoul: I think this was needed to have the range-check of the output possible.

spec/src/lt.toml
RobinJadoul: In this case, it doesn't matter much. In other chips, it may matter if we're doing things like taking `1 - μ` for alternative multiplicities.
In either case, it's probably more of a conceptual thing that we may consider an optimization of deduplicating chip rows, but at least as a base, that's not a requirement.

spec/src/lt.toml
erik-3milabs: I still see no indication that `signed_lt` is constrained in any way.

Please allow me to mark queries as resolved myself. I would have missed this, if it wasn't for my memory helping me remember about this.

spec/src/lt.toml
```diff
@@ -0,0 +1,166 @@
...
+ref = "lt:c:lhs_sub_rhs_msb"
+
+[[constraints.defs]]
+kind = "arith"
+constraint = "$#`lt` = #`signed` dot #`signed_lt` + (1 - #`signed`) dot #`unsigned_lt`$"
+poly = ["+", ["-", "lt"], ["*", "signed", "signed_lt"], ["*", ["-", 1, "signed"], "unsigned_lt"]]
+
+[[constraints.defs]]
+kind = "arith"
+constraint = "#`signed_lt` = $(1 - a) dot b + a dot b dot c + (1 - a) dot (1 - b) dot c$"
```
erik-3milabs:
```suggestion
constraint = "$#`signed_lt` = (1 - a) dot b + a dot b dot c + (1 - a) dot (1 - b) dot c$"
```

spec/src/lt.toml
```diff
@@ -0,0 +1,166 @@
...
+multiplicity = "μ"
+ref = "lt:c:rhs_msb"
+
+[[constraints.defs]]
+kind = "interaction"
+tag = "MSB16"
+input = [["idx", "lsb_sub_rhs", 2]]
+output = "lhs_sub_rhs_msb"
+multiplicity = "μ"
+ref = "lt:c:lhs_sub_rhs_msb"
```
erik-3milabs: Now that we've settled on `lhs_sub_rhs` staying an `DWordHL`, this lookup index is wrong:

```suggestion
[[constraints.defs]]
kind = "interaction"
tag = "MSB16"
input = [["idx", "lsb_sub_rhs", 3]]
output = "lhs_sub_rhs_msb"
multiplicity = "μ"
ref = "lt:c:lhs_sub_rhs_msb"
```

spec/src/lt.toml
```diff
@@ -0,0 +1,166 @@
...
+ref = "lt:a:range_lhs"
+
+[[assumptions]]
+desc = "`IS_HALFWORD[rhs[i]]` and `IS_WORD[rhs[0]]`"
+range = ["i", 1, 2]
+ref = "lt:a:range_rhs"
+
+[[assumptions]]
+desc = "`IS_BIT[signed]`"
+ref = "lt:a:range_signed"
```
erik-3milabs, line 84: Hmmmm, perhaps using <> is more fitting here?


```suggestion
[[assumptions]]
desc = "`IS_BIT<signed>`"
ref = "lt:a:range_signed"
```

spec/src/lt.toml
```diff
@@ -0,0 +1,166 @@
...
+tag = "MSB16"
+input = [["idx", "lsb_sub_rhs", 2]]
+output = "lhs_sub_rhs_msb"
+multiplicity = "μ"
+ref = "lt:c:lhs_sub_rhs_msb"
+
+[[constraints.defs]]
+kind = "arith"
+constraint = "$#`lt` = #`signed` dot #`signed_lt` + (1 - #`signed`) dot #`unsigned_lt`$"
+poly = ["+", ["-", "lt"], ["*", "signed", "signed_lt"], ["*", ["-", 1, "signed"], "unsigned_lt"]]
```
erik-3milabs: In case we want to be consistent with the one below:
```suggestion
[[constraints.defs]]
kind = "arith"
constraint = "$#`lt` = #`signed` dot #`signed_lt` + (1 - #`signed`) dot #`unsigned_lt`$"
poly = ["-", "lt", ["*", "signed", "signed_lt"], ["*", ["-", 1, "signed"], "unsigned_lt"]]
```

erik-3milabs (changes requested):
Couple more details to iron out.

spec/src/lt.toml
RobinJadoul: Should have now been addresses in [73a0b14](https://github.com/yetanotherco/lambda_vm/pull/90/commits/73a0b14bff9e4d0875710fca21f9bbf889040e3f)

spec/src/lt.toml
```diff
@@ -0,0 +1,166 @@
...
+
+[[variables.auxiliary]]
+name = "rhs_msb"
+type = "Bit"
+desc = "The most significant bit of `rhs`"
+
+[[variables.auxiliary]]
+name = "lhs_sub_rhs_msb"
+type = "Bit"
+desc = "The most significant bit of $#`lhs` - #`rhs`$"
```
erik-3milabs:
```suggestion
desc = "The most significant bit of $#`lhs_sub_rhs`$"
```

spec/src/lt.toml
```diff
@@ -0,0 +1,166 @@
...
+ref = "lt:c:lhs_sub_rhs_msb"
+
+[[constraints.defs]]
+kind = "arith"
+constraint = "$#`lt` = #`signed` dot #`signed_lt` + (1 - #`signed`) dot #`unsigned_lt`$"
+poly = ["-", "lt", ["*", "signed", "signed_lt"], ["*", ["-", 1, "signed"], "unsigned_lt"]]
+
+[[constraints.defs]]
+kind = "arith"
+constraint = "$#`signed_lt` = (1 - a) dot b + a dot b dot c + (1 - a) dot (1 - b) dot c$"
```
erik-3milabs: I find this to look much better on the rendered page.
```suggestion
constraint = "$#`signed_lt` = (1 - a) b + a b c + (1 - a) (1 - b) c$"
```
Wdyt?

erik-3milabs: Also, double checking the formula, I believe it contains a mistake. Looking at the Notion,

Image

The second item under **signed arithmetic** states that
> `sgn(a) != sgn(b)`: `a < b` iff `a` is negative and `b` is non-negative (i.e., `sgn(a) = 1` and `sgn(b) = 0`).

But down below in the formula, it reads "`A = 0` and `B=1`". I believe this should have been "`A = 1` and `B=0`". This means that the formula should actually be

```suggestion
constraint = "$#`signed_lt` = a (1 - b) + a b c + (1 - a) (1 - b) c$"
```

If you agree: don't apply the proposal, as that would not update `poly` appropriately 😅

spec/src/lt.toml
erik-3milabs, line 1: I believe we can remove one column from this chip.

Right now, `lt := signed * signed_lt + (1-signed) * unsigned_lt`, with `signed_lt` auxiliary and `unsigned_lt` virtual. The reason `signed` is not virtual, is that that would make the above definition a degree-4 constraint.

But I think we can do better:

```
MSB16[lhs_is_negative; lhs[2]] / μ = signed
!signed => lhs_is_negative = 0

MSB16[rhs_is_negative; rhs[2]] / μ = signed
!signed => rhs_is_negative = 0

MSB16[lhs_sub_rhs_is_negative; lhs_sub_rhs[3]] / μ = signed
!signed => lhs_sub_rhs_is_negative = 0

signed_lt := a(1-b) + abc + (1-a)(1-b)c
```
with
```
a := lhs_is_negative
b := rhs_is_negative
c := lhs_sub_rhs_is_negative
```

And as a result, `signed_lt` will be `=0` when `signed = 0`

This means that `lt := signed_lt + (1-signed) * unsigned_lt`, making it a degree-3 definition and as a result `signed_lt` can become virtual.

(please double check my work before implementing this 😄 )
((and also think about whether you can perhaps morph the solution slightly such that things become more readable?))

erik-3milabs (commented):
Well, I found a bug (= old mistake of mine) and a potential optimisation!

spec/src/lt.toml
RobinJadoul: - Mostly indifferent about the dots, so I'll update it
- After working through the logic again myself, I think it's indeed wrong, will update

RobinJadoul:
@erik-3milabs I've managed to get the cleanup in, along with the optimizations we discussed. Proving was a bit of a struggle, but I think it should be correct now :crossed_fingers:

spec/src/lt.toml
```diff
@@ -0,0 +1,143 @@
...
+kind = "interaction"
+tag = "MSB16"
+input = [["idx", "rhs", 2]]
+output = "rhs_msb"
+multiplicity = "μ"
+ref = "lt:c:rhs_msb"
+
+[[constraints.defs]]
+kind = "arith"
+constraint = "$#`lt` = #`signed` dot (A dot (1 - B) + A dot C + (1 - B) dot C) + (1 - #`signed`) dot #`unsigned_lt`$"
```
erik-3milabs: I would (again) be inclined to remove most of the `dot`s from this formula.

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+For the defining constraint of `lt`, @lt:c:lt, observe that it is a choice
+between two options, depending on the input flag `signed`.
+In the case of unsigned comparison, we simply need `unsigned_lt`, indicating
+that a wraparound (carry bit) modulo $2^64$ is needed to go from `rhs` to `lhs` via addition.
+For the case of signed comparison, we first need some case analysis.
+We can conclude that $a < b$ exactly when any of the following disjoint events happens.
+
+- $(a < 0) and (b >= 0)$
+- $(a < 0) and (b < 0) and (a < b)$
+- $(a >= 0) and (b >= 0) and (a < b)$
```
erik-3milabs: It is confusing to me that we're decomposing `a < b` into three events and then have `a<b` again show up in two of those events. It makes me ask wonder: how could it ever be easier to detect `a < 0 AND b <0 AND a < b` than just `a<b`? The remaining explanation does not answer this question (for me).

Actually, I think we don't check that `a < b`, but rather that `(a as u64 < b as u64)` (i.e., a <  b in the _unsigned_ case). This is a slightly different less-than operation, which makes the decomposition a lot more sense.

Perhaps we should treat the LT operator as defining an ordering on 32-bit strings (in one of two ways), and then have two distinct orderings (e.g. < (unsigned) and ≺ (signed)), such that we can they say

a ≺ b  iff
- a ≺ 0 AND b ≽ 0, or
- a ≺ 0 AND b ≺ 0 AND a < b, or
- a ≽ 0 AND b ≽ 0 AND a < b

It becomes a bit more abstract, perhaps, but I do think it would yield a much more robust explanation. Wdyt?

Also, regarding an earlier remark:

> Actually, I think we don't check that `a < b`, but rather that `(a as u64 < b as u64)` (i.e., a <  b in the _unsigned_ case)

If we do decide on the "more robust" explanation I propose above, we should consider to define `C := unsigned_lt`.

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+
+We represent can the comparisons of inputs to zero as the MSB or sign bits,
```
erik-3milabs: 1. 
```suggestion
We can represent the comparisons of inputs to zero as the MSB or sign bits,
```

2.
> We can represent the comparisons of inputs to zero as the MSB or sign bits,

The phrase "as _the_ MSB or sign bits" leaves me wandering .... "of what? How does 'the MSB or sign bits' relate to the rest of the sentence?"
I don't think swapping `the` with `their` does the trick. The more expressive "that input's [MSB or sign bit]" would do the trick IMO

```suggestion
We represent can the comparisons of inputs to zero as that input's MSB or sign bit,
```

but I feel the sentence would still be lacking detail on _how_ an input's MSB/sign bit can represent its comparison to zero (I know: 1 = negative, 0 is non-negative. But the text doesn't tell me that)

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+which we shall denote as $A$ for `lhs` and $B$ for `rhs`.
+From this, we obtain the boolean formula $A dash(B) or A B C or dash(A) dash(B) C$,
```
erik-3milabs: "From this" refers to the decomposition, which we haven't talked about for a complete sentence. I feel that this is reaching back too far.

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+where we let $C$ also denote the indicator of $(a - b < 0) and (A == B)$.
```
erik-3milabs: This does not seem like an appropriate definition for C. If it is, why not simplify `ABC + (1-A)(1-B)C` to just `C`? 

I think using the two different types of < will do a lot of good here too.

spec/lt.typ
erik-3milabs: Also, perhaps we should make this an enumeration, rather than itemization?

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+Since our cases were disjoint, this can be computed as the binary-valued polynomial
```
erik-3milabs: 1.
```suggestion
Since our cases were disjoint, this can alternatively be expressed as the binary-valued polynomial
```

2. Also, we're switching from Boolean logic to Binary polynomials. While used interchangeably, perhaps we should clarify somewhere that we use `0` to denote `false` and `1` for `true`? I agree that this may seem obvious, but all constraints take the form `... = 0` which would make it relatively easy to make the assumption that we're suddenly using `0` to represent true.

I'm not making this point because _we_ need it, but because the devs need to be very clear on all of this.

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+$P(A, B, C) = A dot (1 - B) + A dot B dot C + (1 - A) dot (1 - B) dot C$.
```
erik-3milabs: I would be inclined to remove the `dot`s here as well.

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+
+Observe that after modular reduction to the range $[0, 2^64)$,
+when $A == B$, the ordering $a < b$ is preserved, so $C$ can be expressed
+---just as in the unsigned case--- by the overflow of the addition.
```
erik-3milabs: > Observe that after modular reduction to the range $[0, 2^64)$,

Modular reduction of what? I think you are referring to the modular reduction of `a` and `b` from `[-2^31, 2^31)` to `[0, 2^64)`? That is not evident from context, imo.

2. The sentence has a weird ordering to it, making it hard to read.

3. The sentence is long. I'd propose we push "so $C$ ..." to a separate sentence.

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+
+The polynomial $P$ can be further simplified to a total degree of two.
```
erik-3milabs: I don't think we have done any "simplification" steps at this point, so "further" is not the apt here.

```suggestion
The polynomial $P$ can be simplified to a total degree of two.
```

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+We claim that the polynomial $Q(A, B, C) = A dot (1 - B) + A dot C + (1 - B) dot C$
+is, for the purposes of this chip, equivalent to $P$.
+Through exhaustive checking, one can verify that the only binary input
+for which $P(A, B, C) != Q(A, B, C)$, is the triple $(A, B, C) = (1, 0, 1)$.
```
erik-3milabs: Unnecessarily complicated, imo:

```suggestion
An exhaustive check reveals exactly binary input for which 
$P(A, B, C) != Q(A, B, C)$: the triple $(A, B, C) = (1, 0, 1)$.
```

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+This however, corresponds to the case where $a > b$ in the reduction to $[0, 2^64)$,
```
erik-3milabs:
```suggestion
This, however, corresponds to the case where $a > b$ in the reduction to $[0, 2^64)$,
```

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+*and* an overflow is said to occur to go $b$ to $a$ by addition,
```
erik-3milabs:
```suggestion
*and* an overflow is said to occur to go from $b$ to $a$ by addition,
```

The phrase "to go from b to a" feels too informal. Also, this does not address the fact that we're talking about addition of unsigned integers, even though we're discussing signed-lt.

Introducing ≺ will allow us to cleanly resolve this issue.

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+which is a contradiction with the correctness of the addition.
```
erik-3milabs:
```suggestion
which is in contradiction with the correctness of the addition.
```
or
```suggestion
which contradicts the correctness of the addition.
```

Also, the "correctness of the addition" feels... weird? Perhaps a better way of reasoning is to assume the "correctness of addition" and conclude that that input simply cannot occur?

spec/lt.typ
```diff
@@ -0,0 +1,75 @@
...
+In more detail, if we let $s$ be the (range-checked) difference $a - b$
+(so the equivalent of the #`lhs_sub_rhs` column),
+and $x'$ be the most significant word of $x$,
+we need $c dot 2^32 + a' = b' + s' + #`carry[0]`$, by the definition of `carry`.
+However, the left hand side of this, is at least $3 dot 2^31$, as $(a, c) = (1, 1)$,
```
erik-3milabs:
```suggestion
However, the left hand side of this, is at least $3 dot 2^31$, as $(A, C) = (1, 1)$,
```

erik-3milabs (commented):
I believe the explanation would significantly benefit from introducing separate symbols for signed and unsigned ordering. In one of my comments, I propose to use < and ≺, but I'm open to using other symbols if you happen to find something even better.

spec/lt.typ
RobinJadoul: The decomposition doesn't use the u64 comparison here yet.
It may seem nonsensical on the surface to define and event `A` in terms of `(A and B) or (A and not B)`, but it's a common trick to condition things and allow treating different cases differently.
In this case, we've implicitly discarded the case `(a < b) and (a >= 0) and (b < 0)` since it's always false.

The rest of the construction first deals with the first case by inspection of sign bits, and then deals with the other two cases by introducing `C`.
Looking towards another comment, indeed `C` is maybe slightly ill defined and should instead be `(a < b)` conditioned on `A == B` instead.

spec/lt.typ
RobinJadoul: See also https://github.com/yetanotherco/lambda_vm/pull/90#discussion_r2667954585

spec/lt.typ
RobinJadoul: > the devs need to be very clear on all of this.

Technically, they just need to pick the constraint out of the table, the justification is less relevant to the implementation.

spec/src/lt.toml
RobinJadoul:
```suggestion
constraint = "$#`lt` = #`signed` dot (A (1 - B) + A C + (1 - B) C) + (1 - #`signed`) dot #`unsigned_lt`$"
```

spec/lt.typ
RobinJadoul:
```suggestion
We can represent the comparison of an input to zero as the input's MSB or sign bit,
```
But I would assume that anyone reading this is expected to understand 2s complement and how a sign bit relates to the sign

spec/lt.typ
RobinJadoul: Not sure I agree with the rewording to "expressed", since we care about _computing_ the boolean formula over the field

spec/lt.typ
RobinJadoul:
```suggestion
$P(A, B, C) = A (1 - B) + A B C + (1 - A) (1 - B) C$.
```

spec/lt.typ
RobinJadoul: I think turning it around would likely complicate the argument more.
The contradiction makes the formal argument very clean, imo.

spec/lt.typ
```diff
@@ -0,0 +1,78 @@
...
+For the defining constraint of `lt`, @lt:c:lt, observe that it is a choice
+between two options, depending on the input flag `signed`.
+In the case of unsigned comparison, we simply need `unsigned_lt`, indicating
+that a wraparound (carry bit) modulo $2^64$ is needed to go from `rhs` to `lhs` via addition.
+For the case of signed comparison, we first need some case analysis.
+
+We split $a < b$ into four disjoint cases, conditioned on the sign of $a$ and $b$.
+Recall that the sign of a number in two's complement can be read off from the MSB,
+being $1$ for a negative number and $0$ for a positive one.
+For this analysis, we denote the MSB of `lhs` as $A$ and the MSB of `rhs` as $B$.
```
erik-3milabs: The enumeration comes seemingly out of nowhere. Perhaps rearranging the sentences and adding a colon suffices:

```suggestion
Let us first recall that the sign of a number in two's complement can be read off from the MSB,
being $1$ for a negative number and $0$ for a positive one.
For this analysis, we denote the MSB of `lhs` as $A$ and the MSB of `rhs` as $B$.
We split $a < b$ into four disjoint cases, conditioned on the sign of $a$ and $b$:
```

erik-3milabs: also, you're talking about 
- "the sign of `a` and `b`", 
- "MSB of `lhs` as A and the msb of `rhs` as `B`" and
- C = unsigned_lt

Either stick to `a`/`b` for an abstract discussion, or replace them with `lhs`/`rhs` for a specific discussion.

spec/lt.typ
```diff
@@ -0,0 +1,78 @@
...
+the relation $a < b$ when $A = B$.
+
+Combining our three remaining cases, we obtain the boolean formula $A dash(B) or A B C or dash(A) dash(B) C$.
+Since the cases are disjoint, this can be computed as the binary-valued polynomial
+$P(A, B, C) = A (1 - B) + A B C + (1 - A) (1 - B) C$.
+
+The polynomial $P$ can be simplified to a total degree of two.
+We claim that the polynomial $Q(A, B, C) = A (1 - B) + A C + (1 - B) C$
+is, for the purposes of this chip, equivalent to $P$.
+An exhaustive check shows that $P(A, B, C) != Q(A, B, C)$, only for the triple $(A, B, C) = (1, 0, 1)$.
```
erik-3milabs:
```suggestion
An exhaustive check shows that $P(A, B, C) != Q(A, B, C)$ only for the triple $(A, B, C) = (1, 0, 1)$.
```

spec/lt.typ
```diff
@@ -0,0 +1,78 @@
...
+This is however impossible due to the correctness of `ADD`.
```
erik-3milabs:
```suggestion
This is, however, impossible due to the correctness of `ADD`.
```

spec/lt.typ
```diff
@@ -0,0 +1,78 @@
...
+In more detail, if we let $s$ be the (range-checked) difference $a - b$
+(so the equivalent of the #`lhs_sub_rhs` column),
+and $x'$ be the most significant word of $x$,
```
erik-3milabs: "let `x'` be the ms-word of `x`" suggests that `x` is an instantiated variable.

```suggestion
and use $x'$ to denote the most significant word of variable $x$,
```

spec/lt.typ
```diff
@@ -0,0 +1,78 @@
...
+we need $c dot 2^32 + a' = b' + s' + #`carry[0]`$, by the definition of `carry`.
+However, the left hand side of this, is at least $3 dot 2^31$, as $(A, C) = (1, 1)$,
```
erik-3milabs:
```suggestion
However, the left hand side of this is at least $3 dot 2^31$, as $(A, C) = (1, 1)$,
```

spec/src/lt.toml
```diff
@@ -0,0 +1,143 @@
...
+kind = "interaction"
+tag = "MSB16"
+input = [["idx", "rhs", 2]]
+output = "rhs_msb"
+multiplicity = "μ"
+ref = "lt:c:rhs_msb"
+
+[[constraints.defs]]
+kind = "arith"
+constraint = "$#`lt` = #`signed` dot (A (1 - B) + A C + (1 - B) C) + (1 - #`signed`) dot #`unsigned_lt`$"
```
erik-3milabs, line 109: I'm just now noticing that `signed_lt` is equivalent to `lhs_msb + (1-rhs_msb) + carry_1 >= 2`.
I don't know what we can do with that, but it is a linear expression 🤔 Maybe there is a way to make `lt` virtual?

spec/src/lt.toml
```diff
@@ -0,0 +1,143 @@
...
+name = "rhs_msb"
+type = "Bit"
+desc = "The most significant bit of `rhs`"
+
+# Virtual
+
+[[variables.virtual]]
+name = "carry"
+type = ["Bit", 2]
+desc = "The carry of subtracting limbs of `rhs` from those of `lhs`"
```
erik-3milabs: Technically speaking, this is the carry from adding `lhs - rhs` to `rhs`. 
Also, it couldn't be "the carry of subtracting ..." since subtracting only knows a `borrow`.

erik-3milabs (approved):

spec/lt.typ
erik-3milabs: > The decomposition doesn't use the u64 comparison here yet.

I'm not saying it _is_. I'm just saying we _could_, and thus express the signed lt in terms of unsigned lt.

> It may seem nonsensical on the surface to define and event A in terms of (A and B) or (A and not B), but it's a common trick to condition things and allow treating different cases differently.

I understand the trick. I'm just saying that if `a<b` becomes easier to compute when you're conditioning on `a` and/or `b`, you're probably not computing `a<b` directly there, but instead taking a slightly different relation (which you can compute easily) to derive `a<b`. In our case: we're taking the unsigned-< to compute signed-<

I'm done discussing this point, however. I'm OK with how things are written now.

spec/lt.typ
erik-3milabs: suffice it to say that the phrase "not(A) and B and (a<b)" is trivially false, took me some time to understand, because I has true=positive/false=negative in my head.

spec/lt.typ
erik-3milabs: I would say that you can _express_ the boolean formula as a binary-valued polynomial and then _compute_ the value of `signed_lt` by _evaluating_ the polynomial.

"Computing something as a polynomial" feels odd to me.

spec/src/lt.toml
RobinJadoul: Hmm, interesting; the only trick I can see immediately to check `>= 2` is to do and `AND_BYTE[..., 2]`, but that still needs an instantiated `lt` column.

PR #92, opened 2026-01-02
spec: BRANCH chip

RobinJadoul:


spec/src/branch.toml
```diff
@@ -0,0 +1,111 @@
...
+# Virtual
+
+[[variables.virtual]]
+name = "add_result"
+type = "DWordWL"
+desc = "The combination of `next_pc_high` and `raw_add_low` to constrain the addition"
+polys = [
+  ["+", ["*", ["^", 2, 16], ["idx", "next_pc_high", 0]], ["*", ["^", 2, 8], ["idx", "raw_add_low", 1]], ["idx", "raw_add_low", 0]],
+  ["+", ["*", ["^", 2, 16], ["idx", "next_pc_high", 2]], ["idx", "next_pc_high", 1]],
+]
```
erik-3milabs: update `polys` to `def`

spec/src/branch.toml
erik-3milabs, line 1: I'm missing assumptions / range checks:
- `pc` and `register` being valid `DWordWL`s
- `next_pc_high` being range checked

spec/src/branch.toml
```diff
@@ -0,0 +1,111 @@
...
+name = "output"
+desc = "Each row contributes the following to the LogUp sum"
+
+[[constraints.output]]
+kind = "interaction"
+tag = "BRANCH"
+input = ["pc", "offset", "register", "JALR"]
+# TODO? Update with summation notation once it lands
+output = ["+", ["*", ["^", 2, 48], ["idx", "next_pc_high", 2]], ["*", ["^", 2, 32], ["idx", "next_pc_high", 1]], ["*", ["^", 2, 16], ["idx", "next_pc_high", 0]], ["*", ["^", 2, 8], ["idx", "raw_add_low", 1]], "masked_add_low"]
+multiplicity = "-μ"
```
erik-3milabs, line 140: this value is 64-bits, but the base field cannot store this, right?

spec/src/branch.toml
```diff
@@ -0,0 +1,111 @@
...
+
+[[variables.auxiliary]]
+name = "raw_add_low"
+type = ["Byte", 2]
+desc = "The low part of the raw addition"
+
+[[variables.auxiliary]]
+name = "masked_add_low"
+type = "Byte"
+desc = "The low part of `raw_add_low`, with it's LSB cleared"
```
erik-3milabs: it feels weird to list `next_pc_high` as output, but not `raw_add_low[1]` and `masked_add_low`.

I'd propose we list it as follows:

```
output
----------
next_pc_hi: Half[3]
next_pc_lo: Byte[2]

aux
-----
unmasked_low_byte: Byte

virtual
--- 
unmasked_next_pc: DWordWL 
    := unmasked_low_byte + 2^8 * next_pc_lo[1] + 2^16 * next_pc_high[0]
    := next_pc_high[1] + 2^16 * next_pc_high[2]
```
as well as
```
virtual
---------
next_pc: DWordWL
    := next_pc_lo[0] + 2^8 * next_pc_lo[1] + 2^16 * next_pc_hi[0]
    := next_pc_hi[1] + 2^16 * next_pc_hi[2]
```
where we can highlight the very delicate difference between `sum_res` and `next_pc`.
And lastly,
```
BRANCH[next_pc; pc, offset, register, JALR] / -μ
```
becomes much more readable.

spec/src/branch.toml
```diff
@@ -0,0 +1,111 @@
...
+input = [["idx", "raw_add_low", "i"]]
+range = ["i", 0, 1]
+multiplicity = "μ"
+
+[[constraints.all]]
+kind = "interaction"
+tag = "AND_BYTE"
+input = [["idx", "raw_add_low", 0]]
+output = "masked_add_low"
+multiplicity = "μ"
```
erik-3milabs, line 128: You're missing a second argument to `AND_BYTE` here:
```suggestion
[[constraints.all]]
kind = "interaction"
tag = "AND_BYTE"
input = [["idx", "raw_add_low", 0], "0xFE"]
output = "masked_add_low"
multiplicity = "μ"
```

spec/src/branch.toml
```diff
@@ -0,0 +1,111 @@
...
+input = ["register", ["cast", "offset", "DWordWL"]]
+output = "add_result"
+cond = "JALR"
+
+[[constraints.all]]
+kind = "interaction"
+tag = "IS_BYTE"
+input = [["idx", "raw_add_low", "i"]]
+range = ["i", 0, 1]
+multiplicity = "μ"
```
erik-3milabs, line 114: we might only have to do this for `i=1`, since `i=0` is pushed into the `AND_BYTE` lookup, which only succeeds when it is a proper byte.

spec/src/branch.toml
erik-3milabs: Perhaps `unmasked_next_pc` would be a better name than `add_result`. While `add_result` is accurate, it does not tell me much about how the variable ties into the complete computation.

spec/branch.typ
```diff
@@ -0,0 +1,31 @@
...
+
+The `BRANCH` chip is comprised of #nr_variables variables that are expressed using #nr_columns columns:
+#render_chip_column_table(chip, config)
+
+== Constraints
+
+#rj[Check correspondence with CPU for passing in `offset` as word or dword]
+We constrain `next_pc` to be `base_address + offset`,
+where `base_address` is `pc` when `JALR = 0` and `register` otherwise.
+#render_constraint_table(chip, config, groups: "all")
```
erik-3milabs: This description is a bit short. I take it this is something you wish to add once the chip designs have settle somewhat?


```suggestion
We constrain `next_pc` to be $#`base_address` + #`offset`$,
where `base_address` equals `pc` when $#`JALR` = 0$ and `register` otherwise.
#render_constraint_table(chip, config, groups: "all")
```

erik-3milabs (changes requested):

spec/branch.typ
RobinJadoul: Yeah, definitely explaining everything properly and proving/arguing they do what they should and are sound is something that will have to arrive for every chip at some point.
Currently I think working towards a baseline spec, even if not proven or even entirely correct, is a better priority to help us flesh out what all is needed on the implementation/data format side too.

spec/src/branch.toml
RobinJadoul: True, this should be as a `DWordWL`, but I'm not sure how to cleanly tuple this up. Maybe I should just introduce it as a virtual column

RobinJadoul: Ah, I should have read further down :)

spec/src/branch.toml
erik-3milabs: I see you applied the change :ok_hand: 

QQ: could you elaborate why you prefer `254` over `0xFE`? Because this is a masking operation, I would be inclined to write this in hexadecimal notation.

spec/src/branch.toml
```diff
@@ -0,0 +1,140 @@
...
+]}
+
+[[variables.virtual]]
+name = "next_pc"
+type = "DWordWL"
+desc = "The computed next pc, after masking off the LSB as required by the ISA."
+def = {idx = "i", polys = [
+  {range = [0], poly = ["+", ["*", ["^", 2, 16], ["idx", "next_pc_high", 0]], ["*", ["^", 2, 8], ["idx", "next_pc_low", 1]], ["idx", "next_pc_low", 0]]},
+  {range = [1], poly = ["+", ["*", ["^", 2, 16], ["idx", "next_pc_high", 2]], ["idx", "next_pc_high", 1]]},
+]}
```
erik-3milabs, line 66: The rendering of this variable in PDF format looks as follows:

Image

I find this slightly problematic: it suggests that the second definitions are for `unmasked_low_byte[0]` / `next_pc_low[0]` rather than `next_pc_unmasked[1]`/`next_pc[1]`.

To fix this, I propose to change line 74 in `chip.typ` from
```
expr_to_math((":=", "  ", poly.poly)),
```
to
```
table.cell(inset: (left: 1.5em), expr_to_math((":=", "", poly.poly))),
```

This would update the rendering to be

Image

which I believe to be much cleaner. 

Wdyt?

erik-3milabs (changes requested):
Solid stuff! I have just one new remark about rendering (which might perhaps best be solved outside this PR?). [This old remark](https://github.com/yetanotherco/lambda_vm/pull/92#discussion_r2657787397) is still open: regardless of how you resolve it, I am on board with this PR being merged.

erik-3milabs (approved):
woopsie, I meant to approve

spec/src/branch.toml
RobinJadoul: Because the interpretation of `"0xFE"` would be as a variable.

spec/src/branch.toml
RobinJadoul: Fine by me, also cleaner than faking indentation with a space

erik-3milabs: Can now be tracked in #98

PR #94, opened 2026-01-03
spec: conditionally render constraint table headers

erik-3milabs:
Only render the *range* and *multiplicity* headers when applicable:

Before:
image

After:
image

### DIFF
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -160,6 +160,7 @@
     groups = (groups,)
   }
   assert(groups.all(group => group in all_groups), message: "unknown group")
+  let selected_constraints = chip.constraints.pairs().filter(p => p.first() in groups).map(p => p.last()).flatten();
 
   // Find the group definition in the constraint_groups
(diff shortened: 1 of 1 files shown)


spec/chip.typ
```diff
@@ -160,6 +160,7 @@
     groups = (groups,)
   }
   assert(groups.all(group => group in all_groups), message: "unknown group")
+  let selected_constraints = chip.constraints.pairs().filter(p => p.first() in groups).map(p => p.last()).flatten();
```
RobinJadoul: Maybe we should use `selected_constraints` when rendering the actual constraints too.

erik-3milabs: I tried to do this in e131207. Not sure it improves the readability much.

The trouble is that the loop also needs the constraint's group, so just `for constraint in selected_constraints` is not immediately an option.

RobinJadoul (approved):
Indeed doesn't change very much in the end, but works for me 👍

PR #97, opened 2026-01-03
spec: `ADD` template

erik-3milabs:
Note: this is based against #95 since it also uses the `condition` variable-type.

### DIFF
--- a/spec/add.typ
+++ b/spec/add.typ
@@ -25,12 +25,12 @@
 The #add constraint template has the following interface:
 #block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => ADD<sum; lhs, rhs>"))
 where `cond` is any value described by an expression _of degree at most $1$_.
-Note that #highlighted_code("ADD<sum; lhs, rhs>") can be used to denote the _unconditional_ application of the #add template to `lhs`, `rhs`, and `sum`.
+#highlighted_code("ADD<sum; lhs, rhs>") can be used to denote the _unconditional_ application of the #add template to `lhs`, `rhs`, and `sum`.
 
 #let sub = raw("SUB")
 === #sub
 For ease of notation, we moreover introduce the #sub constraint template.
-It's interface
+Its interface
 #block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => SUB<diff; lhs, rhs>"))
 maps onto the #add template as 
 #block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => ADD<lhs; rhs, diff>"))
--- a/spec/add.typ
+++ b/spec/add.typ
@@ -27,6 +27,16 @@ The #add constraint template has the following interface:
 where `cond` is any value described by an expression _of degree at most $1$_.
 Note that #highlighted_code("ADD<sum; lhs, rhs>") can be used to denote the _unconditional_ application of the #add template to `lhs`, `rhs`, and `sum`.
 
+#let sub = raw("SUB")
+=== #sub
+For ease of notation, we moreover introduce the #sub constraint template.
+It's interface
+#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => SUB<diff; lhs, rhs>"))
+maps onto the #add template as 
+#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => ADD<lhs; rhs, diff>"))
+It constrains that $#`diff` = #`lhs` - #`rhs` mod 2^64$ when the expression `cond` is non-zero.
+As with #add, #highlighted_code("SUB<diff; lhs, rhs>") can be used to denote the _unconditional_ application of the template.
+
 == Variables
 #render_chip_column_table(chip, config)
 
--- a/spec/add.typ
+++ b/spec/add.typ
@@ -1,0 +1,38 @@
+#import "/book.typ": book-page, et
+#import "/src.typ": load_config, load_chip
+#import "/chip.typ": render_chip_column_table, render_chip_assumptions, render_constraint_table
+
+#show: book-page.with(title: "ADD/SUB")
+
+#let config = load_config()
+#let chip = load_chip("src/add.toml", config)
+
+#let add = raw(chip.name)
+
+#let highlighted_code(code) = {
+  box(
+    inset: (left: 4pt, right: 4pt), 
+    outset: (top: 4pt, bottom: 4pt), 
+    radius: 2pt,
+    fill: luma(230), 
+    raw(code))
+}
+
+= #add template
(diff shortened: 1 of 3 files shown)


spec/src/add.toml
```diff
@@ -0,0 +1,64 @@
...
+type = "DWordWL"
+desc = "$#`lhs` + #`rhs`$"
+
+[[variables.virtual]]
+name = "carry"
+desc = "Carry values used to constrain the addition"
+type = ["Bit", 4]
+def = {idx="i", polys=[
+    {range=0, poly=["*", ["^", 2, -32], ["-", ["+", ["idx", "lhs", 0], ["idx", "rhs", 0]], ["idx", "sum", 0]]]},
+    {range=[1, 3], poly=["*", ["^", 2, -32], ["-", ["+", ["idx", "lhs", "i"], ["idx", "rhs", "i"], ["idx", "carry", ["-", "i", 1]]], ["idx", "sum", "i"]]]},
```
RobinJadoul: Surely we don't have 4 word limbs?

spec/add.typ
```diff
@@ -0,0 +1,38 @@
...
+  box(
+    inset: (left: 4pt, right: 4pt), 
+    outset: (top: 4pt, bottom: 4pt), 
+    radius: 2pt,
+    fill: luma(230), 
+    raw(code))
+}
+
+= #add template
+#add is a constraint template that is used to assert that the variable `sum` is the sum of the variables `lhs` and `rhs`, under the condition that `cond` is non-zero.
```
RobinJadoul: Might clarify here that it's the sum _mod 2^64_

erik-3milabs:
@RobinJadoul would there be a point in making this template generic over the types (`Bit`, which could be interesting as XOR :thinking: ), `Byte`, `Half`, `Word`, `WordBL`, `WordHL`, `DWordBL`, `DWordHL` and/or `DWordWL`?

RobinJadoul:
I considered that, but in the end, I think we currently only ever use it for full `DWord`s, and any `DWord` representation should cast into `DWordWL` (unless we're doing something cursed like `DWordHWH` eventually), so I don't think it's worth the mental overhead.

RobinJadoul:
The XOR is maybe a valid point, but XORs are also usually integrated into some larger constraint, so that would probably not be useful in many cases

spec/add.typ
```diff
@@ -0,0 +1,38 @@
...
+    outset: (top: 4pt, bottom: 4pt), 
+    radius: 2pt,
+    fill: luma(230), 
+    raw(code))
+}
+
+= #add template
+#add is a constraint template that is used to assert that $#`sum` = #`lhs` + #`rhs` mod 2^64$, under the condition that `cond` is non-zero.
+
+== Notation
```
RobinJadoul, line 24: Should we introduce `SUB` here too, then?

erik-3milabs: done :)

spec/src/add.toml
```diff
@@ -0,0 +1,64 @@
...
+# Constraints
+
+[[constraint_groups]]
+name = "all"
+
+[[constraints.all]]
+kind = "template"
+tag = "IS_BIT"
+input = [["idx", "carry", "i"]]
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
range = ["i", 0, 1]
```

spec/src/add.toml
```diff
@@ -0,0 +1,64 @@
...
+ref = "add:a:lhs"
+
+[[assumptions]]
+desc = "`IS_WORD[rhs[i]]`"
+range = ["i", 0, 3]
+ref = "add:a:rhs"
+
+[[assumptions]]
+desc = "`IS_WORD[sum[i]]`"
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
range = ["i", 0, 1]
```

spec/src/add.toml
```diff
@@ -0,0 +1,64 @@
...
+# Assumptions
+
+[[assumptions]]
+desc = "`IS_WORD[lhs[i]]`"
+range = ["i", 0, 3]
+ref = "add:a:lhs"
+
+[[assumptions]]
+desc = "`IS_WORD[rhs[i]]`"
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
range = ["i", 0, 1]
```

spec/src/add.toml
```diff
@@ -0,0 +1,64 @@
...
+    {range=0, poly=["*", ["^", 2, -32], ["-", ["+", ["idx", "lhs", 0], ["idx", "rhs", 0]], ["idx", "sum", 0]]]},
+    {range=1, poly=["*", ["^", 2, -32], ["-", ["+", ["idx", "lhs", 1], ["idx", "rhs", 1], ["idx", "carry", 0]], ["idx", "sum", 1]]]},
+]}
+
+
+# Assumptions
+
+[[assumptions]]
+desc = "`IS_WORD[lhs[i]]`"
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
range = ["i", 0, 1]
```

spec/add.typ
```diff
@@ -0,0 +1,48 @@
...
+The #add constraint template has the following interface:
+#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => ADD<sum; lhs, rhs>"))
+where `cond` is any value described by an expression _of degree at most $1$_.
+Note that #highlighted_code("ADD<sum; lhs, rhs>") can be used to denote the _unconditional_ application of the #add template to `lhs`, `rhs`, and `sum`.
+
+#let sub = raw("SUB")
+=== #sub
+For ease of notation, we moreover introduce the #sub constraint template.
+It's interface
```
RobinJadoul:
```suggestion
Its interface
```

spec/add.typ
```diff
@@ -0,0 +1,48 @@
...
+}
+
+= #add template
+#add is a constraint template that is used to assert that $#`sum` = #`lhs` + #`rhs` mod 2^64$, under the condition that `cond` is non-zero.
+
+== Notation
+The #add constraint template has the following interface:
+#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => ADD<sum; lhs, rhs>"))
+where `cond` is any value described by an expression _of degree at most $1$_.
+Note that #highlighted_code("ADD<sum; lhs, rhs>") can be used to denote the _unconditional_ application of the #add template to `lhs`, `rhs`, and `sum`.
```
RobinJadoul:
```suggestion
#highlighted_code("ADD<sum; lhs, rhs>") can be used to denote the _unconditional_ application of the #add template to `lhs`, `rhs`, and `sum`.
```

RobinJadoul (approved):
Good to go for me, unless you want to generalize to different limb types

erik-3milabs:
> Good to go for me, unless you want to generalize to different limb types

Let's cross that bridge when we get there; generalizing over different limb types will have us introduce quite a bit more notation (especially if we want things to be machine readable)

PR #99, opened 2026-01-05
spec: do not run CI when just changing spec files

erik-3milabs:


erik-3milabs:
@RobinJadoul could you have a look at this? The CI fires cuz this PR actually touches something outside the `spec/` dir. Alternatively, we could set the CI to ignore all `spec` branches, but that feels less safe, since that would allow accidental unchecked code edits :thinking:  wdyt?

RobinJadoul (approved):
I'm not extremely familiar with gh actions, but this seems to match the documentation at least :)
Agreed that excluding the branches may be slightly less certain.

RobinJadoul:
Should this merge to spec/main or regular main though?
Maybe it's more comprehensive to bring this directly to `main` and bring it back to `spec/main` through merge or rebase?

MauroToscano (approved):

PR #102, opened 2026-01-05
Panic and stdout

gianbelinche:
This PR adds tests for writing to stdout and doing a panic.
It defines some functions on those tests involving atomics, `sys_write` and `sys_panic`
Only one syscall is added to the VM which is the one that performs the panic and ends execution


MauroToscano (changes requested):
Can you further explain the description why atomics are needed. They are not part of what we can prove

MauroToscano (approved):

PR #104, opened 2026-01-06
spec: Initial inefficient MEMW chip

RobinJadoul:
Looking at almost as many columns as the CPU, I think we'll probably want to have the fast-path MEMW chip that works for dword-aligned addresses, where the old_timestamp is also the same for all accessed elements

### DIFF
--- a/spec/memw.typ
+++ b/spec/memw.typ
@@ -46,3 +46,4 @@ This chip contributes the following to the lookup argument.
 - MEMB chip that deals does a one-byte write to remove old_timestamp from here (uncertain tradeoffs)
 - Compute `base_address[1] + 1` once and have high words of `address_add` as Words
 - Improve overflow trapping somehow so we don't need `LT` (could tie into previous one by checking carry bit of the +1)
+- Check if `old_timestamp` is already range_checked by inclusion in the memory argument
--- a/spec/memw.typ
+++ b/spec/memw.typ
@@ -1,0 +1,39 @@
+#import "/book.typ": book-page, rj
+#import "/src.typ": load_config, load_chip
+#import "/chip.typ": (
+  render_chip_assumptions,
+  render_chip_column_table,
+  total_nr_variables,
+  total_nr_instantiated_columns,
+  render_constraint_table,
+)
+
+#let config = load_config()
+#let chip = load_chip("src/memw.toml", config)
+
+#show: book-page.with(title: "MEMW chip")
+
+== Columns
+#let nr_variables = total_nr_variables(chip)
+#let nr_columns = total_nr_instantiated_columns(chip, config)
+
+The `MEMW` chip is comprised of #nr_variables variables that are expressed using #nr_columns columns:
+#render_chip_column_table(chip, config)
+
+== Assumptions
+
+#render_chip_assumptions(chip, config)
+
+== Constraints
+
+#render_constraint_table(chip, config, groups: "consistency")
+
+The chip adds the following tuples to the lookup argument,
+to effectuate that part of the memory argument.
+#render_constraint_table(chip, config, groups: "memory")
+
+This chip contributes the following to the lookup argument.
+#render_constraint_table(chip, config, groups: "output")
+
+
+
--- a/spec/memw.typ
+++ b/spec/memw.typ
@@ -28,6 +28,10 @@ The `MEMW` chip is comprised of #nr_variables variables that are expressed using
 
 #render_constraint_table(chip, config, groups: "consistency")
 
+We additionally also check that the address does not overflow
+for more significant bytes of the access.
+#render_constraint_table(chip, config, groups: "overflow")
+
 The chip adds the following tuples to the lookup argument,
 to effectuate that part of the memory argument.
 #render_constraint_table(chip, config, groups: "memory")
@@ -36,4 +40,9 @@ This chip contributes the following to the lookup argument.
 #render_constraint_table(chip, config, groups: "output")
 
 
+== Future optimization ideas
 
+- Fast path for aligned memory access where all bytes have the same old timestamp
+- MEMB chip that deals does a one-byte write to remove old_timestamp from here (uncertain tradeoffs)
+- Compute `base_address[1] + 1` once and have high words of `address_add` as Words
+- Improve overflow trapping somehow so we don't need `LT` (could tie into previous one by checking carry bit of the +1)
--- a/spec/memw.typ
+++ b/spec/memw.typ
@@ -24,6 +24,11 @@ The `MEMW` chip is comprised of #nr_variables variables that are expressed using
 
 #render_chip_assumptions(chip, config)
 
+Our assumptions do not explicitly cover any range checks for the `is_register` and `value` columns,
+as these are not necessary for the correctness of this chip in isolation.
+These properties are necessary for the consistency of the system as a whole, and therefore
+we document it here, keeping the type information as a reading help.
+
 == Constraints
 
 #render_constraint_table(chip, config, groups: "consistency")
@@ -33,7 +38,7 @@ in the memory argument automatically ensures appropriate range checking
 (as long as no external entities provide negative multiplicities without range checking the timestamp).
 This ensures the assumptions for `LT` are satisfied.
 
-We additionally also check that the address does not overflow
+We additionally check that the address does not overflow
 for more significant bytes of the access.
 #render_constraint_table(chip, config, groups: "overflow")
 
--- a/spec/memw.typ
+++ b/spec/memw.typ
@@ -28,6 +28,11 @@ The `MEMW` chip is comprised of #nr_variables variables that are expressed using
 
 #render_constraint_table(chip, config, groups: "consistency")
 
+As long as `timestamp` is properly range-checked, the presence of `old_timestamp`
+in the memory argument automatically ensures appropriate range checking
+(as long as no external entities provide negative multiplicities without range checking the timestamp).
+This ensures the assumptions for `LT` are satisfied.
+
 We additionally also check that the address does not overflow
 for more significant bytes of the access.
 #render_constraint_table(chip, config, groups: "overflow")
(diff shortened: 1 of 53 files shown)


spec/src/memw.toml
```diff
@@ -0,0 +1,233 @@
...
+
+[[variables.input]]
+name = "base_address"
+type = "DWordWL"
+desc = "The base address to read/write from/to, gets offset by $[0, 7]$, depending on how big the access is"
+
+[[variables.input]]
+name = "value"
+type = ["BaseField", 8]
+desc = "The values to store in memory. For regular memory, these should be (up to) 8 range-checked `Byte`s; registers are stored as two range-checked `Word`s"
```
erik-3milabs, line 18: There is no assumption/constraint enforcing that `is_register => write2 & !write4`

spec/src/memw.toml
```diff
@@ -0,0 +1,233 @@
...
+# TODO: does it make sense to add writeX multiplicities?
+[[constraints.consistency]]
+kind = "template"
+tag = "ADD"
+input = ["base_address", ["+", "i", 1]]
+output = ["idx", "address_add", "i"]
+range = ["i", 0, 6]
+
+# TODO: whoops, double loop
+# TODO: does it make sense to add writeX multiplicities?
```
erik-3milabs: why would it not?

spec/src/memw.toml
```diff
@@ -0,0 +1,233 @@
+name = "MEMW"
+
+# Input
+
+[[variables.input]]
+name = "is_register"
+type = "Bit"
+desc = "Whether the address represents a register index"
```
erik-3milabs, line 8: I'm missing an assumption/constraint on `is_register`'s range.

spec/src/memw.toml
```diff
@@ -0,0 +1,233 @@
...
+
+[[variables.multiplicity]]
+name = "μ_write"
+type = "Bit"
+desc = "Whether we are performing a write (and hence not return `out`)"
+
+
+[[assumptions]]
+desc = "`IS_WORD[base_address[i]]`"
+range = ["i", 0, 7]
```
erik-3milabs: I believe `base_address: DWordWL` with only two limbs:
```suggestion
[[assumptions]]
desc = "`IS_WORD[base_address[i]]`"
range = ["i", 0, 1]
```

spec/src/memw.toml
erik-3milabs, line 18: I'm missing an assumption that `value` is range-checked by the caller.

spec/src/memw.toml
```diff
@@ -0,0 +1,233 @@
...
+
+[[variables.input]]
+name = "value"
+type = ["BaseField", 8]
+desc = "The values to store in memory. For regular memory, these should be (up to) 8 range-checked `Byte`s; registers are stored as two range-checked `Word`s"
+
+[[variables.input]]
+name = "timestamp"
+type = "DWordWL"
+desc = "The timestamp at which this memory access is said to occur"
```
erik-3milabs, line 23: I'm missing an assumption on `timestamp`'s range (or did you ignore it because it is precomputed?)

spec/src/memw.toml
```diff
@@ -0,0 +1,233 @@
...
+type = ["BaseField", 8]
+desc = """The old value written at `base_address`. See `value` for information about representation.
+Only the elements corresponding to the `writeN` bits are guaranteed"""
+
+# Auxiliary
+
+[[variables.auxiliary]]
+name = "address_add"
+type = ["DWordHL", 7]
+desc = "`base_address + i + 1`"
```
erik-3milabs: Perhaps we should not implicitly assume `i` as the index.
```suggestion
desc = "`address_add[i] = base_address + i + 1`"
```

spec/src/memw.toml
```diff
@@ -0,0 +1,233 @@
...
+
+[[variables.auxiliary]]
+name = "old_timestamp"
+type = ["DWordWL", 8]
+desc = "The timestamp at which the address was last accessed"
```
erik-3milabs, line 58: I take it `old_timestamp` is range-constrained through the `memory` look-ups?

spec/src/memw.toml
```diff
@@ -0,0 +1,233 @@
...
+kind = "template"
+tag = "IS_BIT"
+input = ["μ_sum"]
+
+[[constraints.consistency]]
+kind = "arith"
+constraint = "$#`write2` => #`μ_sum`$"
+poly = ["*", "write2", ["not", "μ_sum"]]
+
+# TODO: does it make sense to add writeX multiplicities?
```
erik-3milabs: First of all

```suggestion
# TODO: does it make sense to add writeX conditions?
```

Second: the constraints without conditions are of lower degree. So: no, I would not condition these.

erik-3milabs (commented):
Some basic comments; will do another pass once we have agreed on basic optimizations.

spec/src/memw.toml
RobinJadoul: I omitted that because it's not a requirement for the correctness of this chip in isolation.
This description could alternatively go somewhere with the CPU, but I figured that keeping it here made some sense too, since touching any memory in the CPU probably involves looking at the MEMW spec at some point.
(and additionally, it's good documentation for any ECALL implementations touching memory)

One possibility would be to turn `is_register` from a `Bit` into a `BaseField` and name it something like a memory bank selector.
But I feel like this is already a reasonable naming for the way it is used now

spec/src/memw.toml
RobinJadoul: See above

spec/src/memw.toml
RobinJadoul: Again this is some tension in what the chip itself needs, vs what the overall system enforces.

spec/src/memw.toml
RobinJadoul: Nope, this may have to be a proper assumptions, since we'll need it to delegate to `LT`

spec/src/memw.toml
RobinJadoul: Good point, to be double checked, since it may mess with the `LT`

spec/src/memw.toml
RobinJadoul: It's just a counter in the bitops table for IS_HALFWORD
But I suppose having potential zeroes makes computing the actual logup sum faster, etc

erik-3milabs:
Also, there is no check that `base_address + write_i * i` does not overflow

RobinJadoul:
I can't actually find anything in the RISC-V ISA about overflow for load/store instructions, so I'm not sure whether it is required. It's probably not a horrible idea to add some check though

RobinJadoul:
Updated it without doing optimizations for now, but I wrote the rough ideas down for later.
Set to merge against `spec/nested_iters` so that I could already update the syntax and use the double loops.

spec/src/memw.toml
erik-3milabs: please clarify

spec/src/memw.toml
```diff
@@ -0,0 +1,297 @@
...
+
+[[assumptions]]
+desc = "`IS_BIT<write8>`"
+
+[[assumptions]]
+desc = "`IS_BIT<write2 + write4 + write8>`"
+
+[[assumptions]]
+desc = "`IS_HALFWORD[timestamp[i]]`"
+iter = ["i", 0, 3]
```
erik-3milabs: But `timestamp` is a `DWordWL`


```suggestion
[[assumptions]]
desc = "`IS_WORD[timestamp[i]]`"
iter = ["i", 0, 1]
```

spec/src/memw.toml
RobinJadoul: https://github.com/yetanotherco/lambda_vm/pull/104#discussion_r2668213288

spec/src/memw.toml
erik-3milabs: Got an answer on this?

spec/src/memw.toml
```diff
@@ -0,0 +1,297 @@
...
+kind = "interaction"
+tag = "memory"
+input = ["is_register", ["idx", "address_add", ["-", "i", 1]], "timestamp", ["idx", "value", "i"]]
+multiplicity = ["-", "write8"]
+iter = ["i", 4, 7]
+
+
+[[constraint_groups]]
+name = "output"
+prefix = "O"
```
erik-3milabs, line 275: this prefix is shared with "overflow"

spec/src/memw.toml
erik-3milabs: perhaps list this as an optimization?

spec/src/memw.toml
RobinJadoul: I added this as possible future optimization to check in the list, when my brain was telling me that halfword checks were fine; now that I remember that it's actually DWord**W**L, I should probably think through it directly first

spec/src/memw.toml
erik-3milabs: I believe it is worth adding this clarification in the .typ doc (including why we don't range check `value`)

spec/src/memw.toml
erik-3milabs: Do we now agree that `old_timestamp` is indeed range-constrained through the memory lookup and that this fulfills the assumption that `LT` places on the `lhs` param?

spec/memw.typ
```diff
@@ -0,0 +1,54 @@
...
+== Constraints
+
+#render_constraint_table(chip, config, groups: "consistency")
+
+As long as `timestamp` is properly range-checked, the presence of `old_timestamp`
+in the memory argument automatically ensures appropriate range checking
+(as long as no external entities provide negative multiplicities without range checking the timestamp).
+This ensures the assumptions for `LT` are satisfied.
+
+We additionally also check that the address does not overflow
```
erik-3milabs: tautology
```suggestion
We additionally check that the address does not overflow
```

erik-3milabs (approved):
lgtm

spec/src/memw.toml
RobinJadoul: Yes, I think so; iirc I also attempted to add this clarification in the typst doc already

PR #105, opened 2026-01-06
spec: Fix constraint group lookup

RobinJadoul:
Noticed that all groups were getting the name `g` :)


erik-3milabs (approved):
I don't get why typst interprets `g` as the string `"g"`, rather than binding to the variable 🤔 
I checked: this solves the problem though 🤷‍♂️

PR #107, opened 2026-01-06
Syscalls crate

gianbelinche:
This PR adds a crate for defining the needed syscalls as well as the global allocator.
The new syscalls implemented are print (temporary in order to debug), panic, commit and get private inputs.
Under the syscalls file the are several extern C functions definitions which are not technically syscalls from out point of view, but are syscalls from rust pov. And most of them can be defined using rust without interfacing directly with the vm.


executor/src/vm/execution.rs
```diff
@@ -34,25 +41,34 @@ fn run_from_entrypoint(
...
     let mut registers = Registers::default();
     let mut logs = Vec::new();
     while pc != 0 {
         let next_instruction = memory.load_word(pc)?;
         let instruction = Instruction::parse(next_instruction)?;
         if verbose {
-            println!("registers: {}", &registers);
-            println!("Executing instruction at 0x{:08x}: {:?}", pc, instruction);
+            //println!("registers: {}", &registers);
+            //println!("Executing instruction at 0x{:08x}: {:?}", pc, instruction);
```
fmoletta: Is this intentional?

fmoletta (approved):

MauroToscano (approved):

PR #109, opened 2026-01-07
Optimize fft performance

diegokingston:


MauroToscano:
We have implemented this optimization already

PR #113, opened 2026-01-07
Add multitable Stark

jotabulacios:
This PR refactors both the prover and verifier to handle single-table and   multi-table STARK proofs. 
Both prover and verifier execute Round 1 for all AIRs first to ensure all trace commitments are in the transcript before generating challenges for Round 2.
It also adds tests for multi tables.


crypto/stark/src/prover.rs
```diff
@@ -877,42 +882,83 @@ pub trait IsStarkProver<
...
+    ///
+    /// This unified function handles both single-table and multi-table proving.
+    ///
+    /// The function executes Round 1 for all AIRs first (committing all traces to the transcript),
+    /// then executes Rounds 2-4 for each AIR sequentially. This ensures proper Fiat-Shamir challenge
+    /// generation across all tables.
+    ///
+    /// Warning: the transcript must be safely initialized before passing it to this method.
+    fn multi_prove(
+        airs: &mut [AirAndTrace<'_, Field, FieldExtension, PI>],
```
fmoletta: Why does this need to be mutable?

nicole-graus: Thanks, you're right, we fixed it [here](https://github.com/yetanotherco/lambda_vm/pull/113/commits/7056bc68972733f2b1e78c8d8207a15cd9c4c3c9).

fmoletta (approved):

ColoCarletti (approved):

PR #120, opened 2026-01-09
spec: drop `dot` when multiplying constant with one-letter variable

erik-3milabs:
from
image

to
image

### DIFF
--- a/spec/expr.typ
+++ b/spec/expr.typ
@@ -142,7 +142,15 @@
         pp <= PREC.sub
       )
     },
-    "*": (pp, rec, e) => mwrap($#e.slice(1).map(rec.with(PREC.mul)).join($dot$)$, pp < PREC.mul),
+    "*": (pp, rec, e) => {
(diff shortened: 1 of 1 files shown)


erik-3milabs:
I also find it interesting that this line
```
var: v => if v.len() == 1 { $#v$ } else { $#raw(v)$ },
```
in `expr_to_math` does not seem to work.

Curiously
```
#let v = "i"
[$#v$, $i$]
```
gets rendered as 
image

RobinJadoul:
I think `[$#v$, $i$, $"i"$]` explains why it's different
maybe `$italic(#v)$` is a solution

RobinJadoul (approved):

erik-3milabs:
> I think `[$#v$, $i$, $"i"$]` explains why it's different maybe `$italic(#v)$` is a solution

Sadly, it is not (even aside from the fact that it is `emph` rather than `italic`)

PR #122, opened 2026-01-09
spec: `MUL` chip

erik-3milabs:


spec/src/mul.toml
```diff
@@ -0,0 +1,194 @@
...
+
+[[constraints.def]]
+kind = "arith"
+constraint = "$#`!lhs_signed` => #`lhs_is_negative` = 0$"
+poly = ["*", ["not", "lhs_signed"], "lhs_is_negative"]
+ref = "mul:c:lhs_is_negative_if_unsigned"
+
+[[constraints.def]]
+kind = "arith"
+constraint = "$#`lhs_signed` => #`μ_sum`$"
```
RobinJadoul: I don't think this implication works, as with "fusing", μ_sum may be 2 (or even 4, if we end up with a similar situation from DVRM as we did in the 32-bit case)

spec/src/mul.toml
```diff
@@ -0,0 +1,194 @@
...
+kind = "arith"
+constraint = "$#`rhs_signed` => #`μ_sum`$"
+poly = ["*", "rhs_signed", ["not", "μ_sum"]]
+ref = "mul:c:rhs_signed_implies_mu"
+
+[[constraints.def]]
+kind = "interaction"
+tag = "IS_B19"
+input = [["idx", "carry", "i"]]
+range = ["i", 0, 7]
```
RobinJadoul:
```suggestion
range = ["i", 0, 3]
```

spec/src/mul.toml
```diff
@@ -0,0 +1,194 @@
...
+
+[[variables.auxiliary]]
+name = "rhs_is_negative"
+type = "Bit"
+desc = "whether `rhs` is negative (1) or not (0)"
+
+[[variables.auxiliary]]
+name = "raw_product"
+type = ["B51", 4]
+desc = "raw multiplication output"
```
RobinJadoul, line 56: Do we not need both this and `res` to be range-checked for the correctness of the carry construction?
(this probably gets an automatic check from its definition)

erik-3milabs: > Do we not need [this] to be range-checked for the correctness of the carry construction?

`raw_product` is really just casting `limb_product` as `B51`.
I decided to stick with this construction, knowing that inlining `limb_product` in `raw_product` makes for an annoyingly complex formula.
As such, there is nothing to range check here: if the input is range checked, `raw_product` can only assume one value.

> Do we not need [`res`] to be range-checked for the correctness of the carry construction?

yes, you're absolutely correct

spec/src/mul.toml
erik-3milabs: You're right.
Good thing we don't actually need it :)
I'll remove it.

spec/mul.typ
```diff
@@ -0,0 +1,85 @@
...
+
+== Overview
+When `lhs` and `rhs` are _unsigned_ integers, computing their product $mod 2^128$ comes down to evaluating
+$
+(sum_(j=0)^3 2^(16j) dot #`lhs`_j) dot (sum_(i=0)^3 2^(16i) dot #`rhs`_i) mod 2^128.
+$
+If `lhs` and `rhs` are signed instead, the computation remains nearly identical: 
+one must sign extend `lhs` and `rhs` to twice their size --- forming `lhs_ext` and `rhs_ext` respectively --- and compute
+$
+(sum_(j=0)^7 2^(16j) dot #`lhs_ext`_j) dot (sum_(i=0)^7 2^(16i) dot #`rhs_ext`_i) mod 2^128.
```
erik-3milabs, line 45: @RobinJadoul this "mod 2^128" doesn't sit well with me, since we're operating on signed integers.
Got an idea on how to better explain this?

RobinJadoul: After sign extension, it should exactly match the operation mod 128 though, just like addition remains correct regardless of sign bits
From one point of view, a signed int and an unsigned int are just different choices of representing mod 2^k over ℤ
And with that in mind, then the sign/zero extension is just choosing which lift from 2^64 to 2^128 to choose

So concretely, I think rephrasing it as choosing sign/zero extension based on the sign input and then working mod 2^128 is perfectly fine for me

spec/src/mul.toml
```diff
@@ -0,0 +1,169 @@
...
+ref = "mul:c:carry"
+
+[[constraint_groups]]
+name = "prod"
+
+
+[[constraints.prod]]
+kind = "arith"
+constraint = "$#`raw_product[i]` = sum_(#`k`=0)^1 2^16 sum_(#`j`=0)^(2i+k) #`lhs_ext[j]` dot #`rhs_ext[2i+k-j]`$"
+poly = ["-", ["sum", ["=", "k", 0], "1", ["*", ["^", 2, ["*", 16, "k"]], ["sum", ["=", "j", 0], ["+", ["*", 2, "i"], "k"], ["*", ["idx", "lhs_ext", "j"], ["idx", "rhs_ext", ["-", ["+", ["*", 2, "i"], "k"], "j"]]]]]], ["idx", "raw_product", "i"]]
```
RobinJadoul: Mismatch in $2^{16}$ vs $2^{16k}$

spec/mul.typ
```diff
@@ -0,0 +1,85 @@
...
+$
+where the limbs of `lhs_ext` and `rhs_ext` are treated as _unsigned_ integers.
```
RobinJadoul: Not really the _limbs_ being treated as unsigned

spec/mul.typ
```diff
@@ -0,0 +1,85 @@
...
+  &equiv sum_(j=0)^7 sum_(i=0)^7 2^(16(i+j)) dot #`lhs_ext`_j dot #`rhs_ext`_i mod 2^128 \
+  &stackrel(triangle, equiv) sum_(j=0)^7 sum_(i=0)^(7-j) 2^(16(i+j)) dot #`lhs_ext`_j dot #`rhs_ext`_i mod 2^128 \
+  &stackrel(square, equiv) sum_(j=0)^7 sum_(i=j)^(7) 2^(16i) dot #`lhs_ext`_j dot #`rhs_ext`_(i-j) mod 2^128 \
+  &stackrel(penta, equiv) sum_(i=0)^7 sum_(j=0)^(i) 2^(16i) dot #`lhs_ext`_j dot #`rhs_ext`_(i-j) mod 2^128 \
+//   &equiv sum_(i=0)^7 2^(16i) dot sum_(j=0)^(i) #`lhs_ext`_j dot #`rhs_ext`_(i-j) mod 2^128 \
+  &equiv sum_(i=0)^3 sum_(k=0)^1 sum_(j=0)^(2i+k) 2^(16(2i+k)) dot #`lhs_ext`_j dot #`rhs_ext`_(2i+k-j) mod 2^128 \
+  &equiv sum_(i=0)^3 2^(32i) dot sum_(k=0)^1 2^(16k) dot sum_(j=0)^(2i+k) #`lhs_ext`_j dot #`rhs_ext`_(2i+k-j) mod 2^128
+$
+where at step
+- $triangle$ we can ignore $j > 7-i$, since that makes $2^(16(i+j)) equiv 0 mod 2^128$,
```
RobinJadoul: `i` and `j` swapped?

spec/mul.typ
```diff
@@ -0,0 +1,85 @@
...
+- $penta$ we swap the sums.
+
+Note that `raw_product` captures the second summation in this last formula (see @mul:c:raw_product).
+However, the limbs in `raw_product` may require up to 51 bits to be represented.
+What remains then is to carry the overflow of each limb to the next, ensuring `res` represents the same value as `raw_product`, but with limbs in the range $[0, 2^32)$.
+This is simply constrained by `carry`'s definition and @mul:c:carry.
+From these two, we gather that
+$
+  #`raw_product`_0 - #`res`_0 in { i dot 2^32 | i in [0, 2^19) }
+$
```
RobinJadoul, line 73: What is the limit on `2^19` here exactly? 
iirc, we need `res + c_max * 2^32 < p`, right? (Maybe this needs some elaboration, perhaps in a generic form in `ADD` with a reference here)

So one thing to consider is relaxing MUL-C3 to `IS_B20`, which may be more convenient and recurring with the 20-bit lookups we already have due to HWSL.

spec/src/mul.toml
```diff
@@ -0,0 +1,169 @@
...
+type = ["B20", 4]
+desc = "carry values"
+def = {idx="i", polys=[
+    {range=0, poly=["*", ["^", 2, -32], ["-", ["idx", "raw_product", 0], ["idx", "res", 0]]]},
+    {range=[1, 3], poly=["*", ["^", 2, -32], ["-", ["+", ["idx", "raw_product", "i"], ["idx", "carry", ["-", "i", 1]]], ["idx", "res", "i"]]]},
+]}
+
+[[variables.virtual]]
+name = "μ_sum"
+type = "Bit"
```
RobinJadoul: If both multiplicities are `BaseField`, this should be too

```suggestion
type = "BaseField"
```

spec/mul.typ
```diff
@@ -0,0 +1,94 @@
...
+By construction, $#`raw_product`_i < 2^51$ for all $i in [0, 3]$, far exceeding the 32-bits that fit in a single `Word`-limb.
+What remains then is to reduce each limb of `raw_product` $mod 2^32$, carrying the overflow of each limb to the next, constructing the output `res` in doing so.
+
+This reduce-and-carry operation is constrained @mul:a:res and @mul:c:carry, combined with `carry`'s eloquent definition.
+@mul:c:carry and `carry`'s definition enforce that
+$
+  forall i in [0, 3]: #`raw_product`_i + #`carry`_(i-1) - #`res`_i in { k dot 2^32 | k in [0, 2^20) }
+$
+with $#`carry`_(-1) = 0$ for simplicity.
+In other words: $#`res`_i >= #`raw_product`_i + #`carry`_(i-1) mod 2^32$.
```
RobinJadoul:
```suggestion
In other words: $#`res`_i equiv #`raw_product`_i + #`carry`_(i-1) (mod 2^32)$.
```
The `>=` confuses me, but I think this is correct?

RobinJadoul (approved):
Couple of minor things, but good to go otherwise

PR #130, opened 2026-01-12
spec: update range specifications to iters concept

RobinJadoul:


spec/chip.typ
```diff
@@ -54,25 +77,39 @@
...
+    let gather_indices(obj) = iters_of(obj, name: idx).map(it => it.first())
+    let index_all(expr, indices) = {
+      if indices.len() == 0 {
+        expr
+      } else {
+        index_all(("idx", expr, indices.first()), indices.slice(1))
+      }
+    }
     if "poly" in def {
+      // assert(false, message: repr(index_all(var_name, gather_indices(def))))
```
erik-3milabs:
```suggestion
```

spec/chip.typ
```diff
@@ -54,25 +77,39 @@
...
 
+    let idx = def.at("idx", default: none)
+    let gather_indices(obj) = iters_of(obj, name: idx).map(it => it.first())
+    let index_all(expr, indices) = {
+      if indices.len() == 0 {
+        expr
+      } else {
+        index_all(("idx", expr, indices.first()), indices.slice(1))
+      }
+    }
```
erik-3milabs, line 113: I propose to collapse the recursion:
```suggestion
    let index_all(expr, indices) = {
      for index in indices {
        expr = ("idx", expr, index)
      }
      expr
    }

```

spec/chip.typ
```diff
@@ -54,25 +77,39 @@
...
-        expr_to_math((":=", ("idx", var_name, def.idx), def.poly)),
-        render_def_range(def.idx, def.range)
+        expr_to_math((":=", index_all(var_name, gather_indices(def)), def.poly)),
+        render_def_iters(iters_of(def, name: idx))
       )
     } else if "polys" in def {
+      assert(
+        def.polys.all(poly =>
+          gather_indices(poly) == gather_indices(def.polys.first())
+        ), message: "Can only do multiple polys if they're indexed identically")
```
erik-3milabs: perhaps this is nicer?
```suggestion
      assert(
        def.polys.all(poly => gather_indices(poly)).reduce().len() == 1, 
        message: "Can only do multiple polys if they're indexed identically"
      )
```

spec/chip.typ
RobinJadoul: Whoops, debug code :)

spec/chip.typ
RobinJadoul:
```suggestion
      assert(
        def.polys.map(gather_indices).dedup().len() == 1,
        message: "Can only do multiple polys if they're indexed identically"
      )
```

RobinJadoul: Did you mean something like this instead?

erik-3milabs: ah, yeah, this _actually_ works :see_no_evil:

erik-3milabs (approved):

PR #133, opened 2026-01-13
spec: Add support for specifying padding values of columns

RobinJadoul:
Maybe double check I didn't add any dumb padding values to the existing tables :)

This doesn't support iters (see #130 too) yet, which may be good for array types; but in general will probably be overkill for most purposes.

### DIFF
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -24,6 +24,33 @@
   .sum()
 }
 
+#let render_chip_padding_table(chip, config) = {
+  show figure: set block(breakable: true)
+  figure(table(
+    columns: (auto, auto, auto),
+    inset: 6pt,
+    align: (right + top, center + top, left + top),
+    stroke: none,
+    table.header([*Column*], [],  [*Padding value*]),
+    table.hline(stroke: stroke(thickness: 2pt)),
+    ..for (cat, vars) in chip.variables.pairs() {
+      if cat not in config.variables.categories.instantiated {
+        continue
+      }
+      for var in vars {
+        if config.variables.types.filter(t => t.label == var.type).all(t => t.at("preprocessed", default: false)) {
+          continue
+        }
+        (
+          [#raw(var.name)], 
+          [$:=$],
+          [#expr_to_math(var.pad)], 
+        )
+      }
+    },
+  ), caption: [Overview of padding values for #chip.name chip.])
+}
+
 /// Generates a table listing `chip`'s columns.
 #let render_chip_column_table(chip, config) = {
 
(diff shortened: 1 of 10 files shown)


spec/chip.typ
```diff
@@ -24,6 +24,33 @@
...
+        (
+          [#raw(var.name)], 
+          [$:=$],
+          [#expr_to_math(var.pad)], 
+        )
+      }
+    },
+  ), caption: [Overview of padding values for #chip.name chip.])
+}
+
```
erik-3milabs, line 52: Consider the following refactor

```suggestion
#let render_chip_padding_table(chip, config) = {
  // Whether `var` is a preprocessed variable.
  let is_preprocessed(var) = {
    config.variables.types
    .filter(t => t.label == var.type)
    .all(t => t.at("preprocessed", default: false))
  }

  let instantiated_vars = config.variables.categories.instantiated.map(c => chip.variables.at(c)).flatten()

  show figure: set block(breakable: true)
  figure(table(
    columns: (auto, auto, auto),
    inset: 6pt,
    align: (right + top, center + top, left + top),
    stroke: none,
    table.header([*Column*], [], [*Padding value*]),
    table.hline(stroke: stroke(thickness: 2pt)),
    ..for var in instantiated_vars {
      if not is_preprocessed(var) {
        ([#raw(var.name)], [$:=$], [#expr_to_math(var.pad)],)
      }
    },
  ), caption: [Overview of padding values for #chip.name chip.])
}

```

spec/src/branch.toml
```diff
@@ -30,11 +34,13 @@ desc = "Selects between `pc` and `register` as base address, needed for the `JAL
 name = "next_pc_high"
 type = ["Half", 3]
 desc = "The upper part of the next pc"
+pad = 0 # TODO: improve handling for arrays
```
erik-3milabs: I propose we remove this comment here, and rather spin this off as an issue

```suggestion
pad = 0
```

spec/src/cpu.toml
```diff
@@ -1,3 +1,4 @@
+# TODO: padding
```
erik-3milabs: Let's remove this, and spin it off as an issue


```suggestion
```

spec/src/branch.toml
RobinJadoul: Or we do both; I kinda like having the marker there to easily be reminded where that was

erik-3milabs:
With #122 merged, could you add the padding section to that chip as well?

spec/src/branch.toml
erik-3milabs: then we should probably go for
```suggestion
pad = 0 # TODO(#128): improve handling for arrays
```

erik-3milabs (approved):

PR #138, opened 2026-01-13
spec: `BITWISE` chip

erik-3milabs:


spec/src/bitwise.toml
```diff
@@ -0,0 +1,200 @@
...
+precomputed = "true"
+
+[[variables.output]]
+name = "ZERO"
+type = "Bit"
+desc = "whether $#`X` = 0 and #`Y` = 0$"
+precomputed = "true"
+
+[[variables.output]]
+name = "SLL"
```
RobinJadoul, line 58: Name mismatch between here and the interaction

RobinJadoul (approved):
Probably useful to keep track of optimization ideas for this that  we've had (probably just in the typst).

A probably incomplete list:
- Batch `IS_BYTE` to have `IS_BYTE[X, Y]`
- Merge MSB8 into MSB16 if no MSB8 consumer is relying on it to perform byte range checks
- (implementation) split 16-bit and 20-bit tables
- (see also SHIFT) combine HWSL and HWSLC

erik-3milabs:
@RobinJadoul before I merge, do you agree with 769fc5e?

RobinJadoul:
Sure thing :+1:

PR #143, opened 2026-01-14
spec: `DECODE`

erik-3milabs:


spec/src/decode.toml
```diff
@@ -0,0 +1,87 @@
...
+
+[[variables.output]]
+name = "word_instr"
+type = "Bit"
+desc = "Whether the instruction is a `*W` instruction, requiring the inputs and outputs to be (sign) extended."
+
+[[variables.output]]
+name = "one_hot"
+type = ["Bit", 16]
+desc = "A 16-column vector comprised of the flags `(ADD, SUB, SLT, AND, OR, XOR, SHIFT, JALR, BEQ, BLT, LOAD, STORE, MUL, DIVREM, ECALL, EBREAK)`. The vector is 'one-hot', meaning that in every row exactly one element is $1$; the others are $0$."
```
erik-3milabs: I think we should split this up into 16 separate columns.
This would improve machine-readability, as well as improve consistency (e.g., right now, the .typ document mentions these flags, but they don't (clearly) correspond to a variable yet.

spec/decode.typ
```diff
@@ -0,0 +1,154 @@
...
+  [`XORI  rd, rs1, imm`], [`XOR`], [],
+  [`SLLI  rd, rs1, imm`], [`SHIFT`], [],
+  [`SLLIW rd, rs1, imm`], [`SHIFT`], [`word_instr`],
+  [`SRLI  rd, rs1, imm`], [`SHIFT`], [`mp_selector`],
+  [`SRLIW rd, rs1, imm`], [`SHIFT`], [`mp_selector`, `word_instr`],
+  [`SRAI  rd, rs1, imm`], [`SHIFT`], [`mp_selector`, `signed`],
+  [`SRAIW rd, rs1, imm`], [`SHIFT`], [`mp_selector`, `signed`, `word_instr`],
+
+  ..subheader("LUI/AUIPC", []),
+  [`LUI   rd, imm`], [#sym.arrow.double.r], [`ADDI rd, x0, imm`],
```
erik-3milabs: 1. perhaps add a clickable link on the `==>` or the `ADDI` to auto scroll to that definition.
2. alternatively, inline the `==>`, by setting op-flag=`ADDI` and other=`rs1=x0`

Wdyt?

spec/decode.typ
```diff
@@ -0,0 +1,154 @@
...
+  [`DIVW   rd, rs1, rs2`], [`DIVREM`], [`signed`, `word_instr`],
+  [`DIVU   rd, rs1, rs2`], [`DIVREM`], [],
+  [`DIVUW  rd, rs1, rs2`], [`DIVREM`], [`word_instr`],
+  [`REM    rd, rs1, rs2`], [`DIVREM`], [`signed`, `muldiv_selector`],
+  [`REMW   rd, rs1, rs2`], [`DIVREM`], [`signed`, `muldiv_selector`, `word_instr`],
+  [`REMU   rd, rs1, rs2`], [`DIVREM`], [`muldiv_selector`],
+  [`REMUW  rd, rs1, rs2`], [`DIVREM`], [`muldiv_selector`, `word_instr`],
+
+  ..subheader("MISC", []),
+  [`FENCE`], [#sym.arrow.double.r#footnote("Note: this is a no-op")], [`ADDI 0, x0, 0`],
```
erik-3milabs: I'm not sure footnotes are the way to go... 

Perhaps we should just have "notes" after the table, and then (somehow) refer to those, rather than using footnotes.

spec/src/decode.toml
RobinJadoul: With an eye on machine-readability and having the tomls be the source of truth, I would consider having `decode.toml` specify the "actual", compressed columns, as consumed by the CPU.
The contents preprocessed table won't be generated from a machine-readable spec any time soon, but having the columns specced the way they will be actually used or implemented may facilitate some generation, and importantly enable consistency/"type" checking for the composition of all our chips.

For typesetting convenience, we can definitely specify separate variables/columns from e.g. some `decode_uncompressed.toml` file.

spec/decode.typ
RobinJadoul: That sounds reasonable to me. Maybe a wikipedia-style column in the table with an annotation counter.

spec/decode.typ
RobinJadoul: I think it should be simple enough to just inline; maybe with an added note what it's equivalent to.

spec/decode.typ
```diff
@@ -0,0 +1,154 @@
...
+  [`BEQ   rs1, rs2, imm`], [`BEQ`], [],
+  [`BNE   rs1, rs2, imm`], [`BEQ`], [`mp_selector`],
+  [`BLT   rs1, rs2, imm`], [`BLT`], [`signed`],
+  [`BLTU  rs1, rs2, imm`], [`BLT`], [],
+  [`BGE   rs1, rs2, imm`], [`BLT`], [`signed`, `mp_selector`],
+  [`BGEU  rs1, rs2, imm`], [`BLT`], [`mp_selector`],
+  [`BGT   rs1, rs2, imm`], [#sym.arrow.double.r #footnote["BGT, BGTU, BLE, and BLEU can be synthesized by reversing the operands to BLT, BLTU, BGE, and BGEU, respectively", RISC-V unprivileged ISA, page 32.] <bgt-bgtu-ble-bleu>], [`BLT `  *`rs2, rs1`*,` imm`],
+  [`BGTU  rs1, rs2, imm`], [#sym.arrow.double.r  #footnote(<bgt-bgtu-ble-bleu>)], [`BLTU`  *`rs2, rs1`*,` imm`],
+  [`BLE   rs1, rs2, imm`], [#sym.arrow.double.r  #footnote(<bgt-bgtu-ble-bleu>)], [`BGE `  *`rs2, rs1`*,` imm`],
+  [`BLEU  rs1, rs2, imm`], [#sym.arrow.double.r  #footnote(<bgt-bgtu-ble-bleu>)], [`BGEU`  *`rs2, rs1`*,` imm`],
```
RobinJadoul: I think these instructions are superfluous from the point of view of an execution environment, as (afaik) it's the compiler that does the mentioned synthesis, and there is no encoding for the reversed version.

spec/src/decode.toml
```diff
@@ -0,0 +1,87 @@
...
+
+[[variables.output]]
+name = "memory_8bytes"
+type = "Bit"
+desc = "whether the memory access (read or write) touches at least $8$ bytes."
+
+[[variables.output]]
+name = "c_type"
+type = "Bit"
+desc = "Whether the instruction is of type `C`, i.e., whether it is $2$ bytes long instead of $4$."
```
RobinJadoul: This will need some "general" handling that we can't properly specify in the tables

spec/decode.typ
erik-3milabs: The downside of inlining is date duplication.
E.g., if we were to change how `ADDI` is encoded, we also have to update `AUIPC` and `LUI`
We're 100% going to forget about that.
Not a fan of that idea.

erik-3milabs: I decided to go ahead an inline them, adding comments on why we approach it this way.

spec/decode.typ
erik-3milabs: Let me know what you think of the current setup

spec/decode.typ
erik-3milabs: I removed them.

spec/src/decode.toml
erik-3milabs: I added a subsubsection on `c_type`. Let me know what you think!

spec/decode.typ
```diff
@@ -0,0 +1,35 @@
...
+)
+
+#let config = load_config()
+#let chip = load_chip("src/decode.toml", config)
+#show: book-page.with(title: "DECODE chip")
+
+#let decode = raw(chip.name)
+
+= #decode table
+All `RV64ACIM` instruction are to be decoded to a format that can be interpreted by the VM.
```
RobinJadoul: I think it's 

```suggestion
All `RV64IMAC` instruction are to be decoded to a format that can be interpreted by the VM.
```

but even then, I don't think we  support the A extension.

RobinJadoul: https://riscv.github.io/riscv-isa-manual/snapshot/unprivileged/#_subset_naming_convention

spec/decode.typ
```diff
@@ -0,0 +1,35 @@
...
+This section outlines the decoding table being used in the VM.
+For reasons of efficiency, data in this table is significantly compressed.
+Since reasoning about this compressed form is needlessly complex, the `decode (uncompressed)` section presents the same table in uncompressed form, and explains how to decode `RV64IM` the assembly instructions to it.
```
RobinJadoul:
```suggestion
Since reasoning about this compressed form is needlessly complex, the `decode (uncompressed)` section presents the same table in uncompressed form, and explains how to decode `RV64IM` assembly instructions to it.
```

spec/src/decode.toml
```diff
@@ -0,0 +1,56 @@
+name = "DECODE"
+
+[[variables.output]]
+name = "pc"
+type = "DWordWL"
+desc = "value of the program counter this instruction is associated with."
+pad = 0
```
RobinJadoul: This should definitely be odd

RobinJadoul: But ig generally padding is what I wanted to look at soon, together with CPU padding anyway

spec/book.typ
```diff
@@ -8,6 +8,8 @@
     #chapter("variables.typ")[Variables]
     #chapter("is_bit.typ")[IS_BIT template]
     #chapter("add.typ")[ADD template]
+    #chapter("decode.typ")[DECODE chip]
+    #chapter("decode_uncompressed.typ")[DECODE (uncompressed) chip]
```
RobinJadoul: idk whether it makes sense to split this into separate chapters

spec/src/decode_uncompressed.toml
```diff
@@ -0,0 +1,159 @@
...
+
+[[variables.output]]
+name = "mem_4B"
+type = "Bit"
+desc = "whether the memory access (read or write) touches at least $4$ bytes."
+
+[[variables.output]]
+name = "mem_8B"
+type = "Bit"
+desc = "whether the memory access (read or write) touches at least $8$ bytes."
```
RobinJadoul: I think I messed with the "at least" interpretation in favor of "exactly" in the latest MEMW/LOAD rework.

spec/src/decode_uncompressed.toml
```diff
@@ -0,0 +1,159 @@
...
+
+[[variables.output]]
+name = "imm"
+type = "DWordWL"
+desc = "the *fully extended (!)* 64-bit version of the immediate."
+
+[[variables.output]]
+name = "signed"
+type = "Bit"
+desc = "selector used to indicate signed or unsigned input interpretation. Currently, this is used to choose between <TODO>."
```
RobinJadoul: Do we need the "choose between" for signed interpretation? This is more intuitive/normal than the general purpose selector.

spec/src/decode_uncompressed.toml
```diff
@@ -0,0 +1,159 @@
...
+
+[[variables.output]]
+name = "mp_selector"
+type = "Bit"
+desc = """Multi-purpose selector used by different ALU operations for different purposes. Currently, it is used
+    - by the `MUL` chip to select between `MUL`/`MULH` and `MULH[S]U`, and
+    - as flag for inverting the condition of conditional branches (see `branch_cond`)
```
RobinJadoul: This should refer to the CPU instead, I assume

spec/decode_uncompressed.typ
```diff
@@ -0,0 +1,185 @@
...
+== Columns
+For the purposes of this explanation, we decompress the (uncompressed) #decode table's `packed_decode` variable into its constituent variables: 
+#render_chip_column_table(chip, config)
+
+== Decoding
+The below table illustrates how each instruction should be expressed in the decoding table.
+The columns of the table represent the following:
+- *`operation`*: the assembly operation being encoded,
+- *`op-flag`*: which of the "`ALU` selector flags" operation flags to set. Each operation sets exactly one.
+- *`w_reg`*, *`w_inst`*, *`signed`*: whether to set the `write_register`, `word_inst` or `signed` flag, respectively,
```
RobinJadoul:
```suggestion
- *`w_reg`*, *`w_instr`*, *`signed`*: whether to set the `write_register`, `word_instr` or `signed` flag, respectively,
```

spec/decode_uncompressed.typ
```diff
@@ -0,0 +1,185 @@
...
+    table.hline(stroke: 1.5pt),
+    table.footer([*Operation*], [*op-flag*], [*`w_reg`*], [*`w_instr`*], [*`signed`*], [*other*]),
+    ),  
+    caption: [Decoding table]
+  )
+}
+
+#let decoding = (
+    // OP-IMM
+  ([`ADDI[W]   rd, rs1, imm`], [`ADD`], [$#`rd` eq.not 0$], [`[W]`], [], [], []),
```
RobinJadoul: I believe ADD[I] and SUB[I] should be signed for the correctness of the W extension

spec/src/config.toml
```diff
@@ -58,6 +58,11 @@ label = "B35"
 subtypes = ["BaseField"]
 desc = "Variable that can only assume values in the range $[0, 2^35)$."
 
+[[variables.types]]
+label = "B49"
+subtypes = ["BaseField"]
+desc = "Variable that can only assume values in the range $[0, 2^49)$."
+
```
RobinJadoul: Is it worth the annoyance of listing this? Any change to the flags or decoding will instantly impact this, and the bound itself is not too relevant...

spec/decode.typ
RobinJadoul: Looks good!

spec/decode.typ
erik-3milabs: Apologies. I thought we did. Wdyt of

```suggestion
All `RV64IMC` instruction are to be decoded to a format that can be interpreted by the VM.
```

spec/src/decode_uncompressed.toml
erik-3milabs:
```suggestion
desc = "selector used to indicate signed or unsigned input interpretation."
```

spec/src/decode_uncompressed.toml
erik-3milabs: > This should refer to the CPU instead, I assume

wdyt of this?

```suggestion
desc = """Multi-purpose selector used by the CPU to configure several ALU operations in different ways. Currently, it is used
    - to select between `MUL`/`MULH` and `MULHSU`/`MULHU`,
    - as flag for inverting the condition of conditional branches (see `branch_cond`), and
```

erik-3milabs: Resolved in favor of https://github.com/yetanotherco/lambda_vm/pull/143/changes#r2707162862

spec/decode_uncompressed.typ
```diff
@@ -0,0 +1,185 @@
...
+  // OP - M
+  ([`MUL[W]    rd, rs1, rs2`], [`MUL`], [$#`rd` eq.not 0$], [`[W]`], [1], [`mp_selector`], []),
+  ([`MULH      rd, rs1, rs2`], [`MUL`], [$#`rd` eq.not 0$], [], [1], [`mp_selector`, `muldiv_selector`], []),
+  ([`MULHU     rd, rs1, rs2`], [`MUL`], [$#`rd` eq.not 0$], [], [], [`muldiv_selector`], []),
+  ([`MULHSU    rd, rs1, rs2`], [`MUL`], [$#`rd` eq.not 0$], [], [1], [`muldiv_selector`], []),
+  ([`DIV[U][W] rd, rs1, rs2`], [`DIVREM`], [$#`rd` eq.not 0$], [`[W]`], [#sym.not`[U]`], [], []),
+  ([`REM[U][W] rd, rs1, rs2`], [`DIVREM`], [$#`rd` eq.not 0$], [`[W]`], [#sym.not`[U]`], [`muldiv_selector`], []),
+  // LUI/AUIPC
+  ([`LUI       rd, imm`], [`ADDI`], [$#`rd` eq.not 0$], [], [], [], [#ref_note(<note-lui>)]),
+  ([`AUIPC     rd, imm`], [`ADDI`], [$#`rd` eq.not 0$], [], [], [`rs1 := x255`], [#ref_note(<note-auipc>)]),
```
erik-3milabs: These set the `ADD` flag, not the `ADDI` flag 🤦‍♂️ 


```suggestion
  ([`LUI       rd, imm`], [`ADD`], [$#`rd` eq.not 0$], [], [], [], [#ref_note(<note-lui>)]),
  ([`AUIPC     rd, imm`], [`ADD`], [$#`rd` eq.not 0$], [], [], [`rs1 := x255`], [#ref_note(<note-auipc>)]),
```

spec/decode_uncompressed.typ
```diff
@@ -0,0 +1,185 @@
...
+  // STORE
+  ([`SD       rs1, rs2, imm`], [`STORE`], [], [], [], [`mem_2B`, `mem_4B`, `mem_8B`], []),
+  ([`SW       rs1, rs2, imm`], [`STORE`], [], [], [], [`mem_2B`, `mem_4B`], []),
+  ([`SH       rs1, rs2, imm`], [`STORE`], [], [], [], [`mem_2B`], []),
+  ([`SB       rs1, rs2, imm`], [`STORE`], [], [], [], [], []),
+  // ECALL/EBREAK
+  ([`ECALL`], [`ECALL`], [1], [], [], [$#`rs1` := #`x17`$, $#`rs2` := #`x11`$, $#`rd` := #`x10`$], [#ref_note(<note-ecall>)]),
+  ([`EBREAK`], [`EBREAK`], [], [], [], [], []),
+  // FENCE
+  ([`FENCE`], [`ADDI`], [], [], [], [], [#ref_note(<note-fence>)]),
```
erik-3milabs: This guy also sets the `ADD` flag, not the `ADDI` flag


```suggestion
  ([`FENCE`], [`ADD`], [], [], [], [], [#ref_note(<note-fence>)]),
```

spec/decode_uncompressed.typ
erik-3milabs: > I believe ADD[I] [...] should be signed for the correctness of the W-extension

I agree.

> I believe [...] SUB[I] should be signed for the correctness of the W extension

I agree with `SUBW`; AFAIK, there is no `SUBI` or `SUBIW` operation.


Before I apply the change, what is your take on setting `signed := [W]` for `SLL[I]W`, `SRL[I]W`, and `SRA[I]W`? According to the ISA (pg 41):
> SLL, SRL, and SRA perform logical left, logical right, and arithmetic right shifts on the value in register rs1 by the shift amount held in register rs2. In RV64I, only the low 6 bits of rs2 are considered for the shift amount. SLLW, SRLW, and SRAW are RV64I-only instructions that are analogously defined but operate on 32-bit values and sign-extend their 32-bit results to 64 bits.

and
> SLLIW, SRLIW, and SRAIW are RV64I-only instructions that are analogously defined [to the SLLI, SRLI and SRAI instructions,] but operate on 32-bit values and sign-extend their 32-bit results to 64 bits.

I'm not entirely clear on the definition of "sign-extend". If it means "repeat the 31st bit for the upper 32 bits", we would be in trouble here with `SRL[I]W`; that operation needs `signed` to be `0` to compute the logical shift (rather than the arithmetic shift), but would also need `signed=1` to sign-extend the 32-bit output up to 64 bits (which might be a 1-extension when `shift = 0 mod 32`). Sadly, the `mp_selector` is already in use there to indicate whether to left or right-shift.


It looks like the same problem pops up with `DIV[U]W` and `REM[U]W`. According to the ISA, their 32-bit output is to be sign-extended:

> DIVW and DIVUW are RV64 instructions that divide the lower 32 bits of rs1 by the lower 32 bits of rs2, treating them as signed and unsigned integers, placing the 32-bit quotient in rd, sign-extended to 64 bits. REMW and REMUW are RV64 instructions that provide the corresponding signed and unsigned remainder operations. Both REMW and REMUW always sign-extend the 32-bit result to 64 bits, including on a divide by zero.

Here, we have that the output of `DIVUW` and `REMUW` are clearly unsigned, so "sign-extending" them would then be zero-extending? Either way, this operation does not use the `mp_selector`, so we have enough breathing room. We just need to know how to interpret "sign-extend"

spec/src/config.toml
erik-3milabs: Given the current typst code-base, we _have_ to list it for the compilation not to fail; we explicitly test that every variable type used in a chip is defined in this list.

I would not be in favor of removing that check; I think consistency and completeness are very important.

Meanwhile, I also agree that this is (i) cumbersome, (ii) prone to change and (iii) not really informative. Perhaps it would be best to just have `packed_encode: BaseField` and remove `B49` from this list. Do you agree?

spec/src/decode.toml
erik-3milabs: Do you agree with this?

```suggestion
# TODO(#136): fix this when padding the CPU
pad = 1
```

spec/src/decode_uncompressed.toml
erik-3milabs: You're right.

```suggestion
[[variables.output]]
name = "mem_2B"
type = "Bit"
desc = "whether the memory access (read or write) touches exactly $2$ bytes."

[[variables.output]]
name = "mem_4B"
type = "Bit"
desc = "whether the memory access (read or write) touches exactly $4$ bytes."

[[variables.output]]
name = "mem_8B"
type = "Bit"
desc = "whether the memory access (read or write) touches exactly $8$ bytes."
```

spec/book.typ
erik-3milabs: Having given it some thought, I agree that merging this into (just) `decode` is perfectly fine.
Before I do, let us first resolve the other comments and then apply the transfer.

spec/decode_uncompressed.typ
erik-3milabs: We resolved this while chatting this morning.

spec/src/config.toml
RobinJadoul: Yes, that's definitely more in line with what I was thinking of than removing type sanity checks

spec/src/decode_uncompressed.toml
RobinJadoul: Keep in mind the suggested change isn't enough and the decoding semantics also changed with this

spec/src/decode_uncompressed.toml
RobinJadoul: Or just refer to the CPU documentation on the corresponding column. That might make for less duplication and danger of desync

RobinJadoul (approved):
Nice!

PR #144, opened 2026-01-14
spec: LOAD chip

RobinJadoul:


spec/src/load.toml
```diff
@@ -0,0 +1,144 @@
...
+name = "read8"
+type = "Bit"
+desc = "Whether to read exactly 8 values"
+pad = 0
+
+[[variables.input]]
+name = "signed"
+type = "Bit"
+desc = "Whether to sign-extend (1) or zero-extend (0)"
+pad = 0
```
erik-3milabs, line 39: There is no assumption on the range of `signed`

spec/src/load.toml
```diff
@@ -0,0 +1,144 @@
...
+
+# Virtual
+
+# Multiplicity
+
+[[variables.multiplicity]]
+name = "μ"
+type = "Bit"
+desc = ""
+pad = 0
```
erik-3milabs, line 71: Come to think of it: should we not be range checking the multiplicity columns (if they are non-zero)? 🤔 

(this comment applies more broadly than this PR; most chips don't perform this range check I don't think)

spec/src/load.toml
```diff
@@ -0,0 +1,144 @@
...
+input = [["idx", "res", 1]]
+output = "sign_bit"
+multiplicity = "read2"
+
+[[constraints.all]]
+kind = "interaction"
+tag = "MSB8"
+input = [["idx", "res", 3]]
+output = "sign_bit"
+multiplicity = "read4"
```
erik-3milabs, line 132: Missing `MSB[sign_bit; res[0]]` with multiplicity `read1 := 1 - read2 - read4 - read8`

erik-3milabs: Also, (once merged,) we could use the `SIGN` template here.

spec/src/load.toml
```diff
@@ -0,0 +1,144 @@
...
+[[constraints.all]]
+kind = "arith"
+constraint = "$!(#`read4` + #`read8`) => #`res`_i = #`signed` dot #`sign_bit` dot 255$"
+poly = ["*", ["-", 1, "read4", "read8"], ["-", ["idx", "res", "i"], ["*", "signed", "sign_bit", 255]]]
+iter = ["i", 2, 3]
+
+[[constraints.all]]
+kind = "arith"
+constraint = "$!(#`read2` + #`read4` + #`read8`) => #`res`_i = #`signed` dot #`sign_bit` dot 255$"
+poly = ["*", ["-", 1, "read2", "read4", "read8"], ["-", ["idx", "res", 1], ["*", "signed", "sign_bit", 255]]]
```
erik-3milabs, line 149: Mismatch between `constraint` and `poly` on the index of `res`

spec/src/load.toml
```diff
@@ -0,0 +1,144 @@
...
+name = "read4"
+type = "Bit"
+desc = "Whether to read exactly 4 values"
+pad = 0
+
+[[variables.input]]
+name = "read8"
+type = "Bit"
+desc = "Whether to read exactly 8 values"
+pad = 0
```
erik-3milabs, line 33: These descriptions mention "values", whereas `res`'s explicitly mentions "bytes".
I believe this can also be "bytes".

spec/src/load.toml
```diff
@@ -0,0 +1,144 @@
...
+
+[[constraints.all]]
+kind = "arith"
+constraint = "$!(#`read4` + #`read8`) => #`res`_i = #`signed` dot #`sign_bit` dot 255$"
+poly = ["*", ["-", 1, "read4", "read8"], ["-", ["idx", "res", "i"], ["*", "signed", "sign_bit", 255]]]
+iter = ["i", 2, 3]
+
+[[constraints.all]]
+kind = "arith"
+constraint = "$!(#`read2` + #`read4` + #`read8`) => #`res`_i = #`signed` dot #`sign_bit` dot 255$"
```
erik-3milabs: Consider introducing `read1` as a virtual column replacing `1 - read2 - read4 - read8`

spec/src/load.toml
RobinJadoul: I don't think so. 
Since any nonzero multiplicity enforces the constraints, having a higher multiplicity is just the same as having repeated rows.
The only danger comes from overflowing the multiplicity for that tuple, but overflowing on the callee side becomes equivalent with just doing smaller multiplicities, I think.

spec/src/load.toml
RobinJadoul: `SIGN` may be overkill due to the `!SIGNED => sign = 0` constraint?
In this case, it would introduce a few more constraints (one per `SIGN` call), compared to the `signed *` that's already in the extension constraints. OTOH, maybe it brings down the max degree of the chip.

erik-3milabs: I don't think it would bring down the max degree of the chip: you're already conditioning of `readx` here, so the sign extension would still be `signed * sign_bit * 255`

spec/src/load.toml
```diff
@@ -0,0 +1,160 @@
...
+[[constraints.all]]
+kind = "arith"
+constraint = "$!(#`read4` + #`read8`) => #`res`_i = #`signed` dot #`sign_bit` dot 255$"
+poly = ["*", ["-", 1, "read4", "read8"], ["-", ["idx", "res", "i"], ["*", "signed", "sign_bit", 255]]]
+iter = ["i", 2, 3]
+
+[[constraints.all]]
+kind = "arith"
+constraint = "$!(#`read2` + #`read4` + #`read8`) => #`res`_1 = #`signed` dot #`sign_bit` dot 255$"
+poly = ["*", ["-", 1, "read2", "read4", "read8"], ["-", ["idx", "res", 1], ["*", "signed", "sign_bit", 255]]]
```
erik-3milabs, line 149: Update to

```suggestion
[[constraints.all]]
kind = "arith"
constraint = "$read1 => #`res`_1 = #`signed` dot #`sign_bit` dot 255$"
poly = ["*", "read1", ["-", ["idx", "res", 1], ["*", "signed", "sign_bit", 255]]]
```

to take `mu` into account. (technically speaking this is not necessary; the padding rows being all-zeroes would still be OK)

RobinJadoul: I thought about it, but was a bit hesitant since it breaks the flow of the other constraints: suddenly you go from "if the byte isn't covered by the read" to "if you read 1 byte"

erik-3milabs (approved):
lgtm

PR #145, opened 2026-01-14
feat(math): add TwiddleCache for FFT twiddle factor caching

diegokingston:
Add thread-safe cache for precomputed twiddle factors to avoid redundant computation during repeated FFT operations.

Changes:
- Add Hash, PartialEq, Eq derives to RootsConfig for HashMap key usage
- Add TwiddleCache<F> struct with RwLock<HashMap> for thread-safety
- Methods: new(), get_or_compute(), precompute(), clear(), len(), is_empty()
- Add comprehensive tests for cache functionality

Expected impact: ~23-25% FFT speedup when twiddle factors are reused.


fmoletta (approved):

MauroToscano (changes requested):
In another PR

- Add a benchmark for the prover (single table or multi table)
- Add the benchmarks to ignored code on the line counter

PR:

- Discard the hashmap and the cache pilosophy
- Precompute everytime when calling prove single table (check if you need to keep adding twiddles to a "cached" vec of twiddles)
- Misses shouldn't happen (all things should be precomputed and shared)

- Bench against this

MauroToscano:
Implemented

PR #152, opened 2026-01-14
perf(stark): optimize FRI with fixed-size arrays and in-place folding

diegokingston:
- Add FieldElementPairBackend for fixed-size [FieldElement; 2] Merkle leaves
- Add fold_polynomial_doubled_inplace for zero-allocation polynomial folding
- Update FRI commit phase to use in-place folding
- Use FriLayerMerkleTree with pair backend to avoid Vec allocation per leaf

This reduces memory allocations during FRI commitment and improves cache locality.

### DIFF
--- a/crypto/crypto/src/merkle_tree/backends/field_element_vector.rs
+++ b/crypto/crypto/src/merkle_tree/backends/field_element_vector.rs
@@ -9,6 +9,53 @@ use math::{
     traits::AsBytes,
 };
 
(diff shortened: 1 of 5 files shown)


fmoletta (approved):
I added `#[allow(unused)]` to the non-inplace fold polynomial functions but we should consider removing them altogether as they are not accessible outside the fri module

MauroToscano (approved):

PR #162, opened 2026-01-15
feat(math): add SIMD-accelerated Goldilocks field arithmetic

diegokingston:
Add ARM NEON optimized field operations for Goldilocks prime field and its quadratic (Fp2) and cubic (Fp3) extensions.

Key features:
- PackedGoldilocks2: 2-wide SIMD for base field operations
- PackedFp2x2: 2-wide SIMD for quadratic extension (x^2 - 7)
- PackedFp3x2: 2-wide SIMD for cubic extension (x^3 - 2)
- Karatsuba multiplication for extension fields
- Scalar fallback for non-aarch64 architectures
- Comprehensive benchmarks comparing scalar vs SIMD performance

Performance improvements (Apple M3):
- Base field add/sub: 1.4-1.5x faster
- Fp2 multiplication: 1.35x faster
- Fp3 multiplication: 1.21x faster
- Batch operations: up to 1.76x faster

### DIFF
--- a/crypto/math/src/field/simd/packed_fp2.rs
+++ b/crypto/math/src/field/simd/packed_fp2.rs
@@ -1,0 +1,637 @@
+//! SIMD-accelerated Goldilocks quadratic extension (Fp2) arithmetic.
+//!
+//! This module provides `PackedFp2x2`, a type that holds 2 Fp2 elements
+//! and performs operations on them in parallel using SIMD operations.
+//!
+//! # Fp2 Field Structure
+//!
+//! Fp2 = Fp[x] / (x^2 - W) where W = 7 is the quadratic non-residue.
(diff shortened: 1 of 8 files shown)


crypto/math/src/field/simd/packed_fp2.rs
```diff
@@ -0,0 +1,637 @@
...
+        }
+    }
+
+    #[inline(always)]
+    pub fn neg(a: Self) -> Self {
+        Self {
+            real: Goldilocks64Field::neg(&a.real),
+            imag: Goldilocks64Field::neg(&a.imag),
+        }
+    }
```
fmoletta, line 68: Why were these implemented as associated functions rather than methods/trait impls. Is there a performance reason?

PR #164, opened 2026-01-15
spec: Initial version of memory argument ported

RobinJadoul:


spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+We can, however, statically bound the maximal number of memory accesses made during a single execution by a constant $k$
+and derive timestamps from the cycle counter.
+The $i$th possible memory access in cycle $c$ will obtain as timestamp the value $k dot c + i$.
+For simplicity, we will always reserve a timestamp for every possible memory access, and leave the timestamp unused if an instruction does not use it.
+
+#rj[Maybe we can find a nice way to show these notes/asides]
+*Note on “simultaneous” memory accesses*: For reasons of completeness (since temporal integrity as discussed below is a security necessity),
+we cannot deal with multiple accesses to the same address at identical timestamps.
+However, if multiple accesses are guaranteed to be independent (that is, to different addresses), they can still share a timestamp
+--- consider e.g. the case of reading a word as 4 bytes with a LOAD instruction.
```
erik-3milabs:
```suggestion
--- consider, e.g., the case of reading a word as 4 bytes with a LOAD instruction.
```

erik-3milabs: (At this point,) we have not discussed the "granularity" of an address. That is: whether an address indicates a byte, word or double word.

Without this clarified, the example is a touch confusing.

Also, there is no such thing as a `LOAD` instruction; it is `LD`, `LW[U]`, `LH[U]` or `LB[U]`. 

So, perhaps it should be something along the lines of 


```suggestion
However, if multiple accesses are guaranteed to be independent (that is, to different addresses), they can still share a timestamp
--- consider, e.g., the case of reading a word as 4 bytes with the `LW` load instruction.
```

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+This property is already taken into account where possible in the design of the system.
+
+== Permutation argument
+
+To ensure temporal integrity, every memory operation needs to be constrained for the newly emitted token
```
erik-3milabs, line 106: This sentence talks about "_the_ newly emitted token". But up to this point we have not even discussed the concept of a "token" in this context, let alone the concept of emitting / absorbing tokens.

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+to have a strictly greater timestamp than the consumed token.
+This raises the question of how to represent timestamps and cleanly perform this check,
+as over a finite field the “less than” relation is ill-defined
+(though it is common and natural to consider it as the less than relation over the natural lift of the field into the integers).
+We choose to represent timestamps as machine words, using the existing `LT` chip functionality for comparisons.
+#rj[Properly link/refer to the LT chip]
```
erik-3milabs, line 112: The "permutation argument" section only talks about the less-than check.
It feels to me like 
1. this section title is a misnomer, and
2. this should be a subsubsection at best, definitely not on the same level as the other subsections.

erik-3milabs, line 112: The "permutation argument" section only talks about the less-than check.

It feels to me like 
1. in it's current state, this section title is a misnomer.
2. if kept this way, this should be a subsubsection at best, definitely not on the same level as the other subsections.
3. I think we should (additionally) actually explain the permutation argument

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+Since we will have to ensure that memory accesses are temporally consistent within the execution of the VM,
+we additionally consider a _unique timestamp_ for every memory access, that should be strictly increasing.
+Multiple memory accesses can (and indeed will, consider e.g. register reads) occur in a single execution cycle of the VM,
+so we cannot use the cycle counter directly as timestamp for register accesses.
+We can, however, statically bound the maximal number of memory accesses made during a single execution by a constant $k$
+and derive timestamps from the cycle counter.
+The $i$th possible memory access in cycle $c$ will obtain as timestamp the value $k dot c + i$.
+For simplicity, we will always reserve a timestamp for every possible memory access, and leave the timestamp unused if an instruction does not use it.
+
+#rj[Maybe we can find a nice way to show these notes/asides]
```
erik-3milabs: Perhaps we can display these notes using a `#block` with a lightly coloured background and a border around it? That would clearly show that they're "something else" and may be considered as optional reading material.

Also, I'm not a fan of the double-column thing as it is right now. Why did you opt to do it this way?

If we are hell-bent on sticking with two columns, I would 
- make sure there is more empty space between the columns
- let each column have a similar structure. E.g.: 1. technique, 2. upsides, 3. downsides and clearly annotate each of these parts
- have subgrids, where all the text corresponding to each column is together, rather than interwoven as is the case now.

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+    - Range-checking the difference to be sufficiently small w.r.t. the field characteristic.
+  - Increments and basic arithmetic operations are cheap
+]
+
+#rj[reference to CPU chip/timestamp column and MEMW chip]
+
+== Initialization and Finalization
+
+Because the LogUp argument handling token consumption and emission needs the end-total to be fully balanced
+(that is, every token emitted should be consumed, and vice versa),
```
erik-3milabs:
```suggestion
Because the LogUp argument handling token consumption and emission needs to be fully balanced
--- every token emitted should be consumed, and vice versa ---
```

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+we need to have a system to emit the initial tokens and consume the final tokens.
+This needs to ensure that every address has at most a single initializing emission, and at most one finalizing consumption.
+Having at most one initialization will, through the correctness of the lookup argument,
+immediately lead to having at most one correct finalization, and vice versa.
+The initialization will need to correspond to a fixed initial register state for the VM,
+as well as the memory loaded from the program binary, zero-initialization of memory elsewhere, and private input provided by the prover.
+
+The contribution of initialization with static data from the ELF executable and the initial register state to the sum
```
erik-3milabs, line 141: I believe this sentence fits better in the next paragraph:

```suggestion
immediately lead to having at most one correct finalization, and vice versa.

The initialization will need to correspond to a fixed initial register state for the VM,
as well as the memory loaded from the program binary, zero-initialization of memory elsewhere, and private input provided by the prover.
The contribution of initialization with static data from the ELF executable and the initial register state to the sum
```

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+can be handled directly by the verifier, ensuring correctness corresponding to the ELF binary being proven.
+This leaves only zero-initialization and prover input as prover-side concerns for initialization,
+alongside the finalization of the entire used memory.
+
+For our chosen scheme (which we refer to as "paged initialization/finalization"),
+the available memory range is split into equally (power-of-two) sized "pages".
+Each address can then be represented as `address = page_base_address + page_offset`,
+with `page_base_address` being "page-aligned", and `page_offset` belonging to a limited range (the page size).
+As such, initialization or finalization of a page is represented by a table with columns `page`, `offset`, `value`, and ---for finalization--- `timestamp`.
```
erik-3milabs, line 150: The concept of `page` and `page_base_address` are used as synonyms here, while their names suggest they represent something entirely else:

`page` suggests an incrementing index (0, 1, 2, 3, 4, ...), such that `address := page_len * page + page_offset`, whereas `page_base_address` seems to capture `page_len * page` in one variable.

I propose we abandon one of the two terms in favour of the other.

Q: why did you arrive at `page_base_address` rather than `page`/`page_idx` where `address := page(_idx) * page_len + page_offset`? This, to me, feels a lot more natural than `page_base_address`, since we now need to make sure that `page_base_address = 0 mod page_len`.

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+The `page` column is a preprocessed, constant value (which can be entirely virtualized/inlined into the constraints for this table),
+and the `offset` column is a preprocessed column containing its row index.
+Depending on the type of initialization, `value` can be a prover-committed column, or a precomputed, constant column containing `0`.
```
erik-3milabs: Perhaps we should clarify the types. E.g.,

```suggestion
Depending on the type of initialization, `value` can be a prover-committed column (input data), or a precomputed, constant column containing `0` (free memory space).
```

or something of this sort

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+
+#let config = load_config()
+#let chip = load_chip("src/page.toml", config)
+
+#show: book-page.with(title: "Memory argument")
+
+= Memory argument
+
+As part of fully proving the correct execution of a RISC-V program,
+it is necessary to ensure that reads and writes to memory are consistent.
```
erik-3milabs: one cannot "read to".
```suggestion
it is necessary to ensure that memory reads and writes are consistent.
```

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+That is, every value read from some address corresponds to the value that was last written to that address
+--- or the initial value if nothing has been written yet.
+We consider "memory" in a broad sense here:
```
erik-3milabs, line 23:
```suggestion
--- or the initial value if nothing has been written yet.
In the following, we will outline how this VM ensures memory-access consistency.
We consider "memory" in a broad sense here:
```

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+Every memory operation has some conceptual attributes that are relevant to mention or discuss:
+
+- The type of operation (read or write)
+- The memory address --- this is an address in the broad sense:
+  main memory and registers have their own dedicated part of the unified address space.
+- The value being read from or written to the memory address
+- When the value was read or written, see the below paragraph
+
+Since we will have to ensure that memory accesses are temporally consistent within the execution of the VM,
+we additionally consider a _unique timestamp_ for every memory access, that should be strictly increasing.
```
erik-3milabs: 1. To me, the word "consider" in this context suggests that we are not 100% sure about whether we will add it or now. Given that we are confident we will, I propose we remove this word.
2. the phrase _unique_ seems to be too strong; the note on "simultaneous memory accesses" puts an asterisk on the term. As such, perhaps it would be better to omit the word here, and then discuss the restrictions to put on a timestamp (i.e., that a `(timestamp, memory type)` tuple should be unique, rather than _just_  the timestamp)

```suggestion
we must keep track of every memory access' _timestamp_, that should be strictly increasing.
```

Perhaps the concept of "strictly increasing" is more subtle and should be elaborated later on.

```suggestion
we must keep track of every memory access' _timestamp_.
```

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+Multiple memory accesses can (and indeed will, consider e.g. register reads) occur in a single execution cycle of the VM,
+so we cannot use the cycle counter directly as timestamp for register accesses.
+We can, however, statically bound the maximal number of memory accesses made during a single execution by a constant $k$
+and derive timestamps from the cycle counter.
```
erik-3milabs: as mentioned above, this discussion lacks some detail on domain separation. Observing the CPU chip, there are up to 5 memory accesses per cycle, while our `k` is only 4.

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+  - Clean definition of “less-than”, using the already existing `LT` functionality in the ALU
+  - Harder to perform increments, needing extra constraints beyond field arithmetic
+    - But this can be alleviated by providing a precomputed column that has a fixed increment per CPU row
+][
+  - Comparison is more annoying, but can work by:
+    - Decomposition into a machine word and chip interaction with the LT chip
+    - Bit decomposition and comparison constraints
+    - Range-checking the difference to be sufficiently small w.r.t. the field characteristic.
+  - Increments and basic arithmetic operations are cheap
+]
```
erik-3milabs: I believe the first column is not actually an alternative option: it is the option we chose. Why is it in here?

spec/src/page.toml
```diff
@@ -0,0 +1,57 @@
...
+
+[[constraints.all]]
+kind = "interaction"
+tag = "IS_BYTE"
+input = ["fini"]
+
+[[constraints.all]]
+kind = "interaction"
+tag = "memory"
+input = [0, "address", 0, "init"]
```
erik-3milabs, line 50: by setting the first var to `0`, you're hard coding this to only work for the RAM, but not the registers.

spec/src/page.toml
```diff
@@ -0,0 +1,57 @@
...
+multiplicity = -1
+
+[[constraints.all]]
+kind = "interaction"
+tag = "memory"
+input = [0, "address", "timestamp", "fini"]
+multiplity = 1
```
erik-3milabs:
```suggestion
multiplicity = 1
```

spec/memory.typ
```diff
@@ -0,0 +1,177 @@
...
+Since the `offset` column is always the same, it can be reused across all paged initialization and finalization tables.
+
+Concretely, each page gets an associated `PAGE` table, consisting of #total_nr_variables(chip) variables
+over #total_nr_instantiated_columns(chip, config) columns.
+For each such table, the `page` variable is instantiated as the constant base address of the page.
+The `offset` column is preprocessed, which helps the verifier ensure that each page has a single fixed size,
+but the verifier should still check that no pages overlap and all `page` values are page-aligned.
+
+#render_chip_column_table(chip, config)
+#render_constraint_table(chip, config)
```
erik-3milabs, line 176: perhaps we should add a bit more emphasis on the constraints. There is not even a header telling the reader what this is about.

erik-3milabs (commented):
Some comments.

spec/memory.typ
RobinJadoul: > Q: why did you arrive at page_base_address rather than page/page_idx where address := page(_idx) * page_len + page_offset? This, to me, feels a lot more natural than page_base_address, since we now need to make sure that page_base_address = 0 mod page_len.

The problem is that multiplying up the page index as a Word2L doesn't really work

spec/src/page.toml
RobinJadoul: Initializing registers shouldn't need a full page, as there are only a few of them (so let the verifier init them).
For finalization, we can probably integrate known register values into the HALT ecall, so the verifier can handle the logup there.

spec/memory.typ
RobinJadoul: I feel like that addition is kinda "obvious" in the context of a chapter on a memory argument

spec/memory.typ
RobinJadoul: That looks to be me skipping from one section to the next in copying stuff over from notion 🙈

spec/memory.typ
RobinJadoul: See the comment on my accidentally skipping things

RobinJadoul: Resolving this in favor of the more elaborate https://github.com/yetanotherco/lambda_vm/pull/164/changes#r2697515907

RobinJadoul: See the comment on my accidentally skipping things

spec/memory.typ
RobinJadoul: Maybe "alternatives" is not the optimal word here, but it was meant to be a side-by-side of the options

spec/memory.typ
RobinJadoul: 1. I think in an "unsure" meaning, it'd be "we are considering". 
    > 1. to think about carefully: such as [...] (b) to take into account
2. Agreed that the unique is probably too strong; but even `(timestamp, memory type)` is actually too strong for how we used it; I'll try to think of a way to deal with it

spec/memory.typ
RobinJadoul: The independence is actually dealt with in the below note, isn't it?

RobinJadoul: The derivation of `k = 4` itself (technically `k ≥ 3`) itself is, iirc, done in the CPU

spec/memory.typ
RobinJadoul: By ISA definition, memory would be byte-addressed, but I'll clarify this.

spec/memory.typ
RobinJadoul: The columns are mostly because it made sense on notion and I replicated that here.
I don't hold to them too strongly, but it is nice to have a side-by-side view of the things; and not have the aside take up too much vertical space.

spec/memory.typ
RobinJadoul: I also wouldn't tend to think of pages by their index, but mostly by their base address, even in a normal computing context.

RobinJadoul:
I've addressed some of the important things and replied to some remarks, so let's do another round :)

spec/memory.typ
```diff
@@ -0,0 +1,233 @@
...
+= Memory argument
+
+As part of fully proving the correct execution of a RISC-V program,
+it is necessary to ensure that memory reads and writes are consistent.
+That is, every byte read from some address corresponds to the byte that was last written to that address
+--- or the initial value if nothing has been written yet.
+We consider "memory" in a broad sense here:
+both RAM and the general purpose registers can be seen as instantiations of memory
+and are therefore handled simultaneously.#footnote[
+  While RAM is byte addressed, we do choose to store registers as a dword over two word addresses.
```
erik-3milabs: "dword" is just our shorthand notation for "Doubleword":


```suggestion
  While RAM is byte addressed, we do choose to store registers as a doubleword over two word addresses.
```

spec/src/page.toml
erik-3milabs: I see you added a clarification at the end. Thanks!

spec/memory.typ
erik-3milabs: The reason I propose to add this sentence, is that it sets the expectation for the reader: this is not going to be a discussion on the concept of memory arguments in the context of RISC-V (as the other sentences of this paragraph suggest, IMO) where any _application_ of the concept follows in another chapter. Rather, this chapter is all about the application.

Perhaps, it would suffice to rephrase the earlier sentence
> it is necessary to ensure that memory reads and writes are consistent.

as
> the VM must ensure that memory reads and writes are consistent.

It moves the discussion from abstract to application specific.

spec/memory.typ
```diff
@@ -0,0 +1,233 @@
...
+
+= Memory argument
+
+As part of fully proving the correct execution of a RISC-V program,
+it is necessary to ensure that memory reads and writes are consistent.
+That is, every byte read from some address corresponds to the byte that was last written to that address
+--- or the initial value if nothing has been written yet.
+We consider "memory" in a broad sense here:
+both RAM and the general purpose registers can be seen as instantiations of memory
+and are therefore handled simultaneously.#footnote[
```
erik-3milabs:
```suggestion
and are therefore handled simultaneously.
#footnote[
```

spec/memory.typ
```diff
@@ -0,0 +1,233 @@
...
+  While RAM is byte addressed, we do choose to store registers as a dword over two word addresses.
+]
+
+On a high level, we ensure memory consistency by an interacting system of
+reads and writes to a lookup argument, combined with an initialization and finalization scheme.
+The initialization and finalization schemes together ensure both that the necessary preconditions
+for the lookup system are satisfied, and that the program is executed with the correct
+initial memory and register contents as specified by the ELF binary and the ISA.
```
erik-3milabs: I prefer an itemization or enumeration (inline or multi-line) for sentences like these. Did you write it this way on purpose?

spec/memory.typ
```diff
@@ -0,0 +1,233 @@
...
+
+#aside[
+  *Note on “simultaneous” memory accesses*: For reasons of completeness (since temporal integrity as discussed below is a security necessity),
+  we cannot deal with multiple accesses to the same address at identical timestamps.
+  However, if multiple accesses are guaranteed to be independent (that is, to different addresses), they can still share a timestamp
+  --- consider, e.g., the case of reading a word as 4 bytes with the `LW` load instruction.
+  This property is already taken into account where possible in the design of the system.
+  For instance, in the CPU chip, we can ensure that there are at most 3 memory accesses not guaranteed
+  to be independent, so a timestamp granularity of 4 timestamps per cycle is enough.
+]
```
erik-3milabs: I think that 
1. having the header and content separate, and
2. only having a coloured background for some of the "aside" (e.g., the header only)

improves readability.

I was thinking of something like

Image

Wdyt?

Scrappy code you can use in case you think this is an improvement:
```typst
#let aside(title, body) = figure(
    block(
    stroke: black, 
    breakable: false, 
    radius: 5pt, 
    [
        #block(
            inset: (left: 1em, right: 1em, top: .75em, bottom: .75em), 
            fill: rgb("#b5dd94"), 
            stroke: black,
            width: 100%,
            radius: 5pt, 
            align(left, [#strong(title)])
        )
        #block(
         inset: (bottom : 1em, left: 1em, right: 1em), 
         breakable: false, 
        align(left, body)
        )
    ]
))
```

spec/memory.typ
```diff
@@ -0,0 +1,233 @@
...
+last written to memory at $serif("timestamp")$.
+Having exactly one value associated with any address will be ensured (see further down in this document)
+by the interaction of memory initialization, memory finalization, and the effects of memory operations.
+
+Each memory operation will then do two things:
+
+- Consume the current token in the memory
+- Emit a new token to replace it
+
+Naturally, for a read operation, the values embedded in the consumed and emitted tokens must be identical.
```
erik-3milabs: I've stumbled over this sentence so many times, that I think the following emphasis is worth adding:


```suggestion
Naturally, for a read operation, the _values_ embedded in the consumed and emitted tokens must be identical.
```

spec/memory.typ
```diff
@@ -0,0 +1,233 @@
...
+From the need to consume a token even on the first memory access,
+we can see the necessity for a memory initialization procedure
+---in addition to having to make sure the initial memory content lines up with what the binary dictates.
+
+Given that we can properly constrain temporal integrity (that is, no memory operation can consume future tokens),
```
erik-3milabs: I interpret the word "Given" to mean that something that is established (while temporal integrity is yet to be discussed). Perhaps the word "Assuming" is a better fit here? 🤔

spec/memory.typ
RobinJadoul: I have a habit from latex of avoiding spaces before the footnote marker, but typst _seems_ to handle that fine

spec/memory.typ
RobinJadoul: Doing a full enumeration felt rather overkill for two items, but inline might work

erik-3milabs (approved):
lgtm!

PR #171, opened 2026-01-15
Add tracing

gianbelinche:
This PR adds tracing to the executor, replacing prints.
It only leaves the syscalls prints since we want prints from riscv programs to normally appear on stdout


MauroToscano:
We don't need this dependency

PR #174, opened 2026-01-16
Add custom hasher for memory

gianbelinche:
This PR adds a custom hasher for the memory map.
The default hasher (SipHash) is cryptographically secure but slow.
For u32 memory addresses, security isn't needed

This improves the ethrex execution time from 56s to 33s a 41% improvement.


executor/src/vm/memory.rs
```diff
@@ -1,4 +1,42 @@
-use std::collections::BTreeMap;
+use std::collections::HashMap;
+use std::hash::{BuildHasher, Hasher};
+
+/// Fast hasher for u32 keys - uses the key directly as the hash value.
```
fmoletta:
```suggestion
/// Fast hasher for u64 keys - uses the key directly as the hash value.
```

fmoletta (approved):

PR #190, opened 2026-01-20
spec: placeholder chapters for chips to come

RobinJadoul:


erik-3milabs (approved):
Lgtm! What is your reason for introducing these now?

RobinJadoul:
There was a request for `a simple list of all the chips and their status (done, in progress, or to do)` on slack.
And these were the two I could think of that weren't likely to be merged by maybe today.

PR #195, opened 2026-01-21
spec: CPU padding

### LINKED ISSUE #136: spec: CPU padding
The padding value for the CPU chip may get a bit involved to work without needing an extra multiplicity column and associated product columns between multiplicity and functionality selector flags.
The conceptual design works together with the design of the instruction decoding, and embedding a NOP/self-loop into the decoding table,
so specifying it needs to be delayed until the decoding is ready.

RobinJadoul:
Adds the spec for CPU and DECODE padding rows.
The main idea is to let a HALT ecall (to come later) set all registers to 0 at timestamp 2^64 - 1, and have the CPU not execute anything, including register access to the PC.
This involves adding columns to prevent access to the x0 register; which should also decrease some of the pressure those accesses would already put on the MEMW table.

Closes #136

### DIFF
--- a/spec/cpu.typ
+++ b/spec/cpu.typ
@@ -6,6 +6,7 @@
   total_nr_variables,
   total_nr_instantiated_columns,
   render_constraint_table,
+  render_chip_padding_table,
 )
 
 #let config = load_config()
@@ -82,3 +83,12 @@ For @cpu:c:is_equal, refer to the logic of IsZero or IsEqual, in combination wit
 #render_constraint_table(chip, config, groups: "misc")
 
 #rj[Document the choice to not have a multiplicity column here for padding]
+
+== Padding
+
+The CPU can be padded with the following values, which have a corresponding row
+in the DECODE table, at the _odd_ address 1, only reachable through a HALT ecall.
+
+#render_chip_padding_table(chip, config)
+
+This approach minimizes the number of dependent lookups, increasing only multiplicities in the DECODE table and the IS_BYTE lookup.
--- a/spec/decode.typ
+++ b/spec/decode.typ
@@ -33,7 +33,8 @@ The #decode table is comprised of #nr_variables variables that are expressed usi
 The #decode table must be padded to a length that is a power of two.
 Empty rows with the following content can be added to achieve this:
 #render_chip_padding_table(chip, config)
-
+Note that this row sets the `EBREAK` flag.
+Given that `CPU` asserts that `EBREAK = 0` (see @cpu:c:ebreak_traps), using this "padding-instruction" would immediately make the CPU table unprovable.
 
 == Decoding
 For the purposes of explaining decoding, we decompress #decode's `packed_decode` variable into its constituent variables.
--- a/spec/decode.typ
+++ b/spec/decode.typ
@@ -53,7 +53,7 @@ We will illustrate how each instruction should be expressed in this (uncompresse
 The columns of the accompanying table represent the following:
 - *`operation`*: the assembly operation being encoded,
 - *`op-flag`*: which of the "`ALU` selector flags" operation flags to set. Each operation sets exactly one.
-- *`w_reg`*, *`w_instr`*, *`signed`*: whether to set the `write_register`, `word_instr` or `signed` flag, respectively,
+- *`w_instr`* and *`signed`*: whether to set the `word_instr` respectively `signed` flag,
 - *other*: the other flags that should be set or variables that should be given specific values.
 
 For the purpose of brevity and readability, the table uses the following rules-of-thumb:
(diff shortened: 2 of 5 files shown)


spec/src/cpu.toml
```diff
@@ -568,13 +642,27 @@ kind = "interaction"
 tag = "MEMW"
 input = [1, ["*", 2, "rs1"], "rv1", ["+", "timestamp", 0], 1, 0, 0]
 output = "rv1"
+multiplicity = "read_register1"
+
+[[constraints.mem]]
+kind = "arith"
+constraint = "`!read_register1` => `rv1[i]` = 0"
+poly = ["*", ["not", "register1"], ["idx", "rv1", "i"]]
```
```suggestion
poly = ["*", ["not", "read_register1"], ["idx", "rv1", "i"]]
```

spec/src/cpu.toml
```diff
@@ -568,13 +642,27 @@ kind = "interaction"
...
 kind = "interaction"
 tag = "MEMW"
 input = [1, ["*", 2, "rs2"], "rv2", ["+", "timestamp", 1], 1, 0, 0]
 output = "rv2"
+multiplicity = "read_register2"
+
+[[constraints.mem]]
+kind = "arith"
+constraint = "`!read_register2` => `rv2[i]` = 0"
+poly = ["*", ["not", "register2"], ["idx", "rv2", "i"]]
```
```suggestion
poly = ["*", ["not", "read_register2"], ["idx", "rv2", "i"]]
```

spec/src/cpu.toml
```diff
@@ -75,199 +98,238 @@ desc = """Multi-purpose selector used by different ALU operations for different
...
+[[variables.virtual]]
+name = "pad"
+type = "Bit"
+desc = "When no flags are set, we must be in a padding row."
+def = ["-", 1, "ADD", "SUB", "SLT", "AND", "OR", "XOR", "SHIFT", "JALR", "BEQ", "BLT", "LOAD", "STORE", "MUL", "DIVREM", "ECALL", "EBREAK"]
+
 
 [[assumptions]]
-desc = "The flags are a one-hot vector in the decoding"
+desc = "At most one flag is 1 by the decoding, and every other flag is 0"
```
erik-3milabs: Perhaps we should be more specific here, since the difference between `flag` and `bit` is a bit blurry?


```suggestion
desc = "At most one ALU selector flag is 1 by the decoding, and every other flag is 0."
```

spec/cpu.typ
```diff
@@ -82,3 +83,12 @@ For @cpu:c:is_equal, refer to the logic of IsZero or IsEqual, in combination wit
...
 #rj[Document the choice to not have a multiplicity column here for padding]
+
+== Padding
+
+The CPU can be padded with the following values, which have a corresponding row
+in the DECODE table, at the _odd_ address 1, only reachable through a HALT ecall.
+
+#render_chip_padding_table(chip, config)
+
+This approach minimizes the number of dependent lookups, increasing only multiplicities in the DECODE table and the IS_BYTE lookup.
```
erik-3milabs, line 94: We could make the `IS_BYTE` lookups run with multiplicity `1-pad`, right? Probably wouldn't save much in practice, though.

spec/src/cpu.toml
RobinJadoul: Yeah, previously it was just "flags" as well, I think, but this definitely works

spec/cpu.typ
RobinJadoul: Fair point, not sure what would be the more efficient thing in the long run, if it changes anything at all.

RobinJadoul: Actually, no; you might be able to get fake/weird padding rows

erik-3milabs: How would that work?

RobinJadoul: If the rsX/rd isn't range checked, you can collide different sums into the packed_decode sum
I suppose you can't change the ALU flags without re-introducing a range check, but it feels dangerous, and you might be able to change some other fields that don't contribute to `pad`

RobinJadoul:
Note for later: we may want to do a manual merge to keep some commits independent (like the x0 fast path)

spec/cpu.typ
erik-3milabs: > If the rsX/rd isn't range checked, you can collide different sums into the packed_decode sum

Wow, good thinking!

> I suppose you can't change the ALU flags without re-introducing a range check

These flags are `IS_BIT`-checked, right? I was just talking about `IS_BYTE`

RobinJadoul: > These flags are IS_BIT-checked, right? I was just talking about IS_BYTE

Yeah, I just meant you can't change those by offsetting the change with the non-byte-checked bytes; as then they become checked bytes.

spec/src/decode.toml
```diff
@@ -4,8 +4,7 @@ name = "DECODE"
 name = "pc"
 type = "DWordWL"
 desc = "value of the program counter this instruction is associated with."
-# TODO(#136): fix this when padding the CPU
-pad = 1
+pad = 0
```
RobinJadoul: I'd be inclined to still pad with an odd value here, just to avoid getting in the way of real instructions.
Yes, they probably shouldn't be at address zero, and it'd make the EBREAK (normally, if we did everything right) superfluous, but it doesn't hurt to be double cautious at no cost.

spec/src/decode.toml
```diff
@@ -43,7 +42,7 @@ A list of each variable and the bit(-range) in which it is located:\\
 [48] `EBREAK`; \\
 the remaining bits are set to zero.
 """
-pad = 0
+pad = ["^", 2, 48]
```
RobinJadoul: I think we still need to integrate the read_registerX into the decode table, so this'll become 2^50

RobinJadoul: And this was a different order of packing compared to the cpu

spec/src/cpu.toml
```diff
@@ -568,13 +642,27 @@ kind = "interaction"
 tag = "MEMW"
 input = [1, ["*", 2, "rs1"], "rv1", ["+", "timestamp", 0], 1, 0, 0]
 output = "rv1"
+multiplicity = "read_register1"
+
+[[constraints.mem]]
+kind = "arith"
+constraint = "`!read_register1` => `rv1[i]` = 0"
```
```suggestion
constraint = "$#`!read_register1` => #`rv1[i]` = 0$"
```

spec/src/cpu.toml
```diff
@@ -568,13 +642,27 @@ kind = "interaction"
...
 [[constraints.mem]]
 kind = "interaction"
 tag = "MEMW"
 input = [1, ["*", 2, "rs2"], "rv2", ["+", "timestamp", 1], 1, 0, 0]
 output = "rv2"
+multiplicity = "read_register2"
+
+[[constraints.mem]]
+kind = "arith"
+constraint = "`!read_register2` => `rv2[i]` = 0"
```
```suggestion
constraint = "$#`!read_register2` => #`rv2[i]` = 0$"
```

erik-3milabs (commented):
two things I forgot to mention yesterday

spec/src/decode.toml
erik-3milabs: Given that we've reserved `1` for decoding the CPU padding row, we might want to go with 7? The first odd number more than 4 steps away from 1?

RobinJadoul: Works for me

spec/decode.typ
```diff
@@ -33,7 +33,8 @@ The #decode table is comprised of #nr_variables variables that are expressed usi
 The #decode table must be padded to a length that is a power of two.
 Empty rows with the following content can be added to achieve this:
 #render_chip_padding_table(chip, config)
-
+Note that this row sets the `EBREAK` flag.
+Given that `CPU` asserts that `EBREAK = 0` (see @cpu:c:ebreak_traps), using this "padding-instruction" would immediately make the CPU table unprovable.
```
RobinJadoul, line 39: If we update pc padding to 7, quick sentence here to justify it

spec/src/decode.toml
RobinJadoul:
```suggestion
pad = 7
```

RobinJadoul (commented):
Can't click approve because I created the PR, but LGTM!

spec/decode.typ
erik-3milabs, line 212: @RobinJadoul Isn't the first syscall argument (also) in A0? Arguments are in A0-A7, and the return value is (also) written to A0.

On the source page, it moreover states:
> For 32-bit, every 64-bit integer argument will use 2 registers. For example, the system call uint64_t my_syscall(uint64_t a, uint64_t b, uint64_t c) would use 6 integer registers for its arguments (A0-A5), and 2 return registers (A0, A1).

which suggests to me that the first syscall arg should be A0

spec/decode.typ
erik-3milabs: Please double check that you agree with the justification

erik-3milabs:
@RobinJadoul please read through 697af3a. It contains some small fixes unrelated to any discussion we've had.

spec/decode.typ
RobinJadoul: I agree, but I don' t think the ref to cpu will work on the shiroa web build
(confirmed that `shiroa build` errors out on that)

spec/decode.typ
RobinJadoul: Also checking the `syscall(2)` manpage, yeah, seems like that was me misreading something somewhere.

RobinJadoul:
```suggestion
      - first syscall argument in A0 (= register `x10`)
```

spec/decode.typ
```diff
@@ -74,76 +80,76 @@ To indicate an instruction is provided in compressed form, the `c_type` flag is
...
-  ([`SH       rs1, rs2, imm`], [`STORE`], [], [], [], [`mem_2B`], []),
-  ([`SB       rs1, rs2, imm`], [`STORE`], [], [], [], [], []),
+  ([`SD       rs1, rs2, imm`], [`STORE`], [], [], [`mem_8B`], []),
+  ([`SW       rs1, rs2, imm`], [`STORE`], [], [], [`mem_4B`], []),
+  ([`SH       rs1, rs2, imm`], [`STORE`], [], [], [`mem_2B`], []),
+  ([`SB       rs1, rs2, imm`], [`STORE`], [], [], [], []),
   // ECALL/EBREAK
-  ([`ECALL`], [`ECALL`], [1], [], [], [$#`rs1` := #`x17`$, $#`rs2` := #`x11`$, $#`rd` := #`x10`$], [#ref_note(<note-ecall>)]),
-  ([`EBREAK`], [`EBREAK`], [], [], [], [], []),
+  ([`ECALL`], [`ECALL`], [], [], [$#`rs1` := #`x17`$, $#`rs2` := #`x11`$, $#`rd` := #`x10`$], [#ref_note(<note-ecall>)]),
```
RobinJadoul:
```suggestion
  ([`ECALL`], [`ECALL`], [], [], [$#`rs1` := #`x17`$, $#`rs2` := #`x10`$, $#`rd` := #`x10`$], [#ref_note(<note-ecall>)]),
```

spec/decode.typ
```diff
@@ -53,7 +53,7 @@ We will illustrate how each instruction should be expressed in this (uncompresse
 The columns of the accompanying table represent the following:
 - *`operation`*: the assembly operation being encoded,
 - *`op-flag`*: which of the "`ALU` selector flags" operation flags to set. Each operation sets exactly one.
-- *`w_reg`*, *`w_instr`*, *`signed`*: whether to set the `write_register`, `word_instr` or `signed` flag, respectively,
+- *`w_instr`* and *`signed`*: whether to set the `word_instr` respectively `signed` flag,
```
RobinJadoul:
```suggestion
+ - *`w_instr`*, *`signed`*: whether to set respectively the `word_instr` and `signed` flags
```

RobinJadoul:
Looks okay otherwise, I think

spec/decode.typ
RobinJadoul: Not sure it's the best rewrite yet, but something bothered my with the previous wording :)
At the least, there there probably be a comma before the respectively; maybe move that to before the `set` or even after everything else

PR #207, opened 2026-01-21
feat(executor): adjust initial log capacity

fmoletta:


fmoletta:
No conclusive improvements

PR #211, opened 2026-01-22
feat(executor): add segment-based instruction cache

fmoletta:
Replace current hashmap-based instruction cache with segment-based instruction cache, containing an ordered list of instruction segments. It uses binary search to find the segment for a given pc and then fetches the instruction directly.
(Misc) Add a couple more instruction-fetch heavy benchmarks
Performance improvement shows small improvements for computation-heavy benches and up to 20% improvement for more instruction-fetch heavy benchmarks (such as syscall_commit, keccak and hashmap)


gianbelinche (changes requested):
This seems a bit too complex for the optimizations it gives, we should try if a simpler approach gives similar results, for example, just having an unique array of instructions accessed by pc (or relative to pc, since instructions won't start at pc 0)

gianbelinche (approved):

MauroToscano (approved):

PR #212, opened 2026-01-22
  refactor(prover): centralize trace generation in TraceBuilder module

ColoCarletti:
Refactors trace generation to use a centralized `Traces` struct that generates all table traces from
  execution logs in a single pass. This prepares the codebase for adding more tables (DECODE, SHIFT, MUL,
  MEMW, etc.) without cluttering the CPU table module.


prover/src/tables/trace_builder.rs
```diff
@@ -0,0 +1,73 @@
...
+use super::lt::{self, LtOperation};
+use super::types::{GoldilocksExtension, GoldilocksField};
+
+/// All generated trace tables.
+pub struct Traces {
+    /// CPU execution trace (one row per instruction)
+    pub cpu: TraceTable<GoldilocksField, GoldilocksExtension>,
+
+    /// BITWISE precomputed lookup table (2^20 rows)
+    pub bitwise: TraceTable<GoldilocksField, GoldilocksExtension>,
```
diegokingston, line 29: As a side note, bitwise table may need to be in a separate category, since this is known even before you execute the program (you could already have it predefined or even make it fully virtual from the provers perspective)

prover/src/tables/trace_builder.rs
```diff
@@ -0,0 +1,73 @@
...
+            }
+
+            cpu_ops.push(op);
+        }
+
+        // Generate CPU trace (handles padding internally)
+        let cpu = cpu::generate_cpu_trace(&cpu_ops);
+
+        // Generate BITWISE trace and update multiplicities
+        let mut bitwise = bitwise::generate_bitwise_trace();
```
diegokingston, line 65: This is the only column you need, since the fingerprints can be computed from the information from other tables

MauroToscano (approved):

prover/src/tables/trace_builder.rs
MauroToscano: We will add something to check this later, but true, it should have a different behaviour

prover/src/tables/trace_builder.rs
MauroToscano: We can change this later

PR #219, opened 2026-01-22
feat(executor): pre-allocate memory

fmoletta:


fmoletta:
No measurable improvement

PR #220, opened 2026-01-23
spec: DVRM chip

erik-3milabs:
This PR introduces the `DVRM` chip.

Note: this PR additionally widens the `ZERO` lookup from 16 to 20 bits. (see 582bc35)

### DIFF
--- a/spec/book.typ
+++ b/spec/book.typ
@@ -9,6 +9,7 @@
   summary: (
     ("memory.typ", [Memory argument], <memory>),
     ("variables.typ", [Variables], <vars>),
+    ("signatures.typ", [Signatures], <signatures>),
     ("is_bit.typ", [IS_BIT template], <isbit>),
     ("sign.typ", [SIGN template], <sign>),
     ("add.typ", [ADD/SUB template], <add>),
--- a/spec/signatures.typ
+++ b/spec/signatures.typ
@@ -1,0 +1,90 @@
+#import "/book.typ": book-page
+#import "/src.typ": load_signatures, load_config
+
+#show: book-page("signatures.typ")
+
+#let config = load_config()
+#let signatures = load_signatures(config)
+
+// Render a signature
+#let render_signature(sig) = {
+  let (lb, rb) = if sig.kind == "interaction" {
+    (`[`, `]`)
+  } else if sig.kind == "template" {
+    (`<`, `>`)
+  }
+
+  let cond = sig.at("cond", default: none)
+  let cond_str = if cond != none {
+    raw(cond) + ` => `
+  } else {``}
+
+  let input_str = sig.input.map(elt => {
+    if type(elt) == array {
+      raw(elt.at(0)) + `[` + raw(str(elt.at(1))) + `]`
+    } else {
+      raw(elt)
+    }
+  }).join(`, `)
+
+  let output = sig.at("output", default: none)
+  let output_str = if output != none {
+    if type(output) == array {
+      raw(output.at(0)) + `[` + raw(str(output.at(1))) + `]`
+    } else {
+      raw(output)
+    } + `; `
+  } else {``}
+
+  return [#cond_str#raw(sig.tag)#lb#output_str#input_str#rb]
+}
+
+// Compute the bus size of an interaction
+#let interaction_bus_size(sig) = {
+  let vars = sig.input + if "output" in sig { (sig.output, )} else {()}
+
+  return vars.map(v => {
+    let (label, factor) = if type(v) == array {
+      (v.at(0), v.at(1))
+    } else {
+      (v, 1)
+    }
+    config.variables.types.filter(type => type.label == label).first().subtypes.len() * factor
+  })
+  .sum()
+}
+
+#let interactions = signatures.signatures.filter(s => s.kind == "interaction")
+The following lists signatures of the #interactions.len() interactions in this VM.
+#figure(
+  table(
+    columns: (1fr, auto),
+    inset: 7pt,
+    align: (top+left, center),
+    stroke: none,
+    table.header([*Signature*], [*Bus size*]),
+    table.hline(stroke: 1pt),
+    table.vline(stroke: 1pt, x: 1),
+    ..for sig in interactions {
+      ([#render_signature(sig)], [#interaction_bus_size(sig)])
+    },
+  ),
+  caption: "Signature overview of interactions",
+)
+
+#let templates = signatures.signatures.filter(s => s.kind == "template")
+Below, we list the signatures of the #templates.len() templates in this VM.
+#figure(
+  table(
+    columns: 1fr,
+    inset: 7pt,
+    align: (top+left, center),
+    stroke: none,
+    table.header([*Signature*]),
+    table.hline(stroke: 1pt),
+    ..for sig in templates {
+      ([#render_signature(sig)], )
+    },
+  ),
+  caption: "Signature overview of templates",
+)
--- a/spec/src.typ
+++ b/spec/src.typ
@@ -1,5 +1,7 @@
 /// Path to the config file.
 #let CONFIG_PATH = "src/config.toml"
+/// Path to the signatures file
+#let SIGNATURES_PATH = "src/signatures.toml"
 
 /// Check the configuration object for internal consistency.
 #let _check_config(config) = {
@@ -30,6 +32,55 @@
   _check_config(config)
   return config
 }
+
+
+// Validate the `signatures` overview
+#let _check_signatures(signatures, config) = {
+  let var_labels = config.variables.types.map(t => t.label)
+
+  // Verify that `var` is a valid variable.
+  let verify_variable(var) = {
+    if type(var) == array {
+        assert(var.at(0) in var_labels, message: "Invalid var type: " + repr(var))
+        assert(type(var.at(1)) == int, message: "Invalid var type: " + repr(var))
+    } else if type(var) == str {
+      assert(var in var_labels, message: "Invalid var type: " + repr(var))
+    } else {
+      assert(false, message: "Invalid var type: " + repr(var))
+    }
+  }
+
+  assert("signatures" in signatures, message: "No signatures listed")
+  for sig in signatures.signatures {
+    assert("tag" in sig, message: "No tag associated with " + repr(sig))
+    assert(type(sig.tag) == str, message: "Tag is not of type str: " + repr(sig.tag))
+
+    assert("kind" in sig, message: "No kind associated with " + repr(sig))
+    assert(type(sig.kind) == str, message: "kind is not of type str: " + repr(sig.kind))
+    assert(sig.kind in ("interaction", "template"), message: "Invalid kind: " + repr(sig.kind))
+
+    if "cond" in sig {
+      assert(sig.kind != "interaction", message: "Invalid condition for interaction: " + repr(sig))
+      verify_variable(sig.cond)      
+    }    
+    
+    assert("input" in sig, message: "No input associated with " + repr(sig))
+    assert(type(sig.input) == array, message: "Invalid input type: " + repr(sig.input))
+    sig.input.map(i => verify_variable(i))
+
+    if "output" in sig {
+      verify_variable(sig.output)
+    }
+  }
+}
+
+// Load the signatures from file
+#let load_signatures(config) = {
+  let signatures = toml(SIGNATURES_PATH)
+  _check_signatures(signatures, config)
+  return signatures
+}
+
 
 /// Check a chip object for internal consistency.
 #let _check_chip(chip, config) = {
--- a/spec/src/signatures.toml
+++ b/spec/src/signatures.toml
@@ -1,0 +1,178 @@
+# cond => IS_BIT<X>
+[[signatures]]
+tag = "IS_BIT"
+kind = "template"
+input = ["BaseField"]
+cond = "BaseField"
+
+# cond => ADD<sum; lhs, rhs>
+[[signatures]]
+tag = "ADD"
+kind = "template"
+input = ["DWordWL", "DWordWL"]
+output = "DWordWL"
+cond = "BaseField"
+
+# cond => SUB<diff; lhs, rhs>
+[[signatures]]
+tag = "SUB"
+kind = "template"
+input = ["DWordWL", "DWordWL"]
+output = "DWordWL"
+cond = "BaseField"
+
+# cond => NEG<neg; X>
+[[signatures]]
+tag = "NEG"
+kind = "template"
+input = ["DWordHL"]
+output = "DWordWL"
+cond = "Bit"
+
+# SIGN<sign; X, signed>
+[[signatures]]
+tag = "SIGN"
+kind = "template"
+input = ["Half", "Bit"]
+output = "Bit"
+
+# DECODE[pc, imm, packed_decode]
+[[signatures]]
+tag = "DECODE"
+kind = "interaction"
+input = ["DWordWL", "DWordWL", "BaseField"]
+
+# SHIFT[out; in, shift, direction, signed, word_instr]
+[[signatures]]
+tag = "SHIFT"
+kind = "interaction"
+input = ["DWordHL", "Byte", "Bit", "Bit", "Bit"]
+output = "DWordWL"
+
+# BRANCH[next_pc; pc, offset, register, JALR]
+[[signatures]]
+tag = "BRANCH"
+kind = "interaction"
+input = ["DWordWL", "Word", "DWordWL", "Bit"]
+output = "DWordWL"
+
+# MEMW[old; is_register, base_address, value, timestamp, write2, write4, write8]
+[[signatures]]
+tag = "MEMW"
+kind = "interaction"
+input = ["Bit", "DWordWL", ["BaseField", 8], "DWordWL", "Bit", "Bit", "Bit"]
+output = ["BaseField", 8]
+
+# MEMW[is_register, base_address, value, timestamp, write2, write4, write8]
+[[signatures]]
+tag = "MEMW"
+kind = "interaction"
+input = ["Bit", "DWordWL", ["BaseField", 8], "DWordWL", "Bit", "Bit", "Bit"]
+
+# LT[lt; lhs, rhs, signed]
+[[signatures]]
+tag = "LT"
+kind = "interaction"
+input = ["DWordWL", "DWordWL", "Bit"]
+output = "Bit"
+
+# MUL[lo/hi; lhs, lhs_signed, rhs, rhs_signed, 0/1]
+[[signatures]]
+tag = "MUL"
+kind = "interaction"
+input = ["DWordHL", "Bit", "DWordHL", "Bit", "Bit"]
+output = "DWordWL"
+
+# DVRM[q/r; n, d, signed, 0/1]
+[[signatures]]
+tag = "DVRM"
+kind = "interaction"
+input = ["DWordHL", "DWordHL", "Bit", "Bit"]
+output = "DWordWL"
+
+# LOAD[res; base_address, timestamp, read2, read4, read8, signed]
+[[signatures]]
+tag = "LOAD"
+kind = "interaction"
+input = ["DWordWL", "DWordWL", "Bit", "Bit", "Bit", "Bit"]
+output = "DWordWL"
+
+# ECALL[timestamp, syscallnr]
+[[signatures]]
+tag = "ECALL"
+kind = "interaction"
+input = ["DWordWL", "DWordWL"]
+
+# AND_BYTE[res; X, Y]
+[[signatures]]
+tag = "AND_BYTE"
+kind = "interaction"
+input = ["Byte", "Byte"]
+output = "Byte"
+
+# OR_BYTE[res; X, Y]
+[[signatures]]
+tag = "OR_BYTE"
+kind = "interaction"
+input = ["Byte", "Byte"]
+output = "Byte"
+
+# XOR_BYTE[res; X, Y]
+[[signatures]]
+tag = "XOR_BYTE"
+kind = "interaction"
+input = ["Byte", "Byte"]
+output = "Byte"
+
+# MSB8[msb; X]
+[[signatures]]
+tag = "MSB8"
+kind = "interaction"
+input = ["Byte"]
+output = "Bit"
+
+# MSB16[msb; X]
+[[signatures]]
+tag = "MSB16"
+kind = "interaction"
+input = ["Half"]
+output = "Bit"
+
+# ZERO[is_zero; X]
+[[signatures]]
+tag = "ZERO"
+kind = "interaction"
+input = ["B20"]
+output = "Bit"
+
+# IS_BYTE[X]
+[[signatures]]
+tag = "IS_BYTE"
+kind = "interaction"
+input = ["Byte"]
+
+# IS_HALF[X]
+[[signatures]]
+tag = "IS_HALF"
+kind = "interaction"
+input = ["Half"]
+
+# IS_B20[X]
+[[signatures]]
+tag = "IS_B20"
+kind = "interaction"
+input = ["B20"]
+
+# HWSL[res; X, shift]
+[[signatures]]
+tag = "HWSL"
+kind = "interaction"
+input = ["Half", "B4"]
+output = "Half"
+
+# HWSLC[res; X, shift]
+[[signatures]]
+tag = "HWSLC"
+kind = "interaction"
+input = ["Half", "B4"]
+output = "Half"


erik-3milabs:
Perhaps 8652c0f and c8d6896 should be merged separately?

RobinJadoul:
Probably 5c983c980e4ffb3593f243566718465b1c21786c rather than 8652c0f6e409c9ea3fa4bdf13addd755ec397cc4, but yeah, that could make sense

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+pad = 0
+
+# Virtual
+
+[[variables.virtual]]
+name = "extended_n_sub_r"
+type = ["Half", 8]
+desc = "sign-extended value of `n_sub_r`."
+def = {idx="i", polys = [
+  {range=[0, 3], poly=["idx", "n_sub_r", "i"]},
```
RobinJadoul:
```suggestion
  {iter=[0, 3], poly=["idx", "n_sub_r", "i"]},
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+  {range=[4, 7], poly=["*", 0xFFFF, "sign_n_sub_r"]}
```
RobinJadoul:
```suggestion
  {iter=[4, 7], poly=["*", 0xFFFF, "sign_n_sub_r"]}
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+type = "Bit"
+desc = ""
+pad = 0
+
+
+# Assumptions
+
+[[assumptions]]
+desc = "`IS_HALF[n[i]]`"
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
iter = ["i", 0, 3]
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+ref = "lt:a:range_n"
+
+[[assumptions]]
+desc = "`IS_HALF[d[i]]`"
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
iter = ["i", 0, 3]
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+ref = "lt:a:range_d"
+
+[[assumptions]]
+desc = "`IS_BIT<signed>`"
+ref = "lt:a:range_signed"
+
+[[assumptions]]
+desc = "`IS_HALF[q[i]]`"
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
iter = ["i", 0, 3]
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+ref = "lt:a:range_q"
+
+[[assumptions]]
+desc = "`IS_HALF[r[i]]`"
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
iter = ["i", 0, 3]
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+desc = "`IS_HALF[d[i]]`"
+range = ["i", 0, 3]
+ref = "lt:a:range_d"
+
+[[assumptions]]
+desc = "`IS_BIT<signed>`"
+ref = "lt:a:range_signed"
+
+[[assumptions]]
+desc = "`IS_HALF[q[i]]`"
```
RobinJadoul: As already discussed, we can't rely on this because there's only one output

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+ref = "dvrm:c:sign_r_equals_sign_n"
+
+[[constraint_groups]]
+name = "abs_diff"
+
+[[constraints.abs_diff]]
+kind = "interaction"
+tag = "LT"
+input = ["abs_r", "abs_d", 0]
+output = ["not", "div_by_zero"]
```
RobinJadoul, line 219: I think this breaks down for 0/0

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+multiplicity = "μ_sum"
+ref ="dvrm:c:abs_r_lt_abs_d"
+
+[[constraints.abs_diff]]
+kind = "interaction"
+tag = "IS_HALF"
+input = [["idx", "abs_r", "i"]]
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
iter = ["i", 0, 3]
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+tag = "SUB"
+input = [0, "r"]
+output = "abs_r"
+cond = "sign_r"
+ref = "dvrm:c:abs_r_if_negative"
+
+[[constraints.abs_diff]]
+kind = "arith"
+constraint = "$not#`sign_r` => #`abs_r[i]`=#`r[i]`$"
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
iter = ["i", 0, 3]
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+poly = ["*", ["-", 1, "sign_r"], ["-", ["idx", "abs_r", "i"], ["idx", "r", "i"]]]
+ref = "dvrm:c:abs_r_if_nonnegative"
+
+[[constraints.abs_diff]]
+kind = "interaction"
+tag = "IS_HALF"
+input = [["idx", "abs_d", "i"]]
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
iter = ["i", 0, 3]
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+tag = "SUB"
+input = [0, "d"]
+output = "abs_d"
+cond = "sign_d"
+ref = "dvrm:c:abs_d_if_negative"
+
+[[constraints.abs_diff]]
+kind = "arith"
+constraint = "$not#`sign_d` => #`abs_d[i]`=#`d[i]`$"
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
iter = ["i", 0, 3]
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+poly = ["*", ["-", 1, "sign_d"], ["-", ["idx", "abs_d", "i"], ["idx", "d", "i"]]]
+ref = "dvrm:c:abs_d_if_nonnegative"
+
+[[constraint_groups]]
+name = "overflow"
+
+[[constraints.overflow]]
+kind = "arith"
+range = ["i", 0, 3]
```
RobinJadoul:
```suggestion
iter = ["i", 0, 3]
```

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+ref = "dvrm:c:n_sub_r_range"
+
+[[constraint_groups]]
+name = "equality"
+
+[[constraints.equality]]
+kind = "interaction"
+tag = "MUL"
+input = ["d", "signed", "q", "sign_q", 0]
+output = ["idx", "extended_n_sub_r", "0:4"]
```
RobinJadoul: Still not loving slices in the "machine-readable" spec, maybe there's some casting solution again

spec/src/dvrm.toml
```diff
@@ -0,0 +1,397 @@
...
+kind = "template"
+tag = "IS_BIT"
+input = ["μ_q"]
+ref = "dvrm:c:is_bit_μ_q"
+
+[[constraints.defs]]
+kind = "template"
+tag = "IS_BIT"
+input = ["μ_r"]
+ref = "dvrm:c:is_bit_μ_r"
```
RobinJadoul: Do we need these?

spec/dvrm.typ
```diff
@@ -5,13 +5,159 @@
...
+
+Leveraging R1 #footnote([Note: we need not worry about the _overflow_ case in applying this relation, since R5 requires specific values for `q` and `r` in this case.]), we can rewrite these as
++ the sign of $#`r`$ must match that of `n` (unless $#`r` = 0$), and 
++ $|#`r`|  < |#`d`|$ (unless $#`d` = 0$).
+
+Focusing on the first statement, we observe that this trivially holds when $#`signed` = 0$, 
+while R3 deals with the case that $#`signed` = 0$.
+The second statement is enforced by @dvrm:c:abs_r_lt_abs_d.
+@dvrm:c:abs_r_if_negative and @dvrm:c:abs_r_if_nonnegative, respectively @dvrm:c:abs_d_if_negative and @dvrm:c:abs_d_if_nonnegative are included to ensure that `abs_r` and `abs_d` are the absolute values of `r` respectively `d`.
+@dvrm:c:abs_r_range_check and @dvrm:c:abs_d_range_check are required to uphold assumption @add:a:lhs required by the `SUB` chip.
```
RobinJadoul:
```suggestion
@dvrm:c:abs_r_if_negative and @dvrm:c:abs_r_if_nonnegative (respectively @dvrm:c:abs_d_if_negative and @dvrm:c:abs_d_if_nonnegative) is included to ensure that `abs_r` (`abs_d`) is the absolute values of `r` (respectively `d`).
@dvrm:c:abs_r_range_check and @dvrm:c:abs_d_range_check are required to uphold assumption @add:a:lhs required by the `SUB` chip.
```
I'd personally probably even shorten it to "resp." everywhere

spec/dvrm.typ
```diff
@@ -5,13 +5,159 @@
...
+In other words,
++ the sign of $#`n`-#`qd`$ must match that of `n` (unless $#`qd` = #`n`$), and 
++ $|#`n`-#`qd`|  < |#`d`|$ (unless $#`d` = 0$).
+
+Leveraging R1 #footnote([Note: we need not worry about the _overflow_ case in applying this relation, since R5 requires specific values for `q` and `r` in this case.]), we can rewrite these as
++ the sign of $#`r`$ must match that of `n` (unless $#`r` = 0$), and 
++ $|#`r`|  < |#`d`|$ (unless $#`d` = 0$).
+
+Focusing on the first statement, we observe that this trivially holds when $#`signed` = 0$, 
+while R3 deals with the case that $#`signed` = 0$.
```
RobinJadoul:
```suggestion
while R3 deals with the case that $#`signed` = 1$.
```

spec/dvrm.typ
```diff
@@ -5,13 +5,159 @@
...
+The second statement is enforced by @dvrm:c:abs_r_lt_abs_d.
+@dvrm:c:abs_r_if_negative and @dvrm:c:abs_r_if_nonnegative, respectively @dvrm:c:abs_d_if_negative and @dvrm:c:abs_d_if_nonnegative are included to ensure that `abs_r` and `abs_d` are the absolute values of `r` respectively `d`.
+@dvrm:c:abs_r_range_check and @dvrm:c:abs_d_range_check are required to uphold assumption @add:a:lhs required by the `SUB` chip.
+
+#render_constraint_table(chip, config, groups:("abs_diff", ))
+
+=== R5: overflow
+The ISA requires that $#`q` = #`n`$ and $#`r` = 0$ in the event of overflow.
+We note that, while $#`n` = #`qd` + #`r`$ (R1) does _not_ hold in the case of overflow, the relation $#`n` = |#`q`|#`d` + #`r`$ _does_.
+We moreover note that the _signed_ two's complement representation of `q` is identical to the _unsigned_ representation of $|#`q`|$.
```
RobinJadoul:
```suggestion
We moreover note that the _signed_ two's complement representation of $-2^63$ is identical to the _unsigned_ representation of $|-2^63| = 2^63$.
```

RobinJadoul: I feel like this makes it more explicit that it's only for this specific value of q; the one we care about for overflow

spec/dvrm.typ
```diff
@@ -5,13 +5,159 @@
...
+While the `IsEqual` template can handle the first and third equality directly, the second equality was rewritten as the set of equalities
+$
+  #`n[`i#`]` &= 0 "for" i in [0, 2],\
+  #`n[`3#`]` - 2^15 dot #`msb_n` &= 0,\
+  #`msb_n` &= 1
+$
+where we note that $#`n[3]` - 2^15 dot #`msb_n` = 0 <=> #`n[3]` equiv 0 mod 2^15$. 
+Hence, the last two equalities are satisfied if and only if $#`n[3]` = 2^15$.
+
+=== R1: $#`n` = #`qd` + #`r`$
```
RobinJadoul: Still checking this further

spec/src/dvrm.toml
RobinJadoul: Or maybe not, because 0 < 0 is still false...

RobinJadoul: Or maybe not, because 0 < 0 is still false...

spec/src/dvrm.toml
erik-3milabs: we could, alternatively, have 
1. `n_sub_r: DWordHL` on the first and `n_sub_r_extension: DWordHL` on the second lookup, or
2. cast `extended_n_sub_r` as `DWordHL[2]` and then index 0 and 1

where I think the first option might be better

spec/src/dvrm.toml
erik-3milabs: :)

spec/src/dvrm.toml
erik-3milabs: I think you're right: we don't

spec/src/dvrm.toml
RobinJadoul: I agree

erik-3milabs:
@greptileai

spec/bitwise.typ
```diff
@@ -38,5 +38,5 @@ The following ideas may prove to be optimizations for the #bitwise chip:
 + Drop `MSB8` column, and instead define the `MSB8` lookup as `MSB8<X> := MSB16[256X]`.
   Note: currently, `MSB8` also implicity range checks the input `X` (the lookup fails if `X` is not a `Byte`).
   This optimization should only be executed when all chips leveraging `MSB8` do _not_ need this implicit range check.
-+ Place the 16-bit (`AND`, `OR`, `XOR`, `MSB16`, `ZERO`, etc.) and 20-bit (`HWSL`, `HWSLC`, `IS_B20`) lookups in separate tables.
++ Place the 16-bit (`AND`, `OR`, `XOR`, `MSB16`, `ZERO`, etc.) and 20-bit (`HWSL`, `HWSLC`, `IS_B20`, `ZERO`) lookups in separate tables.
```
RobinJadoul:
```suggestion
+ Place the 16-bit (`AND`, `OR`, `XOR`, `MSB16`, etc.) and 20-bit (`HWSL`, `HWSLC`, `IS_B20`, `ZERO`) lookups in separate tables.
```

spec/chip.typ
```diff
@@ -156,7 +163,7 @@
       (table.header(level:2, table.cell(colspan: 4, emph(cat))), table.hline(stroke: .6pt))
       for var in vars {
         (
-          [#raw(var.name)], 
+          [#raw(var.name)#lbl(var)],
```
RobinJadoul: This will fail once you actually have a reference. You need some kind of heading or figure to attach the label to, just a `raw` isn't enough.

spec/dvrm.typ
```diff
@@ -5,13 +5,139 @@
...
+
+Leveraging R1 #footnote([Note: we need not worry about the _overflow_ case in applying this relation, since R5 requires specific values for `q` and `r` in this case.]), we can rewrite these as
++ the sign of $#`r`$ must match that of `n` (unless $#`r` = 0$), and 
++ $|#`r`|  < |#`d`|$ (unless $#`d` = 0$).
+
+Focusing on the first statement, we observe that this trivially holds when $#`signed` = 0$, 
+while R3 deals with the case that $#`signed` = 1$.
+The second statement is enforced by @dvrm:c:abs_r_lt_abs_d.
+@dvrm:c:abs_r_if_negative and @dvrm:c:abs_r_if_nonnegative (resp. @dvrm:c:abs_d_if_negative and @dvrm:c:abs_d_if_nonnegative) are included to ensure that `abs_r` (resp. `abs_d`) is the absolute values of `r` (resp. `d`).
+@dvrm:c:abs_r_range_check and @dvrm:c:abs_d_range_check are required to uphold assumption @add:a:lhs required by the `SUB` chip.
```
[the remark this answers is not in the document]
erik-3milabs: While, indeed, we're referencing a constraint in the `ADD` chip when talking about the `SUB` chip, this is correct: `SUB` is not a chip, but a "macro" leveraging the `ADD` chip. There is no such thing as `@sub:a:lhs`.

spec/dvrm.typ
```diff
@@ -5,13 +5,139 @@
...
+Rather, we need only store the lower 64-bits, and a separate bit (`sign_n_sub_r`) indicating whether the top limbs are all-ones or all-zeroes.
+The prover is free to select the value for `sign_n_sub_r`; only one of the two will fit the proof.
+
+#render_constraint_table(chip, config, groups:("n_sub_r", ))
+
+== R4: division-by-zero
+R4 requires that $#`q` = 2^64-1$ (unsigned) or $-1$ (signed) and $#`r` = n$ when $#`d` = 0$.
+Recalling R1, we see that $#`n` = #`q` #`d` + #`r` = #`r`$ when $#`d` = 0$, already enforces the latter.
+Next, we note that, in two's complement, the _unsigned_ value $2^64-1$ and _signed_ value $-1$ are both represented by the bit string `0xFFFFFFFF`.
+Hence, only @dvrm:c:q_if_div_by_zero is required to completely constrain R5; @dvrm:c:div_by_zero just ensures the `div_by_zero` flag is set when $#`d` = 0$.
```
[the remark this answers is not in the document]
erik-3milabs: Please propose a fix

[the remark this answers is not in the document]
erik-3milabs: already done

spec/src/dvrm.toml
```diff
@@ -0,0 +1,380 @@
...
+desc = "sign-extended value of `n`."
+def = {idx="i", polys = [
+  {iter=[0, 3], poly=["idx", "n", "i"]},
+  {iter=[4, 7], poly=["*", 0xFFFF, "sign_n"]}
+]}
+ref = "dvrm:v:extended_n"
+
+[[variables.virtual]]
+name = "extended_r"
+type = ["Half", 8]
```
RobinJadoul: Does it maybe make sense to make these `QuadHL`?

spec/dvrm.typ
```diff
@@ -5,13 +5,139 @@
   total_nr_variables,
   total_nr_instantiated_columns,
   render_constraint_table,
-  render_chip_assumptions,
   render_chip_padding_table,
+  render_chip_assumptions
 )
 
+#set footnote(numbering: "[1]")
```
RobinJadoul: Any specific reason for this?
If so, probably better to move that up to the book.typ or ebook.typ level, since now it's suddenly different from how it looks in memory.typ

spec/src/dvrm.toml
```diff
@@ -0,0 +1,380 @@
...
+name = "n_sub_r"
+
+[[constraints.n_sub_r]]
+kind = "template"
+tag = "SUB"
+input = [["cast", "n", "DWordWL"], ["cast", "r", "DWordWL"]]
+output = ["cast", "n_sub_r", "DWordWL"]
+ref = "dvrm:c:n_sub_r"
+
+[[constraints.n_sub_r]]
```
RobinJadoul, line 296: I suppose we're waiting on #270 to optimize?

spec/src/dvrm.toml
```diff
@@ -0,0 +1,380 @@
...
+tag = "ZERO"
+input = [["+", ["idx", "n", 0], ["idx", "n", 1], ["idx", "n", 2], ["-", ["idx", "n", 3], ["*", ["^", 2, 15], "sign_n"]], ["-", 1, "sign_n"], ["-", 65535, ["idx", "d", 0]], ["-", 65535, ["idx", "d", 1]], ["-", 65535, ["idx", "d", 2]], ["-", 65535, ["idx", "d", 3]]]]
+output = "overflow"
+multiplicity = "μ_sum"
+ref = "dvrm:c:overflow"
+
+[[constraint_groups]]
+name = "n_sub_r"
+
+[[constraints.n_sub_r]]
```
RobinJadoul, line 289: I do think we need to do the SUB over the Quads, otherwise `sign_n_sub_r` is in no way constraint besides being a bit

RobinJadoul: If we inline the resulting SUB (since it doesn't fit the template anyway), we can do it with only 3 IS_BIT calls by batching into 48 bit limbs before computing carries

spec/dvrm.typ
```diff
@@ -5,13 +5,139 @@
...
+It now remains to enforce that `extended_n_sub_r` is the _signed_ 128-bit representation of $#`n`-#`r`$.
+Here, we introduce `extended_n` and `extended_r`.
+By their definition, these variables contain the signed 128-bit representations of `n` and `r`.
+With this in place, @dvrm:c:n_sub_r ensures `extended_n_sub_r` must contain the correct value.
+
+Lastly, observe that $#`n` - #`r` in (-2^64, 2^64)$, _regardless_ of the value of `signed`.
+Moreover, note that the upper halves of all values in this range are either `0xFFFFFFFF` (negative) or `0x00000000` (non-negative).
+This means that we do not need to store all 128 bits of `extended_n_sub_r`.
+Rather, we need only store the lower 64-bits, and a separate bit (`sign_n_sub_r`) indicating whether the top limbs are all-ones or all-zeroes.
+The prover is free to select the value for `sign_n_sub_r`; only one of the two will fit the proof.
```
RobinJadoul, line 122: Is this really true, without doing the subtraction mod 2^128?

spec/src/dvrm.toml
```diff
@@ -0,0 +1,380 @@
...
+kind = "arith"
+iter = ["i", 0, 3]
+constraint = "$#`div_by_zero` => #`q[i]` = 65535$"
+poly = ["*", "div_by_zero", ["-", ["idx", "q", "i"], 65535]]
+ref = "dvrm:c:q_if_div_by_zero"
+
+[[constraints.div_by_zero]]
+kind = "interaction"
+tag = "ZERO"
+input = ["d"]
```
RobinJadoul: This should sum over the limbs of `d`

RobinJadoul (commented):
I will definitely suggest a manual `rebase -i` and merge here, rather than a full squash, to keep #280 its own commit if it gets merged in here.

erik-3milabs:
> I will definitely suggest a manual `rebase -i` and merge here, rather than a full squash, to keep #280 its own commit if it gets merged in here.

I'm slating #280 to merge into this as a "chained" PR; this PR should be merged first, after which #280 will update to be based against spec/main. The reason I'm pointing #280 against this, is that I can use some of things here in there.
Once this PR is merged, I'll rebase the few left-over commits on #280 against `spec/main`.

spec/dvrm.typ
erik-3milabs: > Any specific reason for this?

Prevent footnote to be confused with an exponent (I have it right after an equality some times).

> If so, probably better to move that up to the book.typ or ebook.typ level, since now it's suddenly different from how it looks in memory.typ

Agreed. #283 does the same thing but in `ebook.typ`. (cda69ff)

spec/dvrm.typ
erik-3milabs: I absolutely agree that the SUB should be over 2^128; it was an oversight on my end not to modify this.

spec/src/dvrm.toml
erik-3milabs: Is dropping a single `IS_BIT` template (= just a degree-2 constraint) worth the extra complexity of casting to very unusual 48-bit limbs? :raised_eyebrow: I don't think so.

spec/chip.typ
erik-3milabs: It was a futile attempt. Reverted it.

spec/src/dvrm.toml
erik-3milabs: Yes.

spec/dvrm.typ
RobinJadoul: I suggest we remove it from here then, and I'll suggest a commit to handle both web and pdf over in #283.

```suggestion
```

RobinJadoul: ea43e4e726bba3f64928b6eca8aeddb29b052c14

RobinJadoul:
Seems like #280 landed before this one to stay as separate history against spec/main :)

spec/src/dvrm.toml
RobinJadoul: Probably not, but since we didn't have to type it anyway, I could as well suggest it :)

spec/src/dvrm.toml
```diff
@@ -0,0 +1,404 @@
...
+type = "QuadHL"
+desc = "sign-extended value of `r`."
+def = {idx="i", polys = [
+  {iter=[0, 3], poly=["idx", "r", "i"]},
+  {iter=[4, 7], poly=["*", 0xFFFF, "sign_r"]}
+]}
+
+[[variables.virtual]]
+name = "extension_n_sub_r"
+type = "DWordHL"
```
RobinJadoul, line 122: Nitpick: this could just be a single `Half`, or even integrated into `extended_n_sub_r` like we do for `extended_n` and `extended_r`

RobinJadoul (approved):

erik-3milabs:
> Seems like #280 landed before this one to stay as separate history against spec/main :)

yeah.. my bad

spec/src/dvrm.toml
erik-3milabs: the reason it exists separately, is that we need it inside the second MUL lookup. And slicing was not going to be an option, really.

erik-3milabs:
This was manually merged into `spec/main`; closing it now

PR #223, opened 2026-01-23
spec: minor `MUL` fixes

erik-3milabs:
- fix missing `iter`s
- fix the `res` slice in the lookup contributions; res is a `QuadWL` not a `QuadHL`

### DIFF
--- a/spec/src/mul.toml
+++ b/spec/src/mul.toml
@@ -116,11 +116,11 @@ pad = 0
 
 [[assumptions]]
 desc = "`IS_HALF[lhs[i]]`"
-range = ["i", 0, 3]
+iter = ["i", 0, 3]
 
 [[assumptions]]
 desc = "`IS_HALF[rhs[i]]`"
-range = ["i", 0, 3]
+iter = ["i", 0, 3]
 
 [[assumptions]]
 desc = "`IS_WORD[lo[i]]`"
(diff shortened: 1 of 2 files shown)


spec/src/mul.toml
```diff
@@ -166,14 +166,14 @@ name = "lookup"
 kind = "interaction"
 tag = "MUL"
 input = ["lhs", "lhs_signed", "rhs", "rhs_signed", "0"]
-output = ["idx", "res", "0:4"]
+output = ["idx", "res", "0:2"]
```
RobinJadoul: For the formal language, idk if the slice syntax is that great
imo the ideal is indexing a casted version; otherwise we may have to find some way to do this cleanly

erik-3milabs: This problem essentially needs "split-casting": split a type into parts, and cast each part to another type.
Since we do not have that, the best next thing I can think of is to split `res` into `lo` and `hi` and only define `res` in text for the discussion. See ad342e0.

spec/src/mul.toml
```diff
@@ -109,16 +116,21 @@ pad = 0
...
-range = ["i", 0, 3]
+iter = ["i", 0, 3]
 
 [[assumptions]]
 desc = "`IS_HALF[rhs[i]]`"
-range = ["i", 0, 3]
+iter = ["i", 0, 3]
+
+[[assumptions]]
+desc = "`IS_WORD[lo[i]]`"
```
RobinJadoul: Are these assumptions sound? Only one of the two may get returned.

RobinJadoul (commented):
I see three options for the casting:
- the current "text only" approach
- similarly, but `res` is a split-defined virtual
- cast `res` to `DWordWL[2]` and index

spec/src/mul.toml
erik-3milabs: No. Good catch :ok_hand:

RobinJadoul (approved):

PR #224, opened 2026-01-23
tooling(executor): add a script to compare benchmarks between two different branches

fmoletta:
Adds a script that compares all benchmark programs between two set branches (similar to hyperfine workflow). This should be useful for local development


scripts/bench_compare.sh
```diff
@@ -0,0 +1,107 @@
...
+# Validate
+if [ "$BASE_BRANCH" = "$COMPARE_BRANCH" ]; then
+    echo -e "${RED}Error: Cannot compare a branch against itself.${NC}"
+    exit 1
+fi
+
+if ! git -C "$ROOT_DIR" diff --quiet; then
+    echo -e "${RED}Error: You have uncommitted changes. Please commit or stash them first.${NC}"
+    exit 1
+fi
```
gianbelinche: I think its not useful to compare only with already commited branch, usually i want to know if something gives better performance before commiting it

fmoletta: Why not? You can just revert said commit if the changes don't show the expected improvement.
In order to allow this behaviour we would have to stash and pop changes back and forth when building the cli for each branch which is unnecessary complex and could lead to unexpected results

gianbelinche (approved):

PR #225, opened 2026-01-23
spec: Allow for cross referencing between different chapters, both in pdf and web mode

RobinJadoul:
There are still some weird things with trying to add numbering to headings that shiroa fights with.

We probably also want a separate chapter reference somehow, since that may look even more different between pdf and web?


spec/decode.typ
```diff
@@ -36,7 +36,7 @@ Empty rows with the following content can be added to achieve this:
 #render_chip_padding_table(chip, config)
 
 Note that this row sets the `EBREAK` flag.
-Given that `CPU` asserts that `EBREAK = 0` (see @cpu:c:ebreak_traps), using this "padding-instruction" would immediately make the CPU table unprovable.
+Given that `CPU` asserts that `EBREAK = 0` (see #xref("/cpu.typ", <cpu:c:ebreak_traps>)), using this "padding-instruction" would immediately make the CPU table unprovable.
```
erik-3milabs: Given that we have established the convention of having refs with format `<chip>:<type>:<label>`, could we truncate this to just

```suggestion
Given that `CPU` asserts that `EBREAK = 0` (see #xref(<cpu:c:ebreak_traps>)), using this "padding-instruction" would immediately make the CPU table unprovable.
```
and resolve the page from the ref itself?

In addition to providing shorter notation (which is always nice), the current solution permits the following issue:
- In `a.typ`: `xref("/b.typ", <c:c:ex>)`
- in `b.typ`: `xref("/c.typ", <c:c:ex2>)`
- and in `c.typ` define `<c:c:ex>` and `<c:c:ex2>`

that is to say: if you pass `xref` the wrong file, but that file happens to includes a hidden copy of the file you're actually interested in, the reference resolves to the wrong page. (I managed to make this work, without compile errors.)

spec/book.typ
```diff
@@ -47,3 +47,46 @@
            align(center, strong(text(fill: black, title))))
     #align(left, body)
 ])
+
+#let is-shiroa = is-web-target()
```
erik-3milabs: Browsing the typst documentation, I indeed noticed that typst uses kebab-case. Up to this point, however, we have primarily used snake_case.

Are you implicitly suggesting you want to switch to using kebab-case moving forward? If so, we should decide on this. Otherwise, I'd prefer we be consistent and refactor all of this to snake_case

spec/book.typ
```diff
@@ -47,3 +47,46 @@
...
+    content
+  }
+}
+
+#let xref(file, lbl, ..ref-args) = {
+  if is-shiroa {
+    // Because shiroa does weird url escaping
+    let shiroa-label = label(str(lbl).replace(":", "%3A"))
+    context if file not in _xref-included.get() {
+        // Let's blow up the compile times :)
```
erik-3milabs:
```suggestion
        // Warning: including file increases compile time
```

spec/book.typ
```diff
@@ -47,3 +47,46 @@
...
+    strip-all(content.child)
+  } else {
+    content
+  }
+}
+
+#let xref(file, lbl, ..ref-args) = {
+  if is-shiroa {
+    // Because shiroa does weird url escaping
+    let shiroa-label = label(str(lbl).replace(":", "%3A"))
```
erik-3milabs: Does shiroa only do this to ":", or should other characters also be escaped (such as underscores, and "-", to name a few)

spec/book.typ
```diff
@@ -47,3 +47,46 @@
...
+  if repr(content.func()) == "sequence" {
+    for c in content.children {
+      strip-all(c)
+    }
+  } else if repr(content.func()) == "styled" {
+    strip-all(content.child)
+  } else {
+    content
+  }
+}
```
erik-3milabs, line 71: Could you elaborate on what this function is supposed to do?

spec/book.typ
```diff
@@ -47,3 +47,46 @@
...
+
+#let xref(file, lbl, ..ref-args) = {
+  if is-shiroa {
+    // Because shiroa does weird url escaping
+    let shiroa-label = label(str(lbl).replace(":", "%3A"))
+    context if file not in _xref-included.get() {
+        // Let's blow up the compile times :)
+        hide(box(width: 0%, height: 0%, strip-all(include file)))
+        _xref-included.update(it => it + ((file): true))
```
erik-3milabs:
```suggestion
      hide(box(width: 0%, height: 0%, strip-all(include file)))
      _xref-included.update(it => it + ((file): true))
```

erik-3milabs (commented):
Let the remarks not fool you: hats off for making this work 👌

spec/book.typ
RobinJadoul: AFAICT, dashes and underscores are not affected, and we've not really used any other weird ones

spec/book.typ
RobinJadoul: It might make sense to keep the typst code a bit idiomatic typst, but I have no major issue with the snake case either

spec/decode.typ
RobinJadoul: That could definitely be possible, but I'm also reworking things a lot to make the PDF nicer to work with and allow chapter references, which currently consist of just the file name; but we can look at reworking that after I finish drafting that part
one potential question arises when we consider potentially having typst source files with references located in subdirectories

spec/book.typ
RobinJadoul: It does deserve some documentation, doesn't it :)
The short version is stripping all styling, which importantly includes a `set document` in the `project` (or `book-page`) function.

RobinJadoul:
It _may_ be possible to do some magic with show rules on ref to automatically use the xref logic, but it's already cost me too much time and sanity to keep going with that attempt.
Here be dragons :)

RobinJadoul:
Idk what I did different in the end, but db8b664c5ef5f1438ebf7fbe4ffda32fd256fb52 now works on my machine at least.
The downside is that memory usage went up by about 1GB, and compilation time is a tad slower (maybe 1.5x), compared to the `#xref(<label>)` version from before.
Memory usage seems to be in a 1.5x-2x increase over spec/main.
I think that if it works, we can use it like this, and it's nice. We may have to re-evaluate if it's all too heavy on CI, so maybe some history rewriting for nicely separated commits and a manual merge instead of a squash will be good for that.

spec/templates/page.typ
```diff
@@ -86,7 +86,7 @@
 /// - authors (array | str): The author(s) of the page.
 /// - kind (str): The kind of the page.
 /// - plain-body (content): The plain body of the page.
-#let project(title: "Typst Book", description: auto, authors: (), kind: "page", plain-body) = {
+#let project(title: "Typst Book", description: auto, authors: (), kind: "page", cond: none, plain-body) = {
```
erik-3milabs, line 92: `cond` is undocumented. Perhaps it even deserves a better name

spec/book.typ
```diff
@@ -1,31 +1,36 @@
...
+#book-meta(
+  title: meta.title,
+  authors: meta.authors,
+  summary: meta.summary.map(((ch, title, _ref)) => chapter(ch, title)).join()
 )
 
-// re-export page template
-#import "/templates/page.typ": project
-#let book-page = project
+#let is-shiroa = "x-target" in sys.inputs
```
erik-3milabs: Move this down, closer to the functions actually needing it?

spec/book.typ
```diff
@@ -47,3 +52,84 @@
...
+    strip-all(content.child)
+  } else {
+    content
+  }
+}
+
+#let _toplevel = state("_toplevel", none)
+#let _xref-included = state("_xref-included", (:))
+
+#let xref-include(f) = {
```
erik-3milabs, line 77: Missing docstring

spec/book.typ
```diff
@@ -47,3 +52,84 @@
...
+  context if f not in _xref-included.get() {
+    hide(box(width: 0%, height: 0%, strip-all(include "/" + f)))
+  }
+  context _xref-included.update(x => x + ((f): true))
+}
+
+#let xref(rf, ..ref-args) = {
```
erik-3milabs: missing docstring / missing argument doc

spec/book.typ
```diff
@@ -47,3 +52,84 @@
...
+  assert(is-shiroa, message: "xref should only be used when compiling for shiroa")
+  let lbl = rf.target
+  let found =  meta.summary.find(x => str(lbl).starts-with(str(x.at(2))))
```
erik-3milabs:
```suggestion
  let found = meta.summary.find(((_, _, tag)) => str(lbl).starts-with(str(tag)))
```

spec/ebook.typ
```diff
@@ -1,12 +1,18 @@
...
 
-#show: ebook.project.with(title: "typst-book", spec: "book.typ")
 #style.update((
   foreground: black,
 ))
 
-// set a resolver for inclusion
-#ebook.resolve-inclusion(it => include it)
+#align(center, title(meta.title))
+// #outline()
```
erik-3milabs:
```suggestion
#pagebreak()
#outline(title: none)
```

spec/ebook.typ
```diff
@@ -1,12 +1,18 @@
 #import "@preview/shiroa:0.3.1": *
```
erik-3milabs: This can be removed, right?

spec/ebook.typ
```diff
@@ -1,12 +1,18 @@
...
 
-// set a resolver for inclusion
-#ebook.resolve-inclusion(it => include it)
+#align(center, title(meta.title))
+// #outline()
+
+#meta.summary.map(((ch, title, ref)) => [
+  #heading(supplement: [Chapter], level: 1, title)#ref
+  #include ch
+]).join()
```
erik-3milabs, line 19: Let's give each chapter its own page:
```suggestion
#meta.summary.map(((ch, title, ref)) => [
  #pagebreak()
  #heading(supplement: [Chapter], level: 1, title)#ref
  #include ch
]).join()
```

spec/memory.typ
```diff
@@ -12,9 +12,7 @@
 #let config = load_config()
 #let chip = load_chip("src/page.toml", config)
 
-#show: book-page.with(title: "Memory argument")
-
-= Memory argument
+#show: book-page("memory.typ") // TODO: can we get rid of this?
```
erik-3milabs: Perhaps we can do something along the lines of
```suggestion
#show: book-page(chip.name)
```
instead?

spec/book.typ
```diff
@@ -47,3 +52,84 @@
            align(center, strong(text(fill: black, title))))
     #align(left, body)
 ])
+
+// Strip styling to keep only "pure" content.
+// This is useful to avoid errors on the `set document(...)` in `project`
+// when invisibly including other chapters to resolve xrefs.
+#let strip-all(content) = {
```
erik-3milabs, line 61: perhaps we should move the following xref code to a separate file?
Given that it all belongs together, it feels like it should be put in a separate module.
Doing so would also allow some top-of-document description on what the module is trying to achieve.

Then again, the code does depend on the format of `meta`, so perhaps it is too coupled to be put is a separate file. I'm curious about your thoughts on the matter.

erik-3milabs (commented):
Some minor comments. Again: hats off for making this happen!

spec/book.typ
RobinJadoul: xref itself *could* move to a separate file (along with dependencies), but it does use meta, and we need it in `book-page`, so it's quite entangled.
Additionally, it's also only useful for the shiroa build, so keeping that localized in book.typ seems reasonable

spec/memory.typ
RobinJadoul: This doesn't work in all cases, and I'm a bit worried about having (right now) and enforcing (for the future)  specific correspondences between file names and chip names
Maybe as a "default" it still makes sense, and we just check if it ends with '.typ' to opt-out in the driving code
I suppose it's really only the xref-include that should care about file names

erik-3milabs (approved):
lgtm

PR #229, opened 2026-01-23
feat(executor): return instruction cache instead of instruction hash map

fmoletta:
Based on #211 
* Return the new InstructionCache instead of the instruction hashmap as part of the execution result
* Add new log and instructions builder for prover unit testing


gianbelinche (approved):

diegokingston (approved):

MauroToscano:
Closing as stale — open since March with failing CI and overtaken by subsequent executor changes. If the instruction-cache approach is still worth pursuing, let's reopen a fresh PR against current main with green CI.

PR #233, opened 2026-01-23
Goldilocks operations with AVX

gabrielbosio:
This draft PR for now implements a new Goldilocks struct that uses AVX instructions for its operations. I still have to check if I can have a final implementation that uses the already implemented struct.


gabrielbosio:
Actually this is a subset of #162. I'm closing this unless we want to separate the work in smaller PRs.

PR #235, opened 2026-01-26
spec: `HALT` chip

erik-3milabs:


erik-3milabs:
@RobinJadoul some open questions:
- what should the exact timestamp for zeroizing the registers be?
- deal with the CPU still attempting to write the return value to `rd`.
    - perhaps just set `write_register = 0` for `ECALL`?

RobinJadoul:
> what should the exact timestamp for zeroizing the registers be?

I'd say it should be 2^64 - 1, the max possible

> perhaps just set write_register = 0 for ECALL?

I suppose that is an option while we only have a limited amount of syscalls, until the kernel idea can take over
longer term, there will be syscalls that have return values, so keeping to the abi would be nice
When write_register ≠ 0, we just need it to integrate with the executor so that the return value is known (and the right order of updates makes it to MEMW for the consistency of then clearing the register)

spec/src/halt.toml
```diff
@@ -0,0 +1,76 @@
...
+
+
+
+[[constraints.all]]
+kind = "interaction"
+tag = "MEMW"
+input = [1, 10, 0, ["-", ["^", 2, 64], 1], 1, 0, 0]
+output = 0
+multiplicity = 1
+ref = "halt:c:read_zero_exit_code"
```
RobinJadoul: I would maybe place this one in between the two ranges, so the joined range is clearly complete

spec/src/halt.toml
```diff
@@ -0,0 +1,76 @@
...
+
+[[variables.input]]
+name = "timestamp"
+type = "DWordWL"
+desc = "timestamp at which to halt the program"
+
+[[variables.input]]
+name = "rv2"
+type = "DWordWL"
+desc = "value of register two at the point of calling HALT."
```
RobinJadoul: This is the first syscall argument, right? Then we should probably get rid of the column and assert it's zero, a successful exit code.

spec/src/halt.toml
RobinJadoul: Or it's probably not necessary to treat it separately, as this would be enforced by the rv2 change.

spec/halt.typ
```diff
@@ -0,0 +1,47 @@
...
+#let nr_columns = total_nr_instantiated_columns(chip, config)
+
+The #halt chip leverages #nr_variables variables, spanning #nr_columns columns:
+#render_chip_column_table(chip, config)
+
+== Assumptions
+It is assumed the input is range checked:
+#render_chip_assumptions(chip, config)
+
+== Constraints
```
RobinJadoul: So we're currently letting the verifier do the register cleanup in the lookup argument.
Might be worth documenting, along the alternative option of adding an extra 31 lookups to the "memory" table to remove the known final tokens for the registers

spec/src/halt.toml
erik-3milabs: I got rid of `rv2`, but also removed it from the interaction. As such, we still need to read a 0 there in HALT-C2

spec/halt.typ
```diff
@@ -0,0 +1,53 @@
...
+  render_constraint_table,
+  total_nr_instantiated_columns,
+  total_nr_variables,
+)
+
+#let config = load_config()
+#let chip = load_chip("src/halt.toml", config)
+#let halt = raw(chip.name)
+
+#show: book-page(chip.name)
```
RobinJadoul: Maybe to integrate in the ECALL placeholder chapter directly

spec/src/halt.toml
```diff
@@ -0,0 +1,56 @@
...
+iter = ["i", 0, 1]
+
+
+[[constraint_groups]]
+name = "all"
+
+[[constraints.all]]
+kind = "interaction"
+tag = "MEMW"
+input = [1, "i", 0, ["-", ["^", 2, 64], 1], 1, 0, 0]
```
RobinJadoul: All register indices as addresses should be scaled up

RobinJadoul (approved):

PR #237, opened 2026-01-26
fix(spec): Remove broken reference in `decode.typ`

cdesaintguilhem:
@RobinJadoul This might be already fixed in another PR, just close this one if this temporary fix isn't needed in the end.


RobinJadoul:
This should be solved by #225

PR #240, opened 2026-01-26
Fix: Simplify LT bus packing to 2 elements per operand

MauroToscano:
## Summary
  - Change LT bus receiver from 3× Direct packing to `Packing::DWordHHW` (2 elements)
  - Change CPU LT sender from 3 linear combinations to `Packing::DWordBL` (2 elements)
  - Both produce the same `[lo32, hi32]` fingerprint, enabling interoperability with DWordWL senders
  - Add constraints so everything is properly range checked
  
    ## Motivation
  This change enables MEMW timestamps (DWordWL format) to send to the LT bus using `Packing::DWordWL`, which also produces `[lo32, hi32]`. Previously, LT expected 3 elements per operand which was incompatible.

  ## Changes
  - `prover/src/tables/lt.rs`: Use `Packing::DWordHHW` for lhs/rhs (reads 3 columns, produces 2 bus elements)
  - `prover/src/tables/cpu.rs`: Use `Packing::DWordBL` for arg1/arg2 (reads 8 bytes, produces 2 bus elements)

### DIFF
--- a/prover/src/tables/cpu.rs
+++ b/prover/src/tables/cpu.rs
@@ -1220,101 +1220,26 @@ pub fn bus_interactions() -> Vec<BusInteraction> {
     // -------------------------------------------------------------------------
     // LT interaction (for SLT, BLT)
     // -------------------------------------------------------------------------
-    // LT[arg1::DWordHHW, arg2::DWordHHW, signed] -> res[0]
+    // LT[arg1, arg2, signed] -> res[0]
     // multiplicity = SLT + BLT
     //
-    // DWordHHW format: [Word(0-31), Half(32-47), Half(48-63)]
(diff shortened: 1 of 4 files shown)


diegokingston (approved):

RobinJadoul:
If I'm understanding this correctly; you're essentially casting a HHW to a WL? Which will probably introduce unsoundness on the LT side, because the range-check assumptions (LT-A1/LT-A2) no longer necessarily hold if they're not enforced by the logup.

RobinJadoul (changes requested):
Formally marking it as "request changes" to prevent accidental merging

RobinJadoul (approved):
I don't know enough about the rest of the code, but this specific unsoundness should now be addressed 👍

PR #243, opened 2026-01-27
spec: LogUp: Vanilla protocol description

cdesaintguilhem:
An initial description of the vanilla (i.e., original) LogUp protocol for permutation checks based on logarithmic derivatives.

To-do after initial review:
- [ ] Check against existing code for consistency.


spec/logup.typ
```diff
@@ -0,0 +1,81 @@
...
+#let weightFunction = $w$
+#let multiplicity = $mu$
+
+The $j$-th _interaction_ $Interaction_j$ of table $Table_i$ is defined by the following tuple:
+
+#table(
+  columns: (auto, auto),
+  inset: 6pt,
+  align: horizon,
+  table.header([Symbol], [Description]),
```
erik-3milabs: IMO, this improves the readability of the table:
```suggestion
  stroke: none,
  table.header([*Symbol*], [*Description*]),
  table.hline(stroke: 1pt),
  table.vline(stroke: 1pt, x: 1),
```

Result:
Image

spec/logup.typ
```diff
@@ -0,0 +1,81 @@
...
+  [$id in FF$], 
+  [the _type identifier_ of the interaction, usually the identifier of the chip that is sending or receiving the interaction.],
```
erik-3milabs: Since almost every interaction has a distinct sending and receiving chip, this sentence leaves me wondering which of the two it will be. But, yeah, I cannot find a good way to better describe the relation myself 🙈

spec/logup.typ
```diff
@@ -0,0 +1,81 @@
...
+
+== Interaction Notation
+
+#let Interaction = $I$
+#let id = math.sans()[id]
+#let numElements = $l$
+#let weightFunction = $w$
+#let multiplicity = $mu$
+
+The $j$-th _interaction_ $Interaction_j$ of table $Table_i$ is defined by the following tuple:
```
erik-3milabs, line 46: The interaction table uses `T` and `m`, while the "VM notation" section talks about `T_i` and `m_i`. I understand that the `i` can be ignored _here_. However, the "Vanilla LogUp Protocol" also uses `m` (no subscript), making the thing tough to read.

spec/logup.typ
```diff
@@ -0,0 +1,81 @@
...
+#let TableSet = ${Table_i}_(i in [t])$
+#let numColumns = math.sans()[m]
+#let numRows = math.sans()[N]
+
+- $numTables in NN$: number of tables $Table_i$ in the arithmetisation of the VM.
+- $TableSet$: set of all tables $Table_i$ in the arithmetisation of the VM.
+- $numColumns_i in NN$: number of _columns_ in table $Table_i$ (not the number of variables).
+- $numRows_i in NN$: number of _rows_ in table $Table_i$.
+
+== Interaction Notation
```
erik-3milabs, line 38: The following are not defined in text:
1. `N`, (probably the set of natural numbers. I prefer verbosity over implicit assumption, however.)
1. `FF`, 
1. `ExtensionField`, 
1. `[x]` (I think this is short for the set `{0, ..., x-1}`? or is it `{1, ..., x}`?)
1. `X[y]`. Is this indexing, i.e., of the `y`th element of `X`?

spec/logup.typ
```diff
@@ -0,0 +1,81 @@
...
+  [$numElements in NN$], 
+  [the _length_ of the tuple of elements being looked-up.],
+  [$weightFunction : FF^(numColumns) arrow FF^(numElements)$],
+  [the _weight function_ that maps row elements of table $Table$ to the looked-up tuple.],
+  [$multiplicity in [numColumns] union {1_BaseField, dots}$],
+  [the _multiplicity_ of the interaction, as either a column index in table $Table$ or a constant value.]
+)
+
+
+= Vanilla LogUp Protocol
```
erik-3milabs: a) it feels to me like this protocol is not complete yet. Or perhaps it is complete, but the last two steps are at the incorrect indent.
b) each nesting using the same numbering style (1., 2., 3., ...) is confusing

spec/logup.typ
```diff
@@ -0,0 +1,81 @@
...
+  + For each table $Table_i$, populate the interaction contribution columns and compute the _table (LogUp) contribution_:
+
+    + For each interaction $Interaction_j$ of table $Table_i$, initialize an empty _interaction contribution column_ of length $numRows_i$.
+
+    + Initialise a _table running sum column_ of length $numRows_i$ with $Sigma_0 = 0 in ExtensionField$ in the first row.
+
+    + For each $j$-th row $R_j in BaseField^(numColumns_i)$ of $Table_i$, for $j in [numRows_i]$:
+      + For each $k$-th interaction $Interaction_k$ of table $Table_i$:
+        + Compute the _interaction contribution numerator_ $ n_(j,k) = cases(R_j [multiplicity] quad & "if" multiplicity in [numColumns]",", multiplicity & "otherwise.") $
+        + If $n eq.not 0$, compute the _interaction contribution denominator_ $ d_(j,k) = logupChallenge + fingerprintCoeff dot Interaction_k\.id + sum_(l = 0)^(numElements - 1) fingerprintCoeff^(l + 1) dot weightFunction(R_l)[j] $.
```
erik-3milabs: I think you swapped indices here:

```suggestion
        + If $n eq.not 0$, compute the _interaction contribution denominator_ $ d_(j,k) = logupChallenge + fingerprintCoeff dot Interaction_k\.id + sum_(l = 0)^(numElements - 1) fingerprintCoeff^(l + 1) dot weightFunction(R_j)[l] $.
```

erik-3milabs: Also, why are we choosing here for the power-of-beta solution? I thought we would achieve better soundness when we instead use independently sampled `beta_i`?

spec/logup.typ
```diff
@@ -0,0 +1,81 @@
...
+        + Save the _interaction contribution_ as $n_(j,k)/d_(j,k) in ExtensionField$ in the corresponding interaction contribution column for this interaction.
+        + *Constrain* the interaction contribution column according to the definitions of $n$ and $d$.
```
erik-3milabs: we might want to give them a bit of a hand hold here. Especially because I would have no clue how to constrain `n`.

spec/logup.typ
erik-3milabs: If we do stick with the power-of-beta solution, we will have to increase the power of beta in the sum to `l+2`; right now, `Ik.id` and `w(Rj)[0]` are both multiplied with `beta^1`.


```suggestion
        + If $n eq.not 0$, compute the _interaction contribution denominator_ $ d_(j,k) = logupChallenge + fingerprintCoeff dot Interaction_k\.id + sum_(l = 0)^(numElements - 1) fingerprintCoeff^(l + 2) dot weightFunction(R_l)[j] $.
```

spec/logup.typ
```diff
@@ -0,0 +1,81 @@
...
+
+  + For each table $Table_i$, populate the interaction contribution columns and compute the _table (LogUp) contribution_:
+
+    + For each interaction $Interaction_j$ of table $Table_i$, initialize an empty _interaction contribution column_ of length $numRows_i$.
+
+    + Initialise a _table running sum column_ of length $numRows_i$ with $Sigma_0 = 0 in ExtensionField$ in the first row.
+
+    + For each $j$-th row $R_j in BaseField^(numColumns_i)$ of $Table_i$, for $j in [numRows_i]$:
+      + For each $k$-th interaction $Interaction_k$ of table $Table_i$:
+        + Compute the _interaction contribution numerator_ $ n_(j,k) = cases(R_j [multiplicity] quad & "if" multiplicity in [numColumns]",", multiplicity & "otherwise.") $
```
erik-3milabs: I've never heard of the idea of using `mu` to index into the row? Either this is a mistake, or I would like to be brought up to speed on this concept.

spec/logup.typ
```diff
@@ -0,0 +1,81 @@
...
+
++ Verifier samples a random _(global) LogUp challenge_ $logupChallenge in ExtensionField$ and a random _fingerprint coefficient_ $fingerprintCoeff in ExtensionField$ and sends them to the Prover.
+
++ Prover commits to interaction contribution and table running sum columns and to each table's contribution:
+
+  + For each table $Table_i$, populate the interaction contribution columns and compute the _table (LogUp) contribution_:
+
+    + For each interaction $Interaction_j$ of table $Table_i$, initialize an empty _interaction contribution column_ of length $numRows_i$.
+
+    + Initialise a _table running sum column_ of length $numRows_i$ with $Sigma_0 = 0 in ExtensionField$ in the first row.
```
erik-3milabs:
```suggestion
    + Initialise a _table running sum column_ $Sigma$ of length $numRows_i$ with $Sigma_0 = 0 in ExtensionField$ in the first row.
```
Without `sigma` present here, it is really unclear what $\Sigma_0 = 0 \in G$ is supposed to mean.

Also, distinguishing `sigma` and `sum` is hard. I ask we find a replacement for the $\Sigma$ symbol.

spec/logup.typ
```diff
@@ -0,0 +1,81 @@
...
+        + Save the _interaction contribution_ as $n_(j,k)/d_(j,k) in ExtensionField$ in the corresponding interaction contribution column for this interaction.
+        + *Constrain* the interaction contribution column according to the definitions of $n$ and $d$.
+
+      + Compute the _row contribution_ as the sum $sigma_j = sum_k n_(j,k) / d_(j,k)$ and compute the next row's table running sum value $Sigma_(j+1) = Sigma_j + sigma_j$.
+
+      + *Constrain* the update of the next row's running sum value.
+
+  + Batch-commit to every table's interaction contribution columns and running sum columns with the column commitment scheme and commit to the table's overall contribution $Sigma_(N_i)$ by sending it in the clear to the verifier.
+
+  + Verifier checks that the sum of every table's overall contribution is equal to zero: $sum_i Sigma_i == 0 in ExtensionField$, and delegates the checks of the constraints to the STARK.
```
erik-3milabs: I think a single `=` suffices here:


```suggestion
  + Verifier checks that the sum of every table's overall contribution is equal to zero: $sum_i Sigma_i = 0 in ExtensionField$, and delegates the checks of the constraints to the STARK.
```

erik-3milabs (commented):
Some remarks.

spec/logup.typ
cdesaintguilhem: I tried a different phrasing, let me know 😅

spec/logup.typ
cdesaintguilhem: a) It is meant to be complete, aside from the surrounding steps for the STARK, but the _last_ step was indeed incorrectly indented.
b) I've tried to change that to a less confusing style.

spec/logup.typ
cdesaintguilhem: I've replaced the `sigma`s with `S`s, let me know if that doesn't clarify it enough.

spec/logup.typ
cdesaintguilhem: So here I took a bit of a liberty that I now realise clashes with the notation in the table pages. From what I can tell from the interactions we specified so far, they either have 1) a dedicated column which contains the multiplicity that should be used, or 2) a constant multiplicity that is not stored in a column.

In case 1, the column which contains the multiplicity has an index, and that's the one to which I refer here. In case 2, the multiplicity is just a constant field element.

So in short, my `mu` is the index of the `mu`s in the table pages it seems. I'm not yet sure what is the right way to solve that.

spec/logup.typ
cdesaintguilhem: Regarding the power-of-beta, that's a good point. I'll leave it as that for the time being, and when I compare against their current code I'll adjust (if needed) to what they're doing.

For now I've fixed the off-by-one error.

cdesaintguilhem (commented):
@erik-3milabs I've addressed all but one of your comments with my commit.

@RobinJadoul It's now worth you having a look over it too, if you have some time.

spec/logup.typ
```diff
@@ -0,0 +1,100 @@
...
+#let logupChallenge = math.alpha
+#let fingerprintCoeff = math.beta
+
++ Prover commits to all traces.
+
++ Verifier samples a random _(global) LogUp challenge_ $logupChallenge in ExtensionField$ and a random _fingerprint coefficient_ $fingerprintCoeff in ExtensionField$ and sends them to the Prover.
+
++ Prover commits to interaction contribution and table running sum columns and to each table's contribution:
+
+  #set enum(numbering: "a.")
```
erik-3milabs: Rather than setting the enum at each level, you can also do
```suggestion
#set enum(numbering: "1.a.i.1.a.")
```
at the top level and specify it for all sub-numberings!

spec/logup.typ
```diff
@@ -0,0 +1,100 @@
+#import "/book.typ": book-page, cdsg
+
+#show: book-page.with(title: "LogUp Argument")
```
erik-3milabs: With the update to `spec/main`, this should be updated to be
```suggestion
#show: book-page("logup")
```

The title is now taken from the second entry in the logup-tuple in `book.typ`'s `meta` variable.

spec/logup.typ
```diff
@@ -0,0 +1,100 @@
...
+
+The _LogUp_ proof system conducts a permutation check based on summing partial derivatives. This check ensures that whatever tuple is sent to be "looked-up" by a _source table_ is indeed received in the expected _destination table_.
+
+= Notation
```
erik-3milabs, line 11: you should probably indent all subheaders; Notation now shows up as a new section, rather than a subsection

spec/logup.typ
```diff
@@ -0,0 +1,100 @@
...
+= Vanilla LogUp Protocol
+
+#let logupChallenge = math.alpha
+#let fingerprintCoeff = math.beta
+
++ Prover commits to all traces.
+
++ Verifier samples a random _(global) LogUp challenge_ $logupChallenge in ExtensionField$ and a random _fingerprint coefficient_ $fingerprintCoeff in ExtensionField$ and sends them to the Prover.
+
++ Prover commits to interaction contribution and table running sum columns and to each table's contribution:
```
erik-3milabs: Writing this as an inlined enumeration should improve the readability
```suggestion
+ Prover commits to (i) interaction contribution, (ii) table running sum columns, and (iii) each table's contribution:
```

spec/logup.typ
```diff
@@ -0,0 +1,100 @@
...
+    + For each interaction $Interaction_j$ of table $Table_i$, initialize an empty _interaction contribution column_ of length $numRows_i$.
+
+    + Initialise a _table running sum column_ $S_i in ExtensionField^(numRows_i)$ with $S_i [0] = 0_ExtensionField$ in the first row.
+
+    + For each $j$-th row $R_j in BaseField^(numColumns_i)$ of $Table_i$, for $j in [numRows_i]$:
+      #set enum(numbering: "1.")
+      + For each $k$-th interaction $Interaction_k$ of table $Table_i$:
+        #set enum(numbering: "a.")
+        + Compute the _interaction contribution numerator_ $ n_(j,k) = cases(R_j [multiplicity] quad & "if" multiplicity in [numColumns]",", multiplicity & "otherwise.") $
+        + If $n eq.not 0$, compute the _interaction contribution denominator_ $ d_(j,k) = logupChallenge + fingerprintCoeff dot Interaction_k\.id + sum_(l = 0)^(numElements - 1) fingerprintCoeff^(l + 2) dot weightFunction(R_j)[l] $.
```
erik-3milabs: 1. You updated the definitions of `id` and `w` to be indexed. This update is not reflected here.
2. Consider updating this line such that the period is part of the equation (and doesn't dangle on its own newline)
```suggestion
        + If $n eq.not 0$, compute the _interaction contribution denominator_ $ d_(j,k) = logupChallenge + fingerprintCoeff dot Interaction_k\.id + sum_(l = 0)^(numElements - 1) fingerprintCoeff^(l + 2) dot weightFunction(R_j)[l]. $
```

spec/logup.typ
```diff
@@ -0,0 +1,100 @@
...
+        + Save the _interaction contribution_ as $n_(j,k)/d_(j,k) in ExtensionField$ in the corresponding interaction contribution column for this interaction.
+        + *Constrain* the interaction contribution column according to the definitions of $n$ and $d$.
+
+      + Compute the _row contribution_ as the sum $s_(j) = sum_k n_(j,k) / d_(j,k)$ and compute the next row's table running sum value $S_i [j+1] = S_i [j] + s_(j)$.
+
+      + *Constrain* the update of the next row's running sum value.
+
+  + Batch-commit to every table's interaction contribution columns and running sum columns with the column commitment scheme and commit to the table's overall contribution $S_i [N_i]$ by sending it in the clear to the verifier.
+
++ Verifier checks that the sum of every table's overall contribution is equal to zero: $sum_i S_i [N_i] = 0_ExtensionField$, and delegates the checks of the constraints to the STARK.
```
erik-3milabs: Note that with this construction, `S_i` will have length `2^k + 1` for some integer `k`: 
- 3.a..ii sets `Si[0] = 0 in G`
- 4 requires `S_i[N_i] = 0 in G`

spec/logup.typ
erik-3milabs: Thanks for clarifying; I now understand what you're trying to convey.

I can't say that I have thought of better notation yet...

erik-3milabs (commented):
some more notes

spec/logup.typ
```diff
@@ -0,0 +1,100 @@
...
+- $numTables in NN$: number of tables $Table_i$ in the arithmetisation of the VM.
+- $TableSet$: set of all tables $Table_i$ in the arithmetisation of the VM.
+- $numColumns_i in NN$: number of _columns_ in table $Table_i$ (not the number of variables).
+- $numRows_i in NN$: number of _rows_ in table $Table_i$.
+
+== Interaction Notation
+
+#let Interaction = $I$
+#let id = math.sans()[id]
+#let numElements = $l$
```
RobinJadoul:
```suggestion
#let numElements = $ell$
```

spec/logup.typ
RobinJadoul: Maybe we can unify both $\mu$ and $w$ by making them affine linear combinations of the columns of a table?
So potentially both as a map of weights.

spec/logup.typ
```diff
@@ -0,0 +1,100 @@
...
+  table.header([*Symbol*], [*Description*]),
+  table.hline(stroke: 1pt),
+  table.vline(stroke: 1pt, x: 1),
+  table.header([Symbol], [Description]),
+  [$id_(i,j) in FF$], 
+  [the _type identifier_ of the interaction, usually the identifier of the chip that is constraining the relation expected to hold within the looked-up tuple.],
+  [$numElements_(i,j) in NN$], 
+  [the _length_ of the tuple of elements being looked-up.],
+  [$weightFunction_(i,j) : FF^(numColumns_i) arrow FF^(numElements_j)$],
+  [the _weight function_ that maps row elements of table $Table_i$ to the looked-up tuple.],
```
RobinJadoul: This doesn't cover constants yet, or affine linear combinations

spec/logup.typ
```diff
@@ -0,0 +1,100 @@
...
+      + For each $k$-th interaction $Interaction_k$ of table $Table_i$:
+        #set enum(numbering: "a.")
+        + Compute the _interaction contribution numerator_ $ n_(j,k) = cases(R_j [multiplicity] quad & "if" multiplicity in [numColumns]",", multiplicity & "otherwise.") $
+        + If $n eq.not 0$, compute the _interaction contribution denominator_ $ d_(j,k) = logupChallenge + fingerprintCoeff dot Interaction_k\.id + sum_(l = 0)^(numElements - 1) fingerprintCoeff^(l + 2) dot weightFunction(R_j)[l] $.
+        + Save the _interaction contribution_ as $n_(j,k)/d_(j,k) in ExtensionField$ in the corresponding interaction contribution column for this interaction.
+        + *Constrain* the interaction contribution column according to the definitions of $n$ and $d$.
+
+      + Compute the _row contribution_ as the sum $s_(j) = sum_k n_(j,k) / d_(j,k)$ and compute the next row's table running sum value $S_i [j+1] = S_i [j] + s_(j)$.
+
+      + *Constrain* the update of the next row's running sum value.
```
RobinJadoul: We should probably cover proper transition and boundary contraints for this

spec/logup.typ
cdesaintguilhem: Shouldn't the top-level heading within a Chapter be a Section? Or has the style
so far been to make top-level headings within Chapters actually be level 2 headings?

spec/logup.typ
cdesaintguilhem: Why not? Can't the weight function specify a constant value as one of the
elements of the output tuple?

spec/logup.typ
cdesaintguilhem: Well 4 actually requires that the _sum_ of the `S_i[N_i]` is 0; but the problem
still stands that `N_i` is actually a out-of-bound index, and there is also an
edge case that needs to be solved for tables with exactly `N_i` rows before
padding.

spec/logup.typ
erik-3milabs: You make a good point. Your current "levelling" is perfect for shiroa; it only breaks down for the PDF.
Let me see whether I can whip up a separate PR to automatically fix this problem for the PDF.

Edit: see #264

spec/logup.typ
RobinJadoul: I think I read it the wrong way around and thought of it as assigning weights for a linear combination to each column in the row.

spec/logup.typ
cdesaintguilhem: I'll delay this until we've written a specification of the proof system.

spec/logup.typ
cdesaintguilhem: It might be that the solution for now is that tables can only have $N_i - 1$ rows before they are padded, so that they're always padded with one row and we don't constrain the interaction contribution columns on the last row.

RobinJadoul: I don't think this works for the CPU, as that still issues IS_BYTE interactions, even in padded rows

cdesaintguilhem: I've expanded constraint choices and corrected instructions in protocol description to allow for that.

spec/logup.typ
```diff
@@ -0,0 +1,135 @@
...
+#let BaseField = math.FF
+#let ExtensionField = math.GG
+
+== VM Notation
+
+=== Preliminary notation
+- $NN$: the set of non-zero natural integers.
+- $BaseField$: the base finite field used by the arithmetisation.
+- $ExtensionField$: a finite extension of $BaseField$ of cryptographic size.
+- $[n]$ for $n in NN$: the set of integers ${0, dots, n - 1}$.
```
RobinJadoul, line 22: Either $\mathbb{N}$ should include zero, or this should index from 1 to n
I'd suggest the former, since there's more zero-based indexing

spec/book.typ
```diff
@@ -25,6 +25,7 @@
     ("load.typ", [LOAD chip], <load>),
     ("ecall.typ", [ECALL chips], <ecall>),
     ("bitwise.typ", [BITWISE chips], <bitwise>),
+    ("logup.typ", [LogUp argument], <logup>)
```
RobinJadoul: I'd move this to the top before the memory argument, then in time we can split it into parts
- intro/background
- theoretical foundation
- arithmetization
- whatever else we may find relevant, maybe recursion arithmetization

spec/logup.typ
```diff
@@ -0,0 +1,135 @@
...
+
+#table(
+  columns: (auto, auto),
+  inset: 6pt,
+  align: horizon,
+  stroke: none,
+  table.header([*Symbol*], [*Description*]),
+  table.hline(stroke: 1pt),
+  table.vline(stroke: 1pt, x: 1),
+  table.header([Symbol], [Description]),
```
RobinJadoul: Double header

spec/logup.typ
```diff
@@ -0,0 +1,135 @@
...
+
++ Verifier samples a random _(global) LogUp challenge_ $logupChallenge in ExtensionField$ and a random _fingerprint coefficient_ $fingerprintCoeff in ExtensionField$ and sends them to the Prover.
+
++ Prover commits to (i) interaction contribution, (ii) table running sum columns, and (iii) each table's contribution:
+
+  + For each table $Table_i$, populate the interaction contribution columns and compute the _table (LogUp) contribution_:
+
+    + For each interaction $Interaction_j$ of table $Table_i$, initialize an empty _interaction contribution column_ of length $numRows_i$.
+
+    + Initialise a _table running sum column_ $S_i in ExtensionField^(numRows_i)$ with $S_i [0] = 0_ExtensionField$ in the first row.
```
RobinJadoul: The indicated values of $S_i$ may be differerent based on the selected constraint choice (e.g. $S_i[0]$ may not be zero)

spec/logup.typ
```diff
@@ -0,0 +1,135 @@
...
+
++ (*Boundary, first row*) Constrain first row of running sum column to equal $0_ExtensionField$.
++ (*Transition, looking forward, applied to rows $0, dots, numRows_i - 2$*) For each row _other than the last_, constrain the _next_ running sum value to equal the sum of every current interaction contribution column added to the _current_ running sum value.
++ (*Boundary, last row*) Constrain last row of running sum column added to sum of last row of every interaction column to equal the claimed table contribution. (That is, the claimed table contribution is implicit in the last row of the table, but not written to last value of running sum column.)
+
+Total constraints: 2 boundary + 1 transition over $numRows_i - 1$ rows.
+
+=== Choice 3: circular transitions looking back/forward
+
++ For each row, constrain the _current/next_ (wrapping to first on last if next) running sum value to equal the sum of every current interaction contribution value added to the _previous/current_ (wrapping to last on first if previous) running sum value added to claimed table contribution divided by $numRows_i$.
```
RobinJadoul: A brief argument of why this works would probably be good to have (especially this one, the other ones maybe too, but they're more obvious)

RobinJadoul (commented):
A couple small things

PR #244, opened 2026-01-27
spec: update `ECALL` signature

erik-3milabs:
Note: in addition to updating the values of `rs2` and `rd`, this also implies an update of the `read_register2` and `write_register` flags for this operation.

### DIFF
--- a/spec/src/cpu.toml
+++ b/spec/src/cpu.toml
@@ -705,12 +705,10 @@ desc = "We treat `EBREAK` as an unprovable trap"
 poly = ["not", "EBREAK"]
 ref = "cpu:c:ebreak_traps"
 
-# TODO: no types available, so no casting yet
 [[constraints.sys]]
 kind = "interaction"
 tag = "ECALL"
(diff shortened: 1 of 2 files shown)


spec/src/cpu.toml
```diff
@@ -705,12 +705,10 @@ desc = "We treat `EBREAK` as an unprovable trap"
 poly = ["not", "EBREAK"]
 ref = "cpu:c:ebreak_traps"
 
-# TODO: no types available, so no casting yet
 [[constraints.sys]]
 kind = "interaction"
 tag = "ECALL"
-input = ["rv1", "pc", "timestamp", "rv2"]
-output = "rvd"
+input = ["timestamp", "rv1"]
```
RobinJadoul: Probably cast rv1 to DWordWL instead of keeping the DWordWHH

erik-3milabs: Good catch

RobinJadoul (approved):

PR #246, opened 2026-01-27
spec: Update LT interaction signature so that it can be used properly for timestamps

RobinJadoul:
See #240

### DIFF
--- a/spec/src/lt.toml
+++ b/spec/src/lt.toml
@@ -79,7 +79,6 @@ pad = 0
 
 [[assumptions]]
 desc = "`IS_WORD[lhs[0]]`"
-iter = ["i", 1, 2]
 ref = "lt:a:range_lhs"
 
 [[assumptions]]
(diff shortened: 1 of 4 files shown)


spec/src/lt.toml
```diff
@@ -78,12 +78,12 @@ pad = 0
 
 
 [[assumptions]]
-desc = "`IS_HALFWORD[lhs[i]]` and `IS_WORD[lhs[0]]`"
+desc = "`IS_WORD[lhs[0]]`"
 iter = ["i", 1, 2]
```
erik-3milabs:
```suggestion
```

spec/src/lt.toml
```diff
@@ -78,12 +78,12 @@ pad = 0
...
 ref = "lt:a:range_lhs"
 
 [[assumptions]]
-desc = "`IS_HALFWORD[rhs[i]]` and `IS_WORD[rhs[0]]`"
+desc = "`IS_WORD[rhs[0]]`"
 iter = ["i", 1, 2]
```
erik-3milabs:
```suggestion
```

erik-3milabs (approved):
Two minor comments; aside from that, lgtm

PR #253, opened 2026-01-28
Move stark examples to test-only test_utils module

diegokingston:
- Move crypto/stark/src/examples/ to crypto/stark/src/test_utils/examples/
- Add test-utils feature to expose examples for benchmarks
- Update all imports to use test_utils::examples
- Fix profile_prover.rs to use GoldilocksField (not U64GoldilocksPrimeField)

This reduces the public API surface by ~3,347 LOC of example code that was only used in tests and benchmarks.


JuArce:
@greptile

PR #264, opened 2026-01-30
spec: fix header levels

erik-3milabs:
Lower the header levels on the HTML export.

Note: also fixes a header misalignment on the DECODE chip page.

### DIFF
--- a/spec/book.typ
+++ b/spec/book.typ
@@ -147,10 +147,7 @@
         #body
       ])
     }
-  } else { // PDF
-    (body) => {
-        set heading(offset: 1)
-        body
-    }
+  } else {
+    (body) => body
   }
 }
--- a/spec/book.typ
+++ b/spec/book.typ
@@ -147,7 +147,10 @@
         #body
       ])
     }
-  } else {
-    (body) => body
+  } else { // PDF
+    (body) => {
+        set heading(offset: 1)
+        body
+    }
   }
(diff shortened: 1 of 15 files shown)


erik-3milabs:
@RobinJadoul could you check that this works w/ shiroa? My PC crapped its pants again when I asked it to export :grumble:

erik-3milabs:
I'm really curious what greptile will make of changes to the .toml files when we get to that :thinking:

spec/book.typ
```diff
@@ -147,7 +147,10 @@
         #body
       ])
     }
-  } else {
-    (body) => body
+  } else { // PDF
+    (body) => {
+        set heading(offset: 1)
```
RobinJadoul: Is this perhaps more appropriate in ebook.typ? That would keep minimal PDF functionality in book.typ and make it clearer where to look for something that impacts only web or only pdf

RobinJadoul (approved):
Shiroa seems to build fine.
One remark, but I'm not sure which is better or not, so feel free to do with it what you like best.

spec/book.typ
erik-3milabs: Good call. Could you please double check that `shiroa` still builds as expected with the change included?

RobinJadoul: confirmed :+1:

PR #270, opened 2026-02-02
spec: `NEG` template

erik-3milabs:
This PR introduces the `NEG` template.

The upside to using this template rather than having a conditional `ADD`/`SUB`, is that this template does _not_ require the output value to be range checked.
It thus allows for the construction of absolute values, where the result fits in the non-rangecheckable `DWordWL`.

### DIFF
--- a/spec/expr.typ
+++ b/spec/expr.typ
@@ -153,7 +153,7 @@
       let (val, idxs) = flat_idxs(e)
       $#rec(PREC.idx, val)_(#idxs.map(idx => rec(PREC.idx, idx)).join($, $))$
     },
-    "not": (pp, rec, e) => mwrap($1 - #rec(PREC.not, e.at(1))$, pp < PREC.not),
+    "not": (pp, rec, e) => mwrap(rec(PREC.not, 1) + $ - #rec(PREC.not, e.at(1))$, pp < PREC.not),
     "+": (pp, rec, e) => mwrap($#e.slice(1).map(rec.with(PREC.add)).join($+$)$, pp < PREC.add),
     "sum": (pp, rec, e) => {
       assert(e.len() == 4, message: "invalid sum:" + repr(e))
--- a/spec/expr.typ
+++ b/spec/expr.typ
@@ -92,7 +92,7 @@
 #let expr_to_code = make_expr_formatter(
   (
     "idx": (pp, rec, e) => rec(PREC.MIN, e.at(1)) + `[` + rec(PREC.MAX, e.at(2)) + `]`,
-    "not": (pp, rec, e) => cwrap(`1 - ` + rec(PREC.not, e.at(1)), pp < PREC.not),
+    "not": (pp, rec, e) => cwrap(rec(PREC.not, 1) + ` - ` + rec(PREC.not, e.at(1)), pp < PREC.not),
     "+": (pp, rec, e) => cwrap(e.slice(1).map(rec.with(PREC.add)).join(` + `), pp < PREC.add),
     "sum": (pp, rec, e) => assert(false, message: "sum is unsupported in code."),
     "*": (pp, rec, e) => {
--- a/spec/neg.typ
+++ b/spec/neg.typ
@@ -33,13 +33,7 @@ where `cond` is a bit value (i.e., lies in ${0, 1}$)  described by an expression
 
 == Constraints
 For `neg` to equal $-#`x`$, both values must add to $0 mod 2^64$.
-Zooming in on the addition, we find that the carry values of adding the limbs must both equal to $1$, except when $#`x` = 0$, in which case they should both be $0$.
-
-To this end, we require $#`carry`_0$ is $1$, unless $#`x` = 0$ (@neg:c:carry) and
-that both limbs of `carry` are identical (@neg:c:identical_carry ensuring).
-Of course, both constraints are conditional on $#`cond` = 1$.
-Lastly, note that this @neg:c:carry implicitly enforces that $#`carry`_0 in {0, 1}$, while @neg:c:identical_carry ensures $#`carry`_1$ must also be binary.
-
+This is constrained through the following:
 #render_constraint_table(chip, config)
 
 == Note
--- a/spec/neg.typ
+++ b/spec/neg.typ
@@ -32,9 +32,33 @@ where `cond` is a bit value (i.e., lies in ${0, 1}$)  described by an expression
 #render_chip_assumptions(chip, config)
 
 == Constraints
-For `neg` to equal $-#`x`$, both values must add to $0 mod 2^64$.
-This is constrained through the following:
+We constrain this equality using two constraints:
 #render_constraint_table(chip, config)
+The constraints force the `carry` values to be fixed.
+Writing `carry`'s definition, we then find that
+$
+  #`neg`_0 &= 2^32 dot #`carry`_0 - (#`x as DWordWL`)_0
+ = cases(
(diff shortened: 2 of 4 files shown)


erik-3milabs:
Are you OK with merging 801f5ee along with the NEG template?

spec/neg.typ
```diff
@@ -0,0 +1,48 @@
...
+== Assumptions
+#render_chip_assumptions(chip, config)
+
+== Constraints
+For `neg` to equal $-#`x`$, both values must add to $0 mod 2^64$.
+Zooming in on the addition, we find that the carry values of adding the limbs must both equal to $1$, except when $#`x` = 0$, in which case they should both be $0$.
+
+To this end, we require $#`carry`_0$ is $1$, unless $#`x` = 0$ (@neg:c:carry) and
+that both limbs of `carry` are identical (@neg:c:identical_carry ensuring).
+Of course, both constraints are conditional on $#`cond` = 1$.
```
RobinJadoul:
```suggestion
Of course, both constraints are conditional on $#`cond` != 0$.
```

spec/src/neg.toml
```diff
@@ -0,0 +1,52 @@
...
+constraint = "$#`cond` => #`carry`_0 = #`carry`_1$"
+poly = ["*", "cond", ["-", ["idx", "carry", 0], ["idx", "carry", 1]]]
+ref = "neg:c:identical_carry"
+
+[[constraints.all]]
+kind = "template"
+tag = "IsZero"
+input = [["+", ["idx", "x", 0], ["idx", "x", 1]]]
+output = ["not", ["idx", "carry", 0]]
+multiplicity = "cond"
```
RobinJadoul, line 44: It's a template, so I think it should be cond instead?

```suggestion
cond = "cond"
```

RobinJadoul: Also we haven't actually specced `IsZero` yet?

spec/src/neg.toml
```diff
@@ -0,0 +1,52 @@
...
+[[constraints.all]]
+kind = "arith"
+constraint = "$#`cond` => #`carry`_0 = #`carry`_1$"
+poly = ["*", "cond", ["-", ["idx", "carry", 0], ["idx", "carry", 1]]]
+ref = "neg:c:identical_carry"
+
+[[constraints.all]]
+kind = "template"
+tag = "IsZero"
+input = [["+", ["idx", "x", 0], ["idx", "x", 1]]]
```
RobinJadoul, line 42: iirc, the idea of IsZero was that we could write that and do the summation internally in the template?

```suggestion
input = ["x"]
```

(Otherwise, it's just a conditional interaction with the ZERO bitwise table)

spec/src/neg.toml
```diff
@@ -0,0 +1,52 @@
...
+    {iter=1, poly=["*", ["^", 2, -32], ["+", ["idx", "x", 1], ["idx", "neg", 1], ["idx", "carry", 0]]]}
+]}
+
+
+[[assumptions]]
+desc = "`IS_WORD[x[i]]`"
+iter = ["i", 0, 1]
+
+[[assumptions]]
+desc = "`IS_BIT<cond>`"
```
RobinJadoul, line 33: I think we avoid assuming this, similar to other chips; unless we need it for the downstream IsZero template

spec/src/neg.toml
erik-3milabs: I propose we drop `IsZero` and `IsEqual` from the spec (#278) and instead lookup `ZERO` directly here. In which case it would be a `multiplicity` (again).

spec/src/neg.toml
erik-3milabs: Given that we're now feeding it in as the multiplicity for the `ZERO` chip, I believe we do need it here.

erik-3milabs:
@greptileai

spec/src/neg.toml
```diff
@@ -0,0 +1,53 @@
...
+
+
+[[constraint_groups]]
+name = "all"
+
+[[constraints.all]]
+kind = "interaction"
+tag = "ZERO"
+input = [["+", ["idx", "x", 0], ["idx", "x", 1]]]
```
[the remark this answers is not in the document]
erik-3milabs: No. To constrain `carry[0]`, one need only verify that `x[0]` and `x[1]` are zero; limbs `x[2]` and `x[3]` are irrelevant.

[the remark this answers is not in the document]
erik-3milabs: @greptileai

spec/src/neg.toml
```diff
@@ -0,0 +1,53 @@
...
+desc = "carries of the addition $#`neg` + #`x`$."
+def = {idx="i", polys=[
+    {iter=0, poly=["*", ["^", 2, -32], ["+", ["idx", ["cast", "x", "DWordWL"], 0], ["idx", "neg", 0]]]},
+    {iter=1, poly=["*", ["^", 2, -32], ["+", ["idx", ["cast", "x", "DWordWL"], 1], ["idx", "neg", 1], ["idx", "carry", 0]]]}
+]}
+
+
+[[assumptions]]
+desc = "`IS_WORD[x[i]]`"
+iter = ["i", 0, 1]
```
RobinJadoul: This is out of sync with x being DWordHL

spec/src/neg.toml
```diff
@@ -0,0 +1,53 @@
+name = "NEG"
+
+[[variables.input]]
+name = "x"
+type = "DWordHL"
+desc = "value to compute negation of"
+
+[[variables.input]]
+name = "cond"
+type = "BaseField"
```
RobinJadoul: Let's type this as Bit then too

spec/expr.typ
```diff
@@ -92,7 +92,7 @@
 #let expr_to_code = make_expr_formatter(
   (
     "idx": (pp, rec, e) => rec(PREC.MIN, e.at(1)) + `[` + rec(PREC.MAX, e.at(2)) + `]`,
-    "not": (pp, rec, e) => cwrap(`1 - ` + rec(PREC.not, e.at(1)), pp < PREC.not),
+    "not": (pp, rec, e) => cwrap(rec(PREC.not, 1) + ` - ` + rec(PREC.not, e.at(1)), pp < PREC.not),
```
RobinJadoul, line 95: Let's replicate the same change to `expr_to_math`

spec/neg.typ
```diff
@@ -0,0 +1,66 @@
...
+  #`neg`_1 &= 2^32 dot #`carry`_1 - (#`x as DWordWL`)_1 - #`carry`_0 = cases(
+  2^32 - (#`x as DWordWL`)_1 - 1 & "if" #`x` != 0,
+  0 & "if" #`x` = 0
+ )
+$
+Clearly, $#`neg` = 0$ when $#`x` = 0$ (and `cond` is set); for non-zero `x`, it holds that
+$
+  #`neg` 
+  &= #`neg`_0 + 2^32 dot #`neg`_1 \
+  &= (2^32 - (#`x as DWordWL`)_0) + 2^32 dot (2^32 - (#`x as DWordWL`)_1 - 1) \
```
RobinJadoul: The case split for neg_0 = 0 is very implicit in this (not incorrect, though, I think)

spec/src/neg.toml
erik-3milabs: Also, this is of variable type `condition` rather than `input`

RobinJadoul (approved):

PR #271, opened 2026-02-02
spec: Add initial tooling to check data formats, prepare for more elaborate type checking

RobinJadoul:
I've included some various fixed for minor issues encountered while checking the current spec files here, because they felt too minor to make a bunch of separate PRs for.

### DIFF
--- a/spec/src/config.toml
+++ b/spec/src/config.toml
@@ -4,36 +4,43 @@ version = 1
 [[variables.types]]
 label = "BaseField"
 subtypes = ["BaseField"]
+range = [0, 18446744069414584320]
 desc = "Variable that can assume any value in the base field."
 
 [[variables.types]]
 label = "Bit"
 subtypes = ["BaseField"]
+range = [0, 1]
 desc = "Variable that can only assume values in the set ${0,1}$."
 
 [[variables.types]]
 label = "B4"
 subtypes = ["BaseField"]
+range = [0, 15]
 desc = "Variable that can only assume values in the range $[0, 2^4)$."
 
 [[variables.types]]
 label = "Byte"
 subtypes = ["BaseField"]
+range = [0, 255]
 desc = "Variable that can only assume values in the range $[0, 2^8)$."
 
 [[variables.types]]
 label = "Half"
 subtypes = ["BaseField"]
+range = [0, 65535]
 desc = "Variable that can only assume values in the range $[0, 2^16)$."
 
 [[variables.types]]
 label = "B20"
 subtypes = ["BaseField"]
+range = [0, 1048575]
 desc = "Variable that can only assume values in the range $[0, 2^20)$."
 
 [[variables.types]]
 label = "Word"
 subtypes = ["BaseField"]
+range = [0, 4294967295]
 desc = "Variable that can only assume values in the range $[0, 2^32)$."
 
 [[variables.types]]
@@ -53,13 +60,9 @@ desc = """\
        """
 
 [[variables.types]]
-label = "B35"
-subtypes = ["BaseField"]
-desc = "Variable that can only assume values in the range $[0, 2^35)$."
-
-[[variables.types]]
 label = "B51"
 subtypes = ["BaseField"]
+range = [0, 2251799813685247]
 desc = "Variable that can only assume values in the range $[0, 2^51)$."
 
 [[variables.types]]
--- a/spec/src/config.toml
+++ b/spec/src/config.toml
@@ -44,22 +44,6 @@ range = [0, 4294967295]
 desc = "Variable that can only assume values in the range $[0, 2^32)$."
 
 [[variables.types]]
-label = "WordHL"
-subtypes = ["Half", "Half"]
-desc = """\
-       Variable that can only assume values in the range $[0, 2^32)$. \\
-       Represented as an array of two `Half` variables.\
-       """
-
-[[variables.types]]
-label = "WordBL"
-subtypes = ["Byte", "Byte", "Byte", "Byte"]
-desc = """\
-       Variable that can only assume values in the range $[0, 2^32)$. \\
-       Represented as an array of four `Byte` variables.\
-       """
-
-[[variables.types]]
 label = "B51"
 subtypes = ["BaseField"]
 range = [0, 2251799813685247]
@@ -99,6 +83,15 @@ desc = """\
        """
 
 [[variables.types]]
+label = "DWordWHH"
+subtypes = ["Half", "Half", "Word"]
+desc = """\
+       Variable that can only assume values in the range $[0, 2^64)$. \\
+       Represented as a `Word` and two `Half` variables.\
+       The `Word` is the *most* significant digit.
+       """
+
+[[variables.types]]
 label = "QuadHL"
 subtypes = ["Half", "Half", "Half", "Half", "Half", "Half", "Half", "Half"]
 desc = """\
@@ -112,15 +105,6 @@ subtypes = ["Word", "Word", "Word", "Word"]
 desc = """\
        Variable that can only assume values in the range $[0, 2^128)$. \\
        Represented as an array of four `Word` variables.\
-       """
-
-[[variables.types]]
-label = "DWordWHH"
-subtypes = ["Half", "Half", "Word"]
-desc = """\
-       Variable that can only assume values in the range $[0, 2^64)$. \\
-       Represented as a `Word` and two `Half` variables.\
-       The `Word` is the *most* significant digit.
        """
 
 [[variables.types]]
--- a/spec/src/config.toml
+++ b/spec/src/config.toml
@@ -19,7 +19,6 @@ desc = "Variable that can only assume values in the range $[0, 2^4)$."
 [[variables.types]]
 label = "Byte"
 subtypes = ["BaseField"]
-count = 1
 desc = "Variable that can only assume values in the range $[0, 2^8)$."
 
 [[variables.types]]
--- a/spec/src/config.toml
+++ b/spec/src/config.toml
@@ -4,7 +4,7 @@ version = 1
 [[variables.types]]
 label = "BaseField"
 subtypes = ["BaseField"]
-range = [0, 18446744069414584320]
+range = [0, "18446744069414584320"]
 desc = "Variable that can assume any value in the base field."
 
 [[variables.types]]
--- a/spec/src/cpu.toml
+++ b/spec/src/cpu.toml
@@ -343,6 +343,7 @@ name = "decode"
 kind = "interaction"
 tag = "DECODE"
 input = ["pc", "imm", "packed_decode"]
+multiplicity = 1
 
 
 [[constraint_groups]]
@@ -515,34 +516,40 @@ ref = "cpu:c:range_EBREAK"
 kind = "interaction"
 tag = "IS_BYTE"
 input = ["rs1"]
+multiplicity = 1
 
 [[constraints.range]]
 kind = "interaction"
 tag = "IS_BYTE"
 input = ["rs2"]
+multiplicity = 1
 
 [[constraints.range]]
 kind = "interaction"
 tag = "IS_BYTE"
 input = ["rd"]
+multiplicity = 1
 
 [[constraints.range]]
 kind = "interaction"
 tag = "IS_BYTE"
 input = [["idx", "arg1", "i"]]
 iter = ["i", 0, 7]
+multiplicity = 1
 
 [[constraints.range]]
 kind = "interaction"
 tag = "IS_BYTE"
 input = [["idx", "arg2", "i"]]
 iter = ["i", 0, 7]
+multiplicity = 1
 
 [[constraints.range]]
 kind = "interaction"
 tag = "IS_BYTE"
 input = [["idx", "res", "i"]]
 iter = ["i", 0, 7]
+multiplicity = 1
 
 
 [[constraint_groups]]
--- a/spec/src/cpu.toml
+++ b/spec/src/cpu.toml
@@ -647,7 +647,7 @@ prefix = "M"
 [[constraints.mem]]
 kind = "interaction"
 tag = "MEMW"
-input = [1, ["*", 2, "rs1"], "rv1", ["+", "timestamp", 0], 1, 0, 0]
+input = [1, ["*", 2, "rs1"], "rv1", ["+", "timestamp", ["cast", 0, "DWordWL"]], 1, 0, 0]
 output = "rv1"
 multiplicity = "read_register1"
 
@@ -661,7 +661,7 @@ iter = ["i", 0, 2]
 [[constraints.mem]]
 kind = "interaction"
 tag = "MEMW"
-input = [1, ["*", 2, "rs2"], "rv2", ["+", "timestamp", 1], 1, 0, 0]
+input = [1, ["*", 2, "rs2"], "rv2", ["+", "timestamp", ["cast", 1, "DWordWL"]], 1, 0, 0]
 output = "rv2"
 multiplicity = "read_register2"
 
@@ -675,13 +675,13 @@ iter = ["i", 0, 2]
 [[constraints.mem]]
 kind = "interaction"
 tag = "MEMW"
-input = [1, ["*", 2, "rd"], "rvd", ["+", "timestamp", 2], 1, 0, 0]
+input = [1, ["*", 2, "rd"], "rvd", ["+", "timestamp", ["cast", 2, "DWordWL"]], 1, 0, 0]
 multiplicity = "write_register"
 
 [[constraints.mem]]
 kind = "interaction"
 tag = "LOAD"
-input = [0, "res", ["+", "timestamp", 0], "memory_2bytes", "memory_4bytes", "memory_8bytes", "signed"]
+input = [0, "res", ["+", "timestamp", ["cast", 0, "DWordWL"]], "memory_2bytes", "memory_4bytes", "memory_8bytes", "signed"]
 output = "rvd"
 multiplicity = "LOAD"
 
@@ -689,14 +689,14 @@ multiplicity = "LOAD"
 [[constraints.mem]]
 kind = "interaction"
 tag = "MEMW"
-input = [0, "res", "rv2", ["+", "timestamp", 1], "memory_2bytes", "memory_4bytes", "memory_8bytes"]
+input = [0, "res", "rv2", ["+", "timestamp", ["cast", 1, "DWordWL"]], "memory_2bytes", "memory_4bytes", "memory_8bytes"]
 multiplicity = "STORE"
 
 # TODO: no types available, so no casting yet
 [[constraints.mem]]
 kind = "interaction"
 tag = "MEMW"
-input = [1, ["*", 2, 255], "next_pc", ["+", "timestamp", 1], 1, 0, 0]
+input = [1, ["*", 2, 255], "next_pc", ["+", "timestamp", ["cast", 1, "DWordWL"]], 1, 0, 0]
 output = "pc"
 multiplicity = ["not", "pad"]
 
--- a/spec/src/cpu.toml
+++ b/spec/src/cpu.toml
@@ -618,7 +618,7 @@ multiplicity = "SHIFT"
 [[constraints.alu]]
 kind = "template"
 tag = "ADD"
-input = ["pc", ["cast", ["+", ["*", 2, "c_type_instruction"], ["*", 4, ["not", "c_type_instruction"]]], "DWordWL"]]
+input = ["pc", ["*", ["+", ["*", 2, "c_type_instruction"], ["*", 4, ["not", "c_type_instruction"]]], ["cast", 1, "DWordWL"]]]
 output = ["cast", "res", "DWordWL"]
 cond = "JALR"
 
@@ -817,6 +817,6 @@ multiplicity = "branch_cond"
 [[constraints.misc]]
 kind = "template"
 tag = "ADD"
-input = ["pc", ["cast", ["+", ["*", 2, "c_type_instruction"], ["*", 4, ["not", "c_type_instruction"]]], "DWordWL"]]
+input = ["pc", ["*", ["+", ["*", 2, "c_type_instruction"], ["*", 4, ["not", "c_type_instruction"]]], ["cast", 1, "DWordWL"]]]
 output = "next_pc"
 desc = "Increment `pc` to `next_pc` if we're not branching"
--- a/spec/src/cpu.toml
+++ b/spec/src/cpu.toml
@@ -802,7 +802,7 @@ poly = ["+",
         ["-", "branch_cond"],
         "JALR",
         ["*", ["idx", "res", 0], ["not", "mp_selector"], "BLT"],
-        ["*", ["not", ["idx", "res", 0]], "mp_selector", "BLT"],
+        ["*", ["-", 1, ["idx", "res", 0]], "mp_selector", "BLT"],
         ["*", "is_equal", ["not", "mp_selector"], "BEQ"],
         ["*", ["not", "is_equal"], "mp_selector", "BEQ"]
     ]
--- a/spec/src/page.toml
+++ b/spec/src/page.toml
@@ -38,11 +38,13 @@ name = "all"
 kind = "interaction"
 tag = "IS_BYTE"
 input = ["init"]
+multiplicity = 1
 
 [[constraints.all]]
 kind = "interaction"
 tag = "IS_BYTE"
 input = ["fini"]
+multiplicity = 1
 
 [[constraints.all]]
 kind = "interaction"
--- a/spec/src/page.toml
+++ b/spec/src/page.toml
@@ -2,6 +2,12 @@ name = "PAGE"
 
 # Input
 
+# TODO: add `page` as "constant" column or smth
+[[variables.input]]
+name = "page"
+type = "DWordWL"
+desc = "Constant column containing the page base address; should be integrated into the constraints directly"
+
 [[variables.input]]
 name = "offset"
 type = "RowIndex"
@@ -28,7 +34,7 @@ desc = "The timestamp at which this address was last accessed"
 name = "address"
 type = "DWordWL"
 desc = "Adding `offset` to the page base address `page`. `page` is a constant with respect to a single instance of this table."
-def = ["+", "page", ["cast", "offset", "DWordWL"]]
+def = ["+", "page", ["*", "offset", ["cast", 1, "DWordWL"]]]
 
 
 [[constraint_groups]]
--- a/spec/tooling/chip.py
+++ b/spec/tooling/chip.py
@@ -457,7 +457,7 @@ class ConfigCategories:
             f"Something's not a string: {self.instantiated}",
         )
         reporter.asserts(
-            set(self.instantiated) < set(self.all),
+            set(self.instantiated) <= set(self.all),
             f"Instantiated not a subset of all: {self!r}",
         )
 
@@ -777,7 +777,7 @@ class ArithConstraint:
                 )
             else:
                 reporter.error(
-                    f"Non-scalar value for polynomial constraing: {self!r} {t}"
+                    f"Non-scalar value for polynomial constraint: {self!r} {t}"
                 )
 
         for t in all_iters(self.iters, env, lambda e: [self.poly.typecheck(e)]):
--- a/spec/tooling/chip.py
+++ b/spec/tooling/chip.py
@@ -1,32 +1,42 @@
-from dataclasses import dataclass
+from pathlib import Path
+import copy
 import sys
 import tomllib
-from typing import Optional, Union
+from collections.abc import Callable, Iterable
+from dataclasses import dataclass
+from typing import Optional, Union, Never
+
+def Bit_type():
+    return Type(None, "Bit")
 
 class ErrorReporter:
-    def __init__(self, location):
+    reported: bool
+    location: str
+    
+    def __init__(self, location: str):
         self.reported = False
         self.location = location
 
-    def update_location(self, loc):
+    def update_location(self, loc: str):
         self.reported = False
         self.location = loc
 
-    def error(self, message):
+    def error(self, message: str):
         self.reported = True
         print(f"ERROR {self.location}: {message}", file=sys.stderr)
 
-    def asserts(self, condition, message):
+    def asserts(self, condition: bool, message: str):
         if not condition:
             self.error(message)
 
 reporter = ErrorReporter("unknown")
 
-def assert_no_unexpected(data, possible_keys):
+def assert_no_unexpected(data: dict, possible_keys: Iterable[str]):
     for key in data.keys():
         reporter.asserts(key in possible_keys, f"Unexpected key: {key!r}")
-
-type Expr = (int
+        
+    
+type Expr = (LitExpr
             | VarExpr
             | IdxExpr
             | CastExpr
@@ -39,60 +49,180 @@ type Expr = (int
             | DummyExpr
             )
 
+# We can either have an explicit int literal (or const expression) or a known type
+# Returning 0 as dummy value should work in most cases, as constants can be used for
+# almost anything. The only exception being indexing.
+type TypeCheck = Type | int
+
+@dataclass
+class Environment:
+    config: "Config"
+    valmap: dict[str, int]
+    typemap: dict[str, "Type"]
+
+    def resolve_index(self, base: "Type", idx: int) -> TypeCheck:
+        if base.dimension is not None:
+            if not (0 <= idx < base.dimension):
+                reporter.error(f"Index out of range for {base!r}: {idx!r}")
+                idx = 0
+            if isinstance(base.base, str):
+                return Type(None, base.base)
+            else:
+                return base.base
+
+        assert isinstance(base.base, str), "We somehow made a type that's not an array, but has a non-str base"
+        typeconfigs = [tc for tc in self.config.variables.types if tc.label == base.base]
+        if len(typeconfigs) != 1:
+            reporter.error(f"Unable to resolve type: {base!r}")
+            return 0
+        typeconfig = typeconfigs[0]
+        if not (0 <= idx < len(typeconfig.subtypes)):
+            reporter.error(f"Index out of range for {base!r}: {idx!r}")
+            idx = 0
+        return typeconfig.subtypes[idx]
+
+def type_match(a: TypeCheck, b: TypeCheck, context: str) -> TypeCheck:
+    """Check that `a` and `b` are "compatible" TypeCheck values.
+    That is, either one of them is a constant, or the type is the same"""
+    # TODO: improve here; e.g. by allowing thing to match if their subtype is identical?
+    # Then would have to return the subtype to be sure?
+    # Maybe break everything down to subtypes, then it's purely structural matching?
+    if isinstance(a, int):
+        return b
+    if isinstance(b, int):
+        return a
+    reporter.asserts(a == b, f"Type mismatch between {a!r} and {b!r} [{context}]")
+    return 0
+
+@dataclass
+class LitExpr:
+    lit: int
+
+    def typecheck(self, _env: Environment) -> TypeCheck:
+        return self.lit
+
 @dataclass
 class VarExpr:
     name: str
+
+    def typecheck(self, env: Environment) -> TypeCheck:
+        if self.name in env.valmap:
+            return env.valmap[self.name]
+        if self.name in env.typemap:
+            return env.typemap[self.name]
+        reporter.error(f"Unknown variable: {self.name!r}")
+        return 0
 
 @dataclass
 class IdxExpr:
     base: Expr
     idx: Expr
 
+    def typecheck(self, env: Environment) -> TypeCheck:
+        base = self.base.typecheck(env)
+        idx = self.idx.typecheck(env)
+        if isinstance(base, int):
+            reporter.error(f"Trying to index a constant value: {self.base!r}")
+            return 0
+        if not isinstance(idx, int):
+            reporter.error(f"Trying to index with a non-constant: {self.idx!r}")
+            return 0
+        return env.resolve_index(base, idx)
+
 @dataclass
 class CastExpr:
     base: Expr
     type: "Type"
 
+    def typecheck(self, env: Environment) -> TypeCheck:
+        _base = self.base.typecheck(env)
+        # TODO? encode/list valid casts
+        return self.type
+
 @dataclass
 class MulExpr:
     factors: list[Expr]
 
+    def typecheck(self, env: Environment) -> TypeCheck:
+        t: TypeCheck = 0
+        for f in self.factors:
+            t = type_match(t, f.typecheck(env), repr(self))
+        return t
+
 @dataclass
 class AddExpr:
     terms: list[Expr]
+
+    def typecheck(self, env: Environment) -> TypeCheck:
+        t: TypeCheck = 0
+        for term in self.terms:
+            t = type_match(t, term.typecheck(env), repr(self))
+        return t
 
 @dataclass
 class SubExpr:
     head: Expr
     subs: list[Expr]
 
+    def typecheck(self, env: Environment) -> TypeCheck:
+        t = self.head.typecheck(env)
+        for term in self.subs:
+            t = type_match(t, term.typecheck(env), repr(self))
+        return t
+
 @dataclass
 class PowExpr:
     base: Expr
     exp: Expr
+
+    def typecheck(self, env: Environment) -> TypeCheck:
+        base = self.base.typecheck(env)
+        exp = self.exp.typecheck(env)
+        if not isinstance(base, int):
+            reporter.error(f"Invalid exponentiation with non-const base: {self.base!r}")
+            return 0
+        if not isinstance(exp, int):
+            reporter.error(f"Invalid exponentiation with non-const exponent: {self.exp!r}")
+            return 0
+        return base**exp
+            
 
 @dataclass
 class SumExpr:
     iter: "Iter"
     terms: Expr
 
+    def typecheck(self, env: Environment) -> TypeCheck:
+        t: TypeCheck = 0
+        for tc in self.iter.typecheck(env, lambda e: [self.terms.typecheck(e)]):
+            t = type_match(t, tc, repr(self))
+        return t
+
 @dataclass
 class NotExpr:
     inner: Expr
 
+    def typecheck(self, env: Environment) -> TypeCheck:
+        inner = self.inner.typecheck(env)
+        if isinstance(inner, int):
+            reporter.asserts(inner in {0, 1}, f"Not a bool passed to `not`: {self.inner!r}")
+            return 1 - inner
+        reporter.asserts(inner == Bit_type(), f"Not a bool passed to `not`: {self.inner!r}")
+        return Bit_type()
+
 @dataclass
 class DummyExpr:
-    pass
+    def typecheck(self, _env: Environment) -> TypeCheck:
+        return 0
 
-def build_expr(config: Optional["Config"], data) -> Expr:
-    # TODO
+def build_expr(config: Optional["Config"], data: object) -> Expr:
     # Does this need config, or do we delay any config-checking to when we use the expr?
     match data:
         case int(x):
-            return x
+            return LitExpr(x)
         case str(x):
             reporter.asserts(x.isidentifier(), f"Invalid identifier name for variable {x!r}")
-            return x
+            return VarExpr(x)
         case ["idx", x, y]:
             return IdxExpr(build_expr(config, x), build_expr(config, y))
         case ["cast", x, t]:
@@ -107,6 +237,7 @@ def build_expr(config: Optional["Config"], data) -> Expr:
         case ["^", base, exp]:
             return PowExpr(build_expr(config, base), build_expr(config, exp))
         case ["sum", ["=", str(var), start], stop, terms]:
+            assert config is not None
             return SumExpr(Iter(config, var, start, stop), build_expr(config, terms))
         case ["not", e]:
             return NotExpr(build_expr(config, e))
@@ -120,14 +251,30 @@ class Iter:
     start: Expr
     stop: Expr
 
-    def __init__(self, config, name, start, stop):
+    def __init__(self, config: "Config", name: str, start: object, stop: object):
         self.name = name
         reporter.asserts(isinstance(self.name, str), f"iter name is not a string: {self.name!r}")
         reporter.asserts(self.name.isidentifier(), f"Not a valid identifier: {self.name!r}")
         self.start = build_expr(config, start)
         self.stop = build_expr(config, stop)
 
-def iters_of(obj, name = None) -> list[Iter]:
+    def typecheck[T](self, env: Environment, callback: Callable[[Environment], Iterable[T]]) -> Iterable[T]:
+        start = self.start.typecheck(env)
+        if not isinstance(start, int):
+            reporter.error(f"Starting value of summation not a const: {self!r}")
+            start = 0
+        stop = self.stop.typecheck(env)
+        if not isinstance(stop, int):
+            reporter.error(f"Ending value of summation not a const: {self!r}")
+            stop = 0
+
+        for i in range(start, stop + 1):
+            old_env = copy.deepcopy(env)
+            env.valmap[self.name] = i
+            yield from callback(env)
+            env = old_env
+
+def iters_of(obj: dict, name = None) -> list[Iter]:
     """Return a list of iterators needed by `obj`. Taken from `iters` or `iter`.
     Prepend `name` to every iterator, if given.
     Adapted from the corresponding typst implementation."""
@@ -158,7 +305,7 @@ class Type:
     base: Union["Type", str]
     dimension: Optional[int]
 
-    def __init__(self, valid_types: Optional[list["TypeConfig"]], data):
+    def __init__(self, valid_types: Optional[list["TypeConfig"]], data: object):
         match data:
             case str(x):
                 reporter.asserts(valid_types is None or x in [tc.label for tc in valid_types], f"Invalid variable type: {x!r}")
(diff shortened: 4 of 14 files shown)


spec/src/cpu.toml
```diff
@@ -343,6 +343,7 @@ name = "decode"
 kind = "interaction"
 tag = "DECODE"
 input = ["pc", "imm", "packed_decode"]
+multiplicity = 1
```
erik-3milabs, line 346: Do we want to require `multiplicity` for interactions, or do we say it defaults to `1` when unspecified (just like you do with `pad`)?

spec/tooling/chip.py
erik-3milabs, line 1: This file was not formatted.

erik-3milabs: Also, you've neatly type-annotated pretty much all method parameters.... except for `data`. Why not?

spec/tooling/chip.py
```diff
@@ -0,0 +1,435 @@
...
+            reporter.error(f"Unknown expression: {other!r}")
+            return DummyExpr()
+
+@dataclass
+class Iter:
+    name: str
+    start: Expr
+    stop: Expr
+
+    def __init__(self, config, name, start, stop):
```
erik-3milabs: missing type annotations

erik-3milabs (approved):

spec/tooling/chip.py
RobinJadoul: Mostly because data is just the (sub)blob from the toml, which is pretty hard to type, and we're doing the actual structural checking in the code

spec/src/cpu.toml
RobinJadoul: I think it's better to make it explicit for multiplicities, since you have to be more aware when you're introducing things that act even on padding rows

spec/tooling/chip.py
erik-3milabs: I agree that providing the _specific_ (expected) type for the blob is not worth it. Though, stating that `data: Dict` would be somewhat useful.

> we're doing the actual structural checking in the code

Mostly, yes, but you're assuming just about everywhere that `data` is a `dict`. That is why I'm proposing to annotate `data` at least as such.

spec/tooling/chip.py
```diff
@@ -0,0 +1,839 @@
...
+        reporter.asserts(key in possible_keys, f"Unexpected key: {key!r}")
+
+
+@dataclass(frozen=True)
+class Range:
+    low: int
+    high: int
+
+    def is_bool(self):
+        return self.is_lit and self.low >= 0 and self.high <= 1
```
erik-3milabs: Unless you turn `is_lit` into an `@property`, this does not work.
```suggestion
        return self.is_lit() and self.low >= 0 and self.high <= 1
```

RobinJadoul: The `is_lit()` really shouldn't even be there...
I'll include it in the next update

RobinJadoul:
Still a couple open-ish questions, and a few things to improve, but I think we're at a good point to do a full review and maybe split off the remaining work (signature checking, "constant" columns (for PAGE), better casting checks) for later PRs

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+    def typecheck(self, env: Environment) -> Type:
+        base = self.base.typecheck(env)
+        # TODO? Detect more sorts of invalid casts
+        baselen = len(base) if isinstance(base, list) else 1
+        castlen = len(self.type) if isinstance(self.type, list) else 1
+        reporter.asserts(
+            baselen >= castlen or (isinstance(base, Range) and base.is_lit()),
+            f"Casting from fewer columns to more: {self!r} {base} {self.type}",
+        )
+        return self.type
```
erik-3milabs, line 152: I propose we additionally test that either
1. the range of `base` == range of `type`, or
2. `base` is a literal.

spec/src/config.toml
```diff
@@ -53,14 +59,10 @@ desc = """\
        Represented as an array of four `Byte` variables.\
        """
 
-[[variables.types]]
-label = "B35"
```
erik-3milabs, line 57: In addition to removing B35, I believe we can also drop `WordBL` and `WordHL`; neither seems to be in use (anymore).

spec/src/cpu.toml
```diff
@@ -668,28 +675,28 @@ iter = ["i", 0, 2]
...
 
 [[constraints.mem]]
 kind = "interaction"
 tag = "LOAD"
-input = [0, "res", ["+", "timestamp", 0], "memory_2bytes", "memory_4bytes", "memory_8bytes", "signed"]
+input = [0, "res", ["+", "timestamp", ["cast", 0, "DWordWL"]], "memory_2bytes", "memory_4bytes", "memory_8bytes", "signed"]
 output = "rvd"
 multiplicity = "LOAD"
 
 # TODO: no types available, so no casting yet
```
erik-3milabs, line 688: I think this comment (and similar ones) are no longer relevant?

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+            reporter.error(f"Subtraction of scalar and array types {self!r}")
+            return DEFAULT_TYPE
+        else:
+            return Range(a.low - b.high, a.high - b.low)
+
+    def typecheck(self, env: Environment) -> Type:
+        t = self.head.typecheck(env)
+        for term in self.subs:
+            t = self.type_match(t, term.typecheck(env))
+        return t
```
erik-3milabs, line 231: This does not properly treat the case that `self.subs = none`; there `t`'s range should be "negated".

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+        if isinstance(base, list) or not base.is_lit():
+            reporter.error(f"Invalid exponentiation with non-const base: {self.base!r}")
+            return DEFAULT_TYPE
+        if isinstance(exp, list) or not exp.is_lit():
+            reporter.error(
+                f"Invalid exponentiation with non-const exponent: {self.exp!r}"
+            )
+            return DEFAULT_TYPE
+        val = base.get_lit() ** exp.get_lit()
+        return Range(val, val)
```
erik-3milabs: One may provide invalid `float`s as input to `Range` when `exp` is _negative_.
I happen to know that I've used 2^-32 quite a couple times.

spec/src/config.toml
erik-3milabs: (and maybe put `DWordWHH` right after `DWordHHW`, so that the ordering of the variables makes more sense 🙈 )

spec/src/cpu.toml
```diff
@@ -611,7 +618,7 @@ multiplicity = "SHIFT"
 [[constraints.alu]]
 kind = "template"
 tag = "ADD"
-input = ["pc", ["cast", ["+", ["*", 2, "c_type_instruction"], ["*", 4, ["not", "c_type_instruction"]]], "DWordWL"]]
+input = ["pc", ["*", ["+", ["*", 2, "c_type_instruction"], ["*", 4, ["not", "c_type_instruction"]]], ["cast", 1, "DWordWL"]]]
```
erik-3milabs, line 621: While I agree that this solves the "casting from fewer columns to more" error, adding the 1 here feels like a regression, rather than progression. Perhaps that constraint should only be applied when `base` does not fit in the lowest limb of the type we're casting to?

spec/src/cpu.toml
```diff
@@ -810,6 +817,6 @@ multiplicity = "branch_cond"
 [[constraints.misc]]
 kind = "template"
 tag = "ADD"
-input = ["pc", ["cast", ["+", ["*", 2, "c_type_instruction"], ["*", 4, ["not", "c_type_instruction"]]], "DWordWL"]]
+input = ["pc", ["*", ["+", ["*", 2, "c_type_instruction"], ["*", 4, ["not", "c_type_instruction"]]], ["cast", 1, "DWordWL"]]]
```
erik-3milabs, line 820: see [this previous comment](https://github.com/yetanotherco/lambda_vm/pull/271/changes#r2772786633)

spec/src/page.toml
```diff
@@ -2,6 +2,12 @@ name = "PAGE"
 
 # Input
 
+# TODO: add `page` as "constant" column or smth
```
erik-3milabs, line 5: If you want this to be done in a separate PR, please turn it into an issue so we don't forget.

spec/src/page.toml
```diff
@@ -28,7 +34,7 @@ desc = "The timestamp at which this address was last accessed"
 name = "address"
 type = "DWordWL"
 desc = "Adding `offset` to the page base address `page`. `page` is a constant with respect to a single instance of this table."
-def = ["+", "page", ["cast", "offset", "DWordWL"]]
+def = ["+", "page", ["*", "offset", ["cast", 1, "DWordWL"]]]
```
erik-3milabs, line 37: see previous comment on the topic.

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+    config: "Config"
+    valmap: dict[str, Range]
+    typemap: dict[str, Type]
+
+
+@dataclass
+class LitExpr:
+    lit: int
+
+    def typecheck(self, _env: Environment) -> Type:
```
erik-3milabs, line 91: I find the name `typecheck`  rather confusing:
1. I'm having a hard time wrapping my head around the fact that we're using the word `Type` to describe the (nested list of) range(s) in which an expression may lie. Can't say I've thought of a better term yet, though.
2. the word "check" suggests (just) verification of something; I'd expect that to be a void that just errors when something is wrong. Here, it returns information that is clearly non-boolean. 
 
What would you think of something like `check_and_eval` or even just `evaluate`?

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+    config = Config.from_file(sys.argv[1])
+    if reporter.reported:
+        sys.exit(1)
+    reported = False
+    chips: list[Chip] = []
+    for file in sys.argv[2:]:
+        if file == sys.argv[1]:
+            continue
+        chips.append(Chip.from_file(config, file))
+        reported = reported or reporter.reported
```
erik-3milabs:
```suggestion
        reported |= reporter.reported
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+    if not reported:
+        for chip in chips:
+            reporter.update_location(f"Chip {chip.name}")
+            # TODO: do something with the signatures
```
erik-3milabs, line 986: Spin off as separate issue?

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+
+if __name__ == "__main__":
+    config = Config.from_file(sys.argv[1])
+    if reporter.reported:
+        sys.exit(1)
+    reported = False
+    chips: list[Chip] = []
+    for file in sys.argv[2:]:
+        if file == sys.argv[1]:
+            continue
```
erik-3milabs: why is this even here?

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+        constraint_groups = [grp["name"] for grp in data.get("constraint_groups", [])]
+        assert_no_unexpected(data.get("constraints", {}), constraint_groups)
+        self.constraints = [
+            build_constraint(config, con)
+            for group in data.get("constraints", {}).values()
+            for con in group
+        ]
+
+    @classmethod
+    def from_file(cls, config: Config, filename: str | Path) -> "Chip":
```
erik-3milabs: [typing.Self](https://docs.python.org/3/library/typing.html#typing.Self) was introduced in 3.11

```suggestion
    def from_file(cls, config: Config, filename: str | Path) -> Self:
```
and
```python
from typing import Self
```
up top

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+        reporter.update_location(str(filename))
+        return cls(config, tomllib.load(open(filename, "rb")))
+
+    @classmethod
+    def from_string(cls, config: Config, s: str) -> "Chip":
+        reporter.update_location("<string>")
+        return cls(config, tomllib.loads(s))
```
erik-3milabs, line 952: dead code

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+
+    @classmethod
+    def from_file(cls, filename: str | Path) -> "Config":
+        reporter.update_location(str(filename))
+        return cls(tomllib.load(open(filename, "rb")))
+
+    @classmethod
+    def from_string(cls, s: str) -> "Config":
+        reporter.update_location("<string>")
+        return cls(tomllib.loads(s))
```
erik-3milabs, line 521: dead code

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+
+@dataclass
+class InteractionSignature:
+    tag: str
+    input: list[Type]
+    output: Optional[Type]
+
+
+@dataclass
+class InteractionConstraint:
```
erik-3milabs: `TemplateConstraint` and `InteractionConstraint` share some commonalities. Perhaps it'd be cleaner to give them a shared superclass. After all, we designed the templates to have a similar look and feel to interactions.

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+    for file in sys.argv[2:]:
+        if file == sys.argv[1]:
+            continue
+        chips.append(Chip.from_file(config, file))
+        reported = reported or reporter.reported
+    if not reported:
+        for chip in chips:
+            reporter.update_location(f"Chip {chip.name}")
+            # TODO: do something with the signatures
+            (list(chip.typecheck()))
```
erik-3milabs: Why is this wrapped like this?
```suggestion
            chip.typecheck()
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+
+    def __init__(self, config: Config, category: str, data: dict):
+        assert_no_unexpected(data, set(Variable.__annotations__.keys()) | {"def"})
+        reporter.asserts("def" in data, f"Missing def for virtual column: {data!r}")
+        def_ = data.pop("def", {})
+        super().__init__(config, category, data)
+        self.def_ = VirtualDef(config, self.name, self.type, def_)
+
+    def typecheck(self, env: Environment) -> Type:
+        def structure_match(a: Type, b: Type):
```
erik-3milabs: The name suggests a void-function. Also, no return type spec'd
```suggestion
        def do_structures_match(a: Type, b: Type) -> bool:
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+                    f"Virtual column {self.name!r} not completely defined",
+                )
+            else:
+                for i in range(len(t)):
+                    check_covered(t[i], seen, indices + [i])
+
+        # Special case for better error messages
+        if isinstance(self.type, Range):
+            reporter.asserts(
+                len(self.def_.defs) == 1 and not self.def_.defs[0][0],
```
erik-3milabs: While the first index makes perfect sense (the 0th definition), the second had me touching in the dark for a bit.
I think storing `Iter` and the poly's `Expr` like this, makes the code hard to read.
A `dict` might be a touch overboard, given that there are only two values.
Perhaps `namedtuple` could be the solution here?

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+                        f"Ending value of virtual def iter not a const: {self!r}"
+                    )
+                    stop = Range(start.get_lit(), start.get_lit())
+
+                for i in range(start.get_lit(), stop.get_lit() + 1):
+                    if isinstance(expected, Range):
+                        reporter.error(
+                            f"Virtual definition has an iter for a scalar: {self!r}"
+                        )
+                        break
```
erik-3milabs: this test succeeds/fails irrespective of the `i`.

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+                    if not 0 <= i < len(expected):
+                        reporter.error(
+                            f"Virtual definition index {i} out of range for {expected}: {self!r}"
+                        )
+                        break
```
erik-3milabs: This check can be performed without needing to iterate over `i`.

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+        def handle_iters(
+            env: Environment,
+            iters: list[Iter],
+            poly: Expr,
+            expected: Type,
+            indices: list[int],
+            seen: set[tuple],
+        ):
+            if not iters:
+                asn = poly.typecheck(env)
```
erik-3milabs: not a fan of `asn`; it lacks meaning

erik-3milabs: also, I'd put this after the loop over `seen`

spec/tooling/chip.py
erik-3milabs, line 1: this file still lists a number of todo's. We should set up a issue to (jointly) track these (future) improvements

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+                            f"Virtual definition has an iter for a scalar: {self!r}"
+                        )
+                        break
+                    if not 0 <= i < len(expected):
+                        reporter.error(
+                            f"Virtual definition index {i} out of range for {expected}: {self!r}"
+                        )
+                        break
+                    old_val: Optional[Range] = env.valmap.get(it.name, None)
+                    env.valmap[it.name] = Range(i, i)
```
erik-3milabs: It happens quite often that we're constructing a `Range` with identical start and end value. Perhaps we should add a constructor that accepts only 1? `Constant(x)` for example, which just spits out a Range(x,x).

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+            return self.type_match(b, a)
+        else:
+            extrema = [x * y for x in [a.low, a.high] for y in [b.low, b.high]]
+            return Range(min(extrema), max(extrema))
+
+    def typecheck(self, env: Environment) -> Type:
+        t: Type = Range(1, 1)
+        for f in self.factors:
+            t = self.type_match(t, f.typecheck(env))
+        return t
```
erik-3milabs, line 176: this does not check that `self.factors` is not empty

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+            return Range(a.low + b.low, a.high + b.high)
+
+    def typecheck(self, env: Environment) -> Type:
+        if not self.terms:
+            reporter.error("Empty add")
+            return Range(0, 0)
+        t: Type = self.terms[0].typecheck(env)
+        for term in self.terms[1:]:
+            t = self.type_match(t, term.typecheck(env))
+        return t
```
erik-3milabs, line 202: This might benefit from
```suggestion
        types = [t.typecheck(env) for t in self.terms]
        return functools.reduce(lambda t1, t2: self.type_match(t1, t2), types)
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+
+@dataclass
+class VarExpr:
+    name: str
+
+    def typecheck(self, env: Environment) -> Type:
+        if self.name in env.valmap:
+            return env.valmap[self.name]
+        if self.name in env.typemap:
+            return env.typemap[self.name]
```
erik-3milabs, line 103: could you elaborate on why there is both an `valmap` and a `typemap`? Why could they not be merged?

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+@dataclass
+class NotExpr:
+    inner: Expr
+
+    def typecheck(self, env: Environment) -> Type:
+        inner = self.inner.typecheck(env)
+        if isinstance(inner, list) or not inner.is_bool():
+            reporter.asserts(
+                inner in {0, 1}, f"Not a bool passed to `not`: {self.inner!r}"
+            )
```
erik-3milabs: How could `inner in {0, 1}`? `0` and `1` are not `Type`s.
I suppose one could simply `reporter.error` here.

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+            isinstance(self.name, str), f"name is not a string: {self.name!r}"
+        )
+        reporter.asserts(self.name.isidentifier(), f"Invalid identifier: {self.name!r}")
+        self.variables = [
+            (Variable if cat != "virtual" else VirtualVariable)(config, cat, var)
+            for cat, vars in data["variables"].items()
+            for var in vars
+        ]
+        self.assumptions = [
+            Assumption(config, asm) for asm in data.get("assumptions", [])
```
erik-3milabs: To me, `asm` is short for "assembly". I think simply using `a` would be better here.

```suggestion
            Assumption(config, a) for a in data.get("assumptions", [])
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+        ]
+        constraint_groups = [grp["name"] for grp in data.get("constraint_groups", [])]
+        assert_no_unexpected(data.get("constraints", {}), constraint_groups)
+        self.constraints = [
+            build_constraint(config, con)
+            for group in data.get("constraints", {}).values()
+            for con in group
+        ]
```
erik-3milabs, line 942:
```suggestion
        self.constraints = [
            build_constraint(config, constraint)
            for group in data.get("constraints", {}).values()
            for constraint in group
        ]
```

spec/tooling/chip.py
erik-3milabs, line 1: It seems that you often use two or three-letter names for (temporary/iterator) variables (e.g., `tp`, `con`, `grp`, `asm`, `inp`). What is your reason for using three-letter names, rather than 
1. single-letter names, or
2. just spelling things out in full?

The reason I'm addressing this, is that I prefer verbose names: plain and simple; I don't need to guess what they represent. 

Of course, I acknowledge that verbose names can at times be (too) long / conflict with operator names (`type`, e.g.). In those cases, I try to use a single-letter name. This illustrates they have a small scope, and if I really need to figure out what they mean, I look at their definition. The three letter variables you're introducing feel like a puzzle that I don't want to (also) have to solve; the code itself is complex enough.

E.g.: 
- `tp` is toilet paper not `type`, 😆 
- `con` is too often used to represent `connection`,
- I read `grp` as `grip`, not `group`,
- I interpret `asm` as `assembly`, not `assumption`,
- `asn` =... assertion? assigned/assignee? Still don't get it.

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
+from pathlib import Path
+import sys
+import tomllib
+from collections.abc import Callable, Iterable
+from dataclasses import dataclass
+from typing import Optional, Never
+
+
+class ErrorReporter:
```
erik-3milabs, line 9: I see that by working with this `ErrorReporter`, the tool can now keep running after finding a mistake, thus allowing one to find multiple mistakes in one execution. I wonder whether this is actually desirable.

The reason I do, is that after finding a mistake, the script often attempts to "fix" the mistake by assigning some default value. But who is to say that this default value is reasonably chosen? It may very well lead to both false-positives and false-negatives. The presence of false-negatives suggest the tools must be run anyway after fixing a mistake. False-positives may lead to lines being fixed unnecessarily as well as the root-cause being hard to detect.

Currently, it is common to see
```python
if not ...
     reporter.error(
            ...
     )
     <set-some-default-value>
```
which I find both cluttering (these 5 lines could be represented using a single assertion) as well as harder to read (because the boolean statement is inverted).

As such, I'd be inclined to treat this tool more as a _compiler_ (which exits at the first mistake) rather than a _linter_/_formatter_.

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+@dataclass
+class AddExpr:
+    terms: list[Expr]
+
+    def type_match(self, a: Type, b: Type) -> Type:
+        if isinstance(a, list) and isinstance(b, list):
+            if len(a) != len(b):
+                assert False
+                reporter.error(f"Adding array types of different length {self!r}")
+                return [DEFAULT_TYPE for _ in range(len(b))]
```
erik-3milabs: The `assert False` makes the other two lines unreachable

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+        # While it's tempting to replace this loop by an assignment of Range(start, stop + 1) to self.name
+        # that would break both detection of literals, and narrowing down to the correct type for indexing
+        # heterogenous array types
+        for i in range(start.get_lit(), stop.get_lit() + 1):
+            old_val: Optional[Range] = env.valmap.get(self.name, None)
+            env.valmap[self.name] = Range(i, i)
+            yield from callback(env)
+            env.valmap.pop(self.name)
+            if old_val is not None:
+                env.valmap[self.name] = old_val
```
erik-3milabs:
```suggestion
        for i in range(start.get_lit(), stop.get_lit() + 1):
            yield from callback({**env, self.name: Range(i, i))
```

If you're concerned about the performance (which I don't think should be the case):

If a `None` value inside of `env.valmap` has a meaning, then you're destroying those values by not re-inserting them. Otherwise, this can we written more concisely as
```suggestion
        for i in range(start.get_lit(), stop.get_lit() + 1):
            old_val: Optional[Range] = env.valmap.get(self.name, None)
            env.valmap[self.name] = Range(i, i)
            yield from callback(env)
            env.valmap[self.name] = old_val
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+        else:
+            self.range = None
+            self.subtypes = [lookup(tp) for tp in data["subtypes"]]
+        self.desc = data["desc"]
+        self.preprocessed = data.get("preprocessed", False)
+
+    def as_type(self):
+        if self.range is not None:
+            return self.range
+        return self.subtypes[:]
```
erik-3milabs:
```suggestion
    def as_type(self):
        return self.range or self.subtypes[:]
```
but you'd have to slap a return type hint on this, so that people don't confuse it with `bool`

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+        self.all = data["all"]
+        self.instantiated = data["instantiated"]
+        reporter.asserts(
+            all(isinstance(v, str) for v in self.all),
+            f"Something's not a string: {self.all}",
+        )
+        reporter.asserts(
+            all(isinstance(v, str) for v in self.instantiated),
+            f"Something's not a string: {self.instantiated}",
+        )
```
erik-3milabs, line 458: `substantiated` should be a subset of `all`. Please add this check.

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+    desc: str
+    preprocessed: bool
+
+    def __init__(self, default_name: str, lookup: Callable[[str], Type], data: dict):
+        assert_no_unexpected(data, type(self).__annotations__.keys())
+        self.label = data["label"]
+        if "range" in data:
+            reporter.asserts(
+                data["subtypes"] == [default_name],
+                f"Specified a range non a non-base composite type: {data!r}",
```
erik-3milabs: minor typo
```suggestion
                f"Specified a range on a non-base composite type: {data!r}",
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+    types: list[TypeConfig]
+    categories: ConfigCategories
+
+    def __init__(self, data: dict):
+        assert_no_unexpected(data, type(self).__annotations__.keys())
+        self.types = []
+        base_type = None
+        for tp in data["types"]:
+            if base_type is None:
+                base_type = tp["label"]
```
erik-3milabs: this code sneakily/implicitly assumes the first element is the base element. Either
1. we should make this assumption more explicit

```suggestion
        base_type = data["types"][0]["label"]
        for tp in data["types"]:
```

2. we should find a way for things to be order-agnostic (e.g., filter for the element that has itself as unique subtype)

```suggestion
        base_types = list(filter(lambda t: len(t.subtypes) == 1 and t.subtypes[0] == t.label, data["types"]))
        reporter.assert(len(base_types) == 1, f"Invalid number of base types: {base_types})
        base_type = base_types[0]
        for tp in data["types"]:
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+            self.types.append(TypeConfig(base_type, self.lookup_type, tp))
```
erik-3milabs, line 476: This construction also assumes the types listed in `config.toml` are (somewhat) ordered. Do we want to state this assumption somewhere?

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+
+
+@dataclass
+class VirtualVariable(Variable):
+    def_: VirtualDef
+
+    def __init__(self, config: Config, category: str, data: dict):
+        assert_no_unexpected(data, set(Variable.__annotations__.keys()) | {"def"})
+        reporter.asserts("def" in data, f"Missing def for virtual column: {data!r}")
+        def_ = data.pop("def", {})
```
erik-3milabs, line 607: you just asserted "def" to be in data


```suggestion
        def_ = data.pop("def")
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+                reporter.asserts(
+                    structure_match(asn, expected),
+                    f"Invalid structure for definition to virtual column: {self!r}",
+                )
+                # Check type fits?
+
+                seen.add(tuple(indices))
+            else:
+                it, *its = iters
+                # Some duplicated code/concepts from Iter.typecheck
```
erik-3milabs, line 651: perhaps that could be merged then?

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+                        reporter.error(
+                            f"Virtual definition index {i} out of range for {expected}: {self!r}"
+                        )
+                        break
+                    old_val: Optional[Range] = env.valmap.get(it.name, None)
+                    env.valmap[it.name] = Range(i, i)
+                    handle_iters(env, its, poly, expected[i], indices + [i], seen)
+                    env.valmap.pop(it.name)
+                    if old_val is not None:
+                        env.valmap[it.name] = old_val
```
erik-3milabs:
```suggestion
                    handle_iters({**env, it.name: Range(i,i)}, its, poly, expected[i], indices + [i], seen)
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+
+        def is_covered(seen: set[tuple], indices: list[int]) -> bool:
+            for s in seen:
+                if len(s) <= len(indices) and s == tuple(indices[: len(s)]):
+                    return True
+            return False
```
erik-3milabs, line 692:
```suggestion
        def is_covered(seen: set[tuple], indices: list[int]) -> bool:
            for s in seen:
                if len(s) <= len(indices) and s == tuple(indices[:len(s)]):
                    return True
            return False
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+
+        def check_covered(t: Type, seen: set[tuple], indices: list[int]):
+            if isinstance(t, Range):
+                reporter.asserts(
+                    is_covered(seen, indices),
+                    f"Virtual column {self.name!r} not completely defined",
+                )
+            else:
+                for i in range(len(t)):
+                    check_covered(t[i], seen, indices + [i])
```
erik-3milabs:
```suggestion
                for i, elt in enumerate(t):
                    check_covered(elt, seen, indices + [i])
```

spec/tooling/chip.py
erik-3milabs: also

```suggestion
        if isinstance(a, list) and isinstance(b, list):
            if len(a) != len(b):
                assert False
                reporter.error(f"Adding array types of different length {self!r}")
                return [DEFAULT_TYPE for _ in b]
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+@dataclass
+class SubExpr:
+    head: Expr
+    subs: list[Expr]
+
+    def type_match(self, a: Type, b: Type) -> Type:
+        if isinstance(a, list) and isinstance(b, list):
+            if len(a) != len(b):
+                reporter.error(f"Subtracting array types of different length {self!r}")
+                return [DEFAULT_TYPE for _ in range(len(a))]
```
erik-3milabs:
```suggestion
                return [DEFAULT_TYPE for _ in a]
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+@dataclass
+class SumExpr:
+    iter: "Iter"
+    terms: Expr
+
+    def type_match(self, a: Type, b: Type) -> Type:
+        if isinstance(a, list) and isinstance(b, list):
+            if len(a) != len(b):
+                reporter.error(f"Summing array types of different length {self!r}")
+                return [DEFAULT_TYPE for _ in range(len(b))]
```
erik-3milabs:
```suggestion
                return [DEFAULT_TYPE for _ in b]
```

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+
+        def check_includes_zero(t: Type):
+            if isinstance(t, Range):
+                reporter.asserts(
+                    t.low <= 0 <= t.high,
+                    f"Unsatisfiable constraint, 0 not in range: {self!r} {t}",
+                )
+            else:
+                for sub in t:
+                    check_includes_zero(sub)
```
erik-3milabs: shouldn't all constraints be operating on the `BaseField`? I'd say that if we need to recurse here, there is a mistake in the constraint.

spec/tooling/chip.py
```diff
@@ -0,0 +1,944 @@
...
+            if len(a) != len(b):
+                assert False
+                reporter.error(f"Adding array types of different length {self!r}")
+                return [DEFAULT_TYPE for _ in range(len(b))]
+            return [self.type_match(x, y) for x, y in zip(a, b)]
+        elif isinstance(a, list) or isinstance(b, list):
+            reporter.error(f"Adding of scalar and array types {self!r}")
+            return DEFAULT_TYPE
+        else:
+            return Range(a.low + b.low, a.high + b.high)
```
erik-3milabs, line 193: This implicitly assumes `low < high`, but we don't test for that.

spec/tooling/chip.py
RobinJadoul: Doesn't work, due to things like one-hot assumptions

spec/src/cpu.toml
RobinJadoul: Some still are, until we get to check the signatures, so I was planning on cleaning them all up at once

spec/src/cpu.toml
RobinJadoul: Not sure if that'll work or not, because ranges are more conservative than we sometimes know them to be; e.g. (the classic) one-hot sums

spec/tooling/chip.py
RobinJadoul: It's about to become a PR soon after this one is merged, so fairly sure I'll remember it ¯\_(ツ)_/¯

spec/tooling/chip.py
RobinJadoul: Because I call it as `python chip.py ../src/config.toml ../src/*.toml` out of convenience

spec/tooling/chip.py
RobinJadoul: The outer parens because I was printing during debugging, the list because it's a generator and it needs to materialize

spec/tooling/chip.py
RobinJadoul: iirc, most of the todos were more of "open questions" for things to think about as either potential improvements, or whether they're worth it

spec/tooling/chip.py
RobinJadoul: At best, I'd call it Range.lit or smth then, but then again, it's not that counterintuitive to do it like this; and not common enough to bother me enough to introduce it while writing the code :)

spec/tooling/chip.py
RobinJadoul: An empty product as 1 is technically valid; though _probably_ (?) unintended

spec/tooling/chip.py
RobinJadoul: I wanted valmap to have precedence over typemap, since that's the hierarchy you'd get when shadowing a column name in an iter

spec/tooling/chip.py
RobinJadoul: Yep, that seems to be a remnant from the time before the big Range refactor

spec/tooling/chip.py
RobinJadoul: Mostly, I do this for short-lived variables that have necessary context immediately nearby.
It's a tradeoff over single-letter ones, giving a bit more intuition without blowing up the noise of full words.
Some things like `grp` occur in other contexts, like the UNIX group database, or `asn` as a latex alias for leftarrow (assign/assigned/...).

spec/tooling/chip.py
RobinJadoul: Even compilers attempt to do error recovery to report as many errors as possible/reasonable.
I started off with regular assertions, but while developing and testing it out on the existing bugs, I quickly got annoyed by having to re-run the script every time again only to hit a minor variation of the thing I just fixed again in almost the same context. I rather have a list of things where I can pick off a few or related things at a time, and then re-run the check.
For false negatives, that's true, but that's also why most default values attempt to be conservative.
False positives are less of an issue, imo, since (like for standard compiler output), you usually want to start as far up as possible in the output, and hence you come across the initial failure earlier, which is more likely to be the root cause of any cascading failures. The spirit of the defaults is "it's most likely to have meant this", and it produces far less noise if it gets that right.
The tool must indeed be run again after fixing everything reported to be certain, but again we're more likely to have caught the majority of issues, and more likely to have treated entire classes of issues at once.

spec/tooling/chip.py
RobinJadoul: I'd just prefer not leaving `None`s lying around in the map afterwards; but probably the splat you suggested below is a better approach

spec/tooling/chip.py
RobinJadoul: I don't generally love the short-circuiting hacks, but it could make sense in the context of a short thing like this

spec/tooling/chip.py
RobinJadoul: I think it is the most natural order that  anyone would be writing new types, and introducing a statement to that fact may make the mental load up-front bigger (actively trying to remember what you'd be doing naturally anyway), and the resulting error message if you do violate the assumption is likely clear enough as an indicator anyway

spec/tooling/chip.py
RobinJadoul: Trust me, I tried, but that becomes a giant mess in threading state and typing everything.
I thought I added some of that reasoning to the comment as well, but apparently not. Maybe it's floating around in a commit message then, maybe it just got lost.

spec/tooling/chip.py
RobinJadoul: That space was inserted by ruff, not me :)

spec/tooling/chip.py
RobinJadoul: Hmm, maybe. I was considering a case where we want e.g. a DWordWL to be zero as shorthand usecase without introducing iters.

spec/tooling/chip.py
RobinJadoul: I think this is enforced everywhere as invariant, except the input reading from config.toml; maybe a check there then

spec/tooling/chip.py
RobinJadoul: idk...
While typechecking here is done as a sort of abstract evaluation, I don't love tying this to the mental concept of evaluating constraints.
I toyed with the idea of naming it `typeof` or similar for a while, but in the end, typecheck remained the most intuitive to me, as it does checks and assigns a type to the expression; both of which are what typechecking is about (ignoring any kind of advanced type system with inference and unification, ofc).

spec/tooling/chip.py
RobinJadoul: Hmm, very good point. I'm not sure how to best address it, though. Maybe some fractional representation, but that's definitely messed up.
The alternative is to work mod `p`, I suppose; which still works and keeps is a more or less clean `lit`

spec/tooling/chip.py
RobinJadoul: Weird, I thought that gave issues when I tried it before :shrug:

spec/tooling/chip.py
RobinJadoul: But plausibly useful for future consumers and not too big to be in the way

spec/tooling/chip.py
RobinJadoul: Not worth needing functools for this, imo

spec/tooling/chip.py
RobinJadoul: It's reporter-asserted, so it could still be missing and need a fallback

spec/tooling/chip.py
RobinJadoul: After checking, we don't do that anywhere, so reverting to error instead now

spec/src/cpu.toml
RobinJadoul: I actually kinda like the explicitness of indicating where to inject a value into a bigger type

spec/src/page.toml
erik-3milabs: For future reference: #303

spec/tooling/chip.py
```diff
@@ -0,0 +1,988 @@
...
+            f"Something's not a string: {self.all}",
+        )
+        reporter.asserts(
+            all(isinstance(v, str) for v in self.instantiated),
+            f"Something's not a string: {self.instantiated}",
+        )
+        reporter.asserts(
+            set(self.instantiated) < set(self.all),
+            f"Instantiated not a subset of all: {self!r}",
+        )
```
erik-3milabs, line 462: Technically, it can be the same set...

```suggestion
        reporter.asserts(
            set(self.instantiated) <= set(self.all),
            f"Instantiated not a subset of all: {self!r}",
        )
```

spec/tooling/chip.py
erik-3milabs: I see you chose the first option (rather than filtering for the element that has itself as subtype). Why did you do this? I though the second solution was a touch neater.

spec/tooling/chip.py
```diff
@@ -0,0 +1,988 @@
...
+
+        def check_includes_zero(t: Type):
+            if isinstance(t, Range):
+                reporter.asserts(
+                    t.low <= 0 <= t.high,
+                    f"Unsatisfiable constraint, 0 not in range: {self!r} {t}",
+                )
+            else:
+                reporter.error(
+                    f"Non-scalar value for polynomial constraing: {self!r} {t}"
```
erik-3milabs:
```suggestion
                    f"Non-scalar value for polynomial constraint: {self!r} {t}"
```

spec/tooling/chip.py
erik-3milabs: This does hinge on the assumption that
1. plans don't change, and
2. you'll be the one working on this ;-)

spec/tooling/chip.py
erik-3milabs: ahhh, I didn't know that shell would expand that `*`. nice

spec/tooling/chip.py
erik-3milabs: In its current incarnation, `InteractionLike` is not independent of `Template` and `Interaction`. I'd be inclined to only put the shared (unparameterized) code in `InteractionLike` and the subclasses to deal with their own parts.

spec/tooling/chip.py
erik-3milabs: I see most of them have been cleaned up. Perhaps the cond/multiplicity check question can be spun off?

spec/tooling/chip.py
erik-3milabs: Apparently, [these are called](https://en.wikipedia.org/wiki/Interval_(mathematics)#Degenerate_intervals) "degenerate" (?) 

Can't say that I'm a fan of `Range.lit`. Why would you see a constant as a literal? (or does `lit` mean something else?)

spec/tooling/chip.py
erik-3milabs: Not checking it would permit `["^", 2, ["*"]]`, which would be displayed as `2` (actually, no, since we only permit constant exponents, but OK), while being evaluated as `1`.

spec/tooling/chip.py
erik-3milabs: could you elaborate on what `valmap` and `typemap` are supposed to contain? I'm having a hard time gathering this from the code.

spec/tooling/chip.py
erik-3milabs: Got an example?

spec/tooling/chip.py
RobinJadoul: What's the dependence? I thought I managed to separate all important bits by injecting the relevant variables in the subclasses

spec/tooling/chip.py
RobinJadoul: Maybe, I just don't really know _what_ we can check about them

spec/tooling/chip.py
RobinJadoul: If it's a constant, it's equivalent to a literal; and thinking about things like `["*", 2, 255]` as constants rather than literals makes little difference imo.
It just fit well with the `is_lit` that existed before

spec/tooling/chip.py
RobinJadoul: Note, this was also fixed

RobinJadoul: Surely this would be both evaluated and displayed as 2, since the empty product is 1 and you get 2^1

spec/tooling/chip.py
RobinJadoul: `valmap` contains constant values introduced by iters and sums, while `typemap` is the base layer that contains all the types introduced by the columns of the chip itself

spec/tooling/chip.py
RobinJadoul: We already have introduced the assumption of dependence from the subtype arrays needing the type to be introduced before.
This just seemed like the simpler version for that; and we're highly unlikely to ever move the basefield out of first position anyway imo

spec/tooling/chip.py
RobinJadoul: If you're e.g. computing `b * x + (1 - b) * y` for some bit `b` and (say) `DWordBL`s `x` and `y`, the range goes outside Byte range, so casting to `DWordWL` would fail; because we can't see into the `b` and `1 - b` being mutually exclusive
The situation only gets worse if you get things like `(1 - STORE - LOAD) * x + (1 - BLT - BEQ) * y` or similar things in the CPU

spec/tooling/chip.py
erik-3milabs: There is an _architectural_  dependence of sorts.

What I'm trying to say is that `InteractionLike`'s `conditional_name`, `conditional_required` and `signature` feel like a forced attempt to merge `Template` and `Interaction` into the same box, making the box a touch too generic in the process. I'd be inclined to offload the tests related to these three parameters to `Template` and `Interaction`, having the super class only for the stuff that is truly shared.

spec/tooling/chip.py
erik-3milabs: Please provide your definition of literal. I'm thinking of [this](https://en.wikipedia.org/wiki/Literal_(mathematical_logic)), which, in my understanding, does not combine with the concept of constants.

spec/tooling/chip.py
erik-3milabs: Sorry, I fudged up. Of course, `["*"]` evaluates to 1.

Still, you'd have that `["-", ["*"], 1]` would be displayed as `-1`, while it would evaluate to `0`.

Anyway, this is bikeshedding; you already added the non-empty test.

spec/tooling/chip.py
erik-3milabs: I'm leaving this open for easy retrieval if we wish to continue this discussion later.

spec/tooling/chip.py
RobinJadoul: idk if for just the overlap of what, tag, input and output, adding in a base class makes that much sense then.
At the same time, from a using it point of view, the cond and multiplicity are essentially the same, just not on a lower level and that's why we renamed cond.  From some point of view, a template is "interaction-light".
And if we use that identification, the signature and the rest of the overlap make some sense, imo.

In the end, I also don't expect this to be extended much beyond what it is now, as a system.

spec/tooling/chip.py
erik-3milabs: Having given it a bit more thought, I'm mostly annoyed by the cluttered notation of it all.
Sadly, I cannot think of a better way to make this happen :confused:

spec/tooling/chip.py
RobinJadoul: I'm thinking "integer literal" as in `42`, here. Which is a sub-case of constant if you want to think of it that way, but given that our abstract evaluation gives us constant folding, the difference becomes more blurry.

spec/tooling/chip.py
```diff
@@ -0,0 +1,988 @@
...
+        return DEFAULT_TYPE
+
+
+@dataclass
+class ArrExpr:
+    elems: list[Expr]
+
+    def typecheck(self, env: Environment) -> Type:
+        reporter.asserts(self.elems != [], f"Empty array: {self!r}")
+        return [e.typecheck(env) for e in self.elems]
```
erik-3milabs, line 114: how did this guy end up in here, rather than #295?

RobinJadoul: Hmm, probably somewhere with the stacked branches, not exactly sure...

RobinJadoul: Ah, yes, I see; I address the empty product thing, and while doing so, added the empty array check, which pulled the surrounding things in

spec/tooling/chip.py
erik-3milabs: That clears some things up.
I don't think that the term "integer literal" is helpful here. While I agree that these constraints "teeter on the edge of maths and computer science", I'd be inclined to treat them as mathematical concepts.
And indeed, given that we're folding constants, the case for viewing these things as integer literals is, imo, thin.

While I don't think using the term "degenerate" will do us any good, I do think that replacing `Range.lit` by `Constant` or `Singleton` would improve the code base.

RobinJadoul: The original is_lit was there to match the LitExpr (which is far more CS, though type systems and languages definitely still are too). But I can rename everything to is_const and const, sure

RobinJadoul: Adding it as a separate type (or any other way I can think of to end up with a capital C) doesn't make too much sense though, imo

erik-3milabs (approved):

PR #280, opened 2026-02-03
spec: signatures

erik-3milabs:
List all interaction signatures. Used as reference material, as well as source of truth for validating lookup signatures in chips

### DIFF
--- a/spec/src/signatures.toml
+++ b/spec/src/signatures.toml
@@ -93,7 +93,7 @@ output = "DWordWL"
 [[signatures]]
 tag = "ECALL"
 kind = "interaction"
-input = ["DWordWL", "Word"]
+input = ["DWordWL", "DWordWL"]
 
 # AND_BYTE[res; X, Y]
 [[signatures]]
--- a/spec/src/signatures.toml
+++ b/spec/src/signatures.toml
@@ -26,7 +26,7 @@ cond = "BaseField"
 tag = "NEG"
 kind = "template"
 input = ["DWordHL"]
-output = "DWordDL"
+output = "DWordWL"
 cond = "Bit"
 
 # SIGN<sign; X, signed>
--- a/spec/src/signatures.toml
+++ b/spec/src/signatures.toml
@@ -82,7 +82,7 @@ kind = "interaction"
 input = ["DWordHL", "DWordHL", "Bit", "Bit"]
 output = "DWordWL"
 
-# LOAD[res; base_address, timestamp, read2, read4, read8]
+# LOAD[res; base_address, timestamp, read2, read4, read8, signed]
 [[signatures]]
 tag = "LOAD"
 kind = "interaction"
--- a/spec/src/signatures.toml
+++ b/spec/src/signatures.toml
@@ -1,0 +1,170 @@
+# cond => IS_BIT<X>
+[[signatures]]
+tag = "IS_BIT"
+kind = "template"
+input = ["BaseField"]
+cond = "Bit"
+
+# cond => ADD<sum; lhs, rhs>
+[[signatures]]
+tag = "ADD"
+kind = "template"
+input = ["DWordWL", "DWordWL"]
+output = "DWordWL"
+cond = "BaseField"
+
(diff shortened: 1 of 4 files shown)


spec/src/signatures.toml
```diff
@@ -0,0 +1,170 @@
...
+kind = "interaction"
+input = ["DWordHL", "DWordHL", "Bit", "Bit"]
+output = "DWordWL"
+
+# LOAD[res; base_address, timestamp, read2, read4, read8]
+[[signatures]]
+tag = "LOAD"
+kind = "interaction"
+input = ["DWordWL", "DWordWL", "Bit", "Bit", "Bit", "Bit"]
+output = "DWordWL"
```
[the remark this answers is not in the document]
erik-3milabs: While it is correctly observed that there is a mismatch between the number of input parameters and the number of input types, this indicates that there is missing input parameter, rather than a superfluous type: `signed` is missing.

@greptileai, please propose an alternative fix.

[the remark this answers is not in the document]
erik-3milabs: @RobinJadoul do you have an opinion on this? Currently, the LOAD-C9 does not even list this bit (hence, we find a bug here!), while CPU-CM51 places `signed` _after_ the `readx` bits in the signature. In the variable-description of `LOAD`, it is also listed _after_ the `readx` bits.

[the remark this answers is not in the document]
RobinJadoul: I'd stick with what's already in the CPU then, and put it last

spec/src/signatures.toml
```diff
@@ -0,0 +1,170 @@
+# cond => IS_BIT<X>
+[[signatures]]
+tag = "IS_BIT"
+kind = "template"
+input = ["BaseField"]
+cond = "Bit"
```
RobinJadoul: I believe `cond` for `IS_BIT` could also be a generic BaseField (and is described that way in the IS_BIT spec)

spec/src/signatures.toml
```diff
@@ -0,0 +1,170 @@
...
+output = "DWordWL"
+
+# LOAD[res; base_address, timestamp, read2, read4, read8]
+[[signatures]]
+tag = "LOAD"
+kind = "interaction"
+input = ["DWordWL", "DWordWL", "Bit", "Bit", "Bit", "Bit"]
+output = "DWordWL"
+
+# ECALL[timestamp, syscallnr]
```
RobinJadoul, line 100: syscallnr being Word is a mismatch with the CPU sending it in as DWordWL
I would suggest going for DWordWL so we don't implicitly take the register mod 2^32 when doing an ECALL

RobinJadoul (commented):
Should we already add a NEG signature here, in expectation of #270?

spec/src/signatures.toml
erik-3milabs: Fixed in 61c7aad

spec/src/signatures.toml
erik-3milabs: Fixed in 8744a67

spec/src/signatures.toml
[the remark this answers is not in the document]
erik-3milabs: Applied in a25be10.
Fixed the LOAD bug in #284

erik-3milabs:
@greptileai

erik-3milabs:
> Should we already add a NEG signature here, in expectation of #270?

Done in d09ebe6

spec/src/signatures.toml
```diff
@@ -0,0 +1,178 @@
...
+input = ["DWordWL", "DWordWL"]
+output = "DWordWL"
+cond = "BaseField"
+
+# cond => NEG<neg; X>
+[[signatures]]
+tag = "NEG"
+kind = "template"
+input = ["DWordHL"]
+output = "DWordDL"
```
RobinJadoul:
```suggestion
output = "DWordWL"
```

RobinJadoul (approved):

erik-3milabs:
Well, crap. should not have merged this :man_facepalming:

PR #283, opened 2026-02-04
spec: `COMMIT` chip

erik-3milabs:
In addition to introducing the `COMMIT` chip, this PR also updates the way footnotes are numbered, by wrapping numbers with square brackets. This is to prevent footnotes as being interpreted as exponents.

### DIFF
--- a/spec/book.typ
+++ b/spec/book.typ
@@ -33,6 +33,11 @@
   summary: meta.summary.map(((ch, title, _ref)) => chapter(ch, title)).join()
 )
 
+#let common-formatting(body) = {
+  set footnote(numbering: "[1]")
+  body
+}
+
 
 #let todo(background: white, foreground: black, name: none, body) = block(fill: background, outset: 0.5em, radius: 20%, stroke: black)[
   #set text(fill: foreground)
@@ -134,6 +139,7 @@
   assert(meta.summary.find(((f, _, _)) => f == file) != none, message: "Couldn't resolve typst source file " + file)
   if is-shiroa {
     (body) => {
+      show: common-formatting
       context _xref-included.update(x => x + ((file): true))
       context _toplevel.update(s => {
         if s == none {
@@ -151,6 +157,6 @@
       ])
     }
   } else {
-    (body) => body
+    common-formatting
   }
 }
--- a/spec/book.typ
+++ b/spec/book.typ
@@ -157,6 +157,6 @@
       ])
     }
   } else {
-    common-formatting
+    body => body
   }
 }
--- a/spec/ecall.typ
+++ b/spec/ecall.typ
@@ -12,6 +12,15 @@
 #let config = load_config()
 
 #show: book-page("ecall.typ")
+= About `ECALL`
+When `ECALL` is executed, it is assumed that:
+- register `A7` contains the system call number
+  #footnote(link("https://libriscv.no/docs/concepts/syscalls/#the-risc-v-system-call-abi")[The RISC-V system call ABI; libriscv.no. Accessed Feb 4, 2026.]),
+- the arguments are located in registers `A0`-`A6`, and
+- the return value is written to `A0`,
+where `A0`-`A7` are symbolic names for the registers `x10`-`x17`
+#footnote(link("https://en.wikipedia.org/wiki/RISC-V#Register_sets")[RISC-V - Register sets; en.wikipedia.org. Accessed on Feb 4, 2026.]).
+
 
 #let config = load_config()
 #let chip = load_chip("src/halt.toml", config)
@@ -55,3 +64,72 @@ This chip should only contain a single row.
 Given that $2^0 = 1$, this chip does not need to be padded.
 As such, no padding is defined.
 
+
+#let config = load_config()
+#let chip = load_chip("src/commit.toml", config)
+#let commit = raw(chip.name)
+= #commit chip
+
+== Columns
+#let nr_variables = total_nr_variables(chip)
+#let nr_columns = total_nr_instantiated_columns(chip, config)
+
+The #commit chip leverages #nr_variables variables, spanning #nr_columns columns:
+#render_chip_column_table(chip, config)
+
+== Constraints
+In this VM, committing is considered equivalent to writing a value to `stdout`.
+Hence, this chip responds to `ECALL`s with system call number 64.
+#footnote([$64$ is the system call number corresponding to `sys_write`. #link("https://github.com/riscv-collab/riscv-gnu-toolchain/blob/master/linux-headers/include/asm-generic/unistd.h#L174")[[src]]])
+Since we do not know how many bytes are to be committed, this chip employs a recursive design:
+each iteration commits one byte, and recursively "call" itself to commit the remaining bytes.
+As such, only the call from the CPU to this chip (i.e., the `first` in the recursion tree) should accept the `ECALL`; later recursive calls should not.
+This is why @commit:c:receive_ecall has multiplicity $-#`first`$.
+#render_constraint_table(chip, config, groups: "incoming")
+*Note*: the prover is free to specify the value of `first` as they see fit; @commit:c:range_first only makes sure it must be a `Bit`.
+Also, `first` being set must imply that that this is not a padding row (@commit:c:first_implies_mu).
+
+The `write` operation --- writing to a file descriptor --- has the following signature:
+#footnote([Linux man-page on `write`; man7.org, #link("https://man7.org/linux/man-pages/man2/write.2.html")[[src]]. Accessed Feb 4, 2026.])
+#[
+#show raw.where(block: true): it => block(it, fill: luma(230), inset: 1em, width: 100%, radius: 5pt)
+```c
+ssize_t write(size_t count; int fd, const void buf[count], size_t count);
+```
+]
+That is to say,
+- `A0` contains the file descriptor,
+- `A1` contains the address of `buf`'s first byte, 
+- `A2` contains `count`, and
+- the written count should be written to `A0`.
+
+Since we only support writing to `stdout` (which corresponds to $#`fd` = 1$
+#footnote([The Open Group Base Specifications, `unistd.h`; The Open Group, issue 7, #link("https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/unistd.h.html")[[src]]. Accessed Feb 4, 2026.]))
+we assert that `x10` contains $1$ in @commit:c:read_fd_write_count.
+Note that this constraint _also_ writes `count` to `A0`; 
+in this VM it is impossible for multiple for a commit to be interrupted or fail.
+Furthermore, @commit:c:read_address reads `address` from `x11` and @commit:c:read_count reads `count` from `x12`.
+Again, these memory interactions only take place when this is the `first` call in the recursion tree.
+#render_constraint_table(chip, config, groups: "read_input")
+
+Next, we read the `value` located at buffer address `address` and commit to it:
+#render_constraint_table(chip, config, groups: "commit")
+
+In parallel, we compute $#`address_incr` = #`address` + 1$ (@commit:c:address_incr) as address of the next byte to commit, and $#`count_decr` = #`count` - 1$ (@commit:c:count_decr) as the number of bytes that still has to be committed.
+@commit:c:range_address_incr and @commit:c:range_count_decr are included to satisfy @add:a:sum respectively @add:a:rhs.
+#render_constraint_table(chip, config, groups: "incr_decr")
+
+When `count_decr` (the number of bytes still to be committed) hits $0$, we should stop recursing.
+To this end, `last` is set this is the case (@commit:c:last).
+To prevent undesired lookups from occurring, `last` should only be set when we're not padding (@commit:c:last_implies_mu).
+Also, we must make sure `mu` is a bit (@commit:c:range_mu).
+#render_constraint_table(chip, config, groups: "last")
+
+Lastly, when this was not the `last` byte to commit in this recursion sequence, we recursively _Commit the Next Byte_ (`CNB`), specifying the timestamp, address to continue reading and the number of bytes that should still be committed (@commit:c:send_commit_next_byte).
+Since that certainly won't be the `first` call in the sequence, we read `address_incr` and `count_decr` from the previous recursion level into `address` and `count` and continue executing the commit.
+#render_constraint_table(chip, config, groups: "lookups")
+
+== Padding
+To pad this chip, use the below data.
+This corresponds to committing a $0$ from address $0$.
+#render_chip_padding_table(chip, config)
--- a/spec/ecall.typ
+++ b/spec/ecall.typ
@@ -123,7 +123,6 @@ This is only performed when we have not yet reached the `end` of the commit sequ
 #render_constraint_table(chip, config, groups: "commit")
 
 In parallel, we compute $#`address_incr` = #`address` + 1$ (@commit:c:address_incr) as address of the next byte to commit, and $#`count_decr` = #`count` - 1$ (@commit:c:count_decr) as the number of bytes that still has to be committed after committing this byte.
-Both of these constraints are only required when $#`μ` = 1$, to permit a simpler padding configuration.
 @commit:c:range_address_incr and @commit:c:range_count_decr are included to satisfy @add:a:sum respectively @add:a:rhs.
 #render_constraint_table(chip, config, groups: "incr_decr")
 
--- a/spec/ecall.typ
+++ b/spec/ecall.typ
@@ -152,3 +152,7 @@ These are required to ensure the multiplicities $#`first` - #`μ`$ and $#`μ` - 
 == Padding
 To pad this chip, use the below data.
 #render_chip_padding_table(chip, config)
+
+== Notes
+- The current version only supports writing to `stdout`.
+  This chip could potentially be extended to support writing to arbitrary `fd`s
--- a/spec/ecall.typ
+++ b/spec/ecall.typ
@@ -83,12 +83,10 @@ In this VM, committing is considered equivalent to writing a value to `stdout`.
 Hence, this chip responds to `ECALL`s with system call number 64.
 #footnote([RISC-V GNU-toolchain, `unistd.h`; version 2026-01-23, #link("https://github.com/riscv-collab/riscv-gnu-toolchain/blob/2026.01.23/linux-headers/include/asm-generic/unistd.h#L174")[[src]]])
 Since we do not know how many bytes are to be committed, this chip employs a recursive design:
-each iteration commits one byte, and recursively "call" itself to commit the remaining bytes.
+each iteration commits one byte, and recursively "calls" itself to commit the remaining bytes.
 As such, only the call from the CPU to this chip (i.e., the `first` in the recursion tree) should accept the `ECALL`; later recursive calls should not.
 This is why @commit:c:receive_ecall has multiplicity $-#`first`$.
 #render_constraint_table(chip, config, groups: "incoming")
-*Note*: the prover is free to specify the value of `first` as they see fit; @commit:c:range_first only makes sure it must be a `Bit`.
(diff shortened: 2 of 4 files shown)


erik-3milabs:
Currently, it is possible to have a 2^64-length recursion by starting an `ECALL[64]` with `count = 0`.
We could prevent this from happening, but doing so would probably add a column. Is this a problem we really want to prevent?

spec/ecall.typ
```diff
@@ -55,3 +64,72 @@ This chip should only contain a single row.
...
+#render_constraint_table(chip, config, groups: "incoming")
+*Note*: the prover is free to specify the value of `first` as they see fit; @commit:c:range_first only makes sure it must be a `Bit`.
+Also, `first` being set must imply that that this is not a padding row (@commit:c:first_implies_mu).
+
+The `write` operation --- writing to a file descriptor --- has the following signature:
+#footnote([Linux man-page on `write`; man7.org, #link("https://man7.org/linux/man-pages/man2/write.2.html")[[src]]. Accessed Feb 4, 2026.])
+#[
+#show raw.where(block: true): it => block(it, fill: luma(230), inset: 1em, width: 100%, radius: 5pt)
+```c
+ssize_t write(size_t count; int fd, const void buf[count], size_t count);
```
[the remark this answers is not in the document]
erik-3milabs: This comment is incorrect. Here, count is both an input parameter (indicating the number of bytes that _should_ be written) and an output parameter (indicating how many bytes were _actually_ written).

spec/ecall.typ
```diff
@@ -55,3 +64,72 @@ This chip should only contain a single row.
...
+- `A0` contains the file descriptor,
+- `A1` contains the address of `buf`'s first byte, 
+- `A2` contains `count`, and
+- the written count should be written to `A0`.
+
+Since we only support writing to `stdout` (which corresponds to $#`fd` = 1$
+#footnote([The Open Group Base Specifications, `unistd.h`; The Open Group, issue 7, #link("https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/unistd.h.html")[[src]]. Accessed Feb 4, 2026.]))
+we assert that `x10` contains $1$ in @commit:c:read_fd_write_count.
+Note that this constraint _also_ writes `count` to `A0`; 
+in this VM it is impossible for multiple for a commit to be interrupted or fail.
```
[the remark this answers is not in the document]
erik-3milabs: Fixed in 0dea13c

spec/ecall.typ
```diff
@@ -55,3 +64,72 @@ This chip should only contain a single row.
...
+
+Next, we read the `value` located at buffer address `address` and commit to it:
+#render_constraint_table(chip, config, groups: "commit")
+
+In parallel, we compute $#`address_incr` = #`address` + 1$ (@commit:c:address_incr) as address of the next byte to commit, and $#`count_decr` = #`count` - 1$ (@commit:c:count_decr) as the number of bytes that still has to be committed.
+@commit:c:range_address_incr and @commit:c:range_count_decr are included to satisfy @add:a:sum respectively @add:a:rhs.
+#render_constraint_table(chip, config, groups: "incr_decr")
+
+When `count_decr` (the number of bytes still to be committed) hits $0$, we should stop recursing.
+To this end, `last` is set this is the case (@commit:c:last).
```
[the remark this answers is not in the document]
erik-3milabs: Fixed in 0dea13c

spec/book.typ
```diff
@@ -151,6 +155,6 @@
       ])
     }
   } else {
-    (body) => body
+    common-formatting
```
RobinJadoul: Hmm, upon further consideration, maybe this should actually be in ebook.typ (just the call to apply common-formatting). wdyt?

spec/ecall.typ
```diff
@@ -12,6 +12,15 @@
 #let config = load_config()
 
 #show: book-page("ecall.typ")
+= About `ECALL`
+When `ECALL` is executed, it is assumed that:
+- register `A7` contains the system call number
+  #footnote(link("https://libriscv.no/docs/concepts/syscalls/#the-risc-v-system-call-abi")[The RISC-V system call ABI; libriscv.no. Accessed Feb 4, 2026.]),
```
RobinJadoul: The alternative to the "accessed" annotation could be to make them web archive links.

spec/ecall.typ
```diff
@@ -55,3 +64,72 @@ This chip should only contain a single row.
...
+ssize_t write(size_t count; int fd, const void buf[count], size_t count);
+```
+]
+That is to say,
+- `A0` contains the file descriptor,
+- `A1` contains the address of `buf`'s first byte, 
+- `A2` contains `count`, and
+- the written count should be written to `A0`.
+
+Since we only support writing to `stdout` (which corresponds to $#`fd` = 1$
```
RobinJadoul, line 106: Another option would be to support arbitrary fds, and embed them in the commitments as kind of "domain separation"; but that's probably not too useful in practice.
Maybe something for the future, should we have a more versatile way to deal with this.

spec/src/commit.toml
```diff
@@ -0,0 +1,210 @@
...
+poly = ["*", "first", ["not", "μ"]]
+ref = "commit:c:first_implies_mu"
+
+[[constraint_groups]]
+name = "read_input"
+
+[[constraints.read_input]]
+kind = "interaction"
+tag = "MEMW"
+input = [1, 10, "count", "timestamp", 1, 0, 0]
```
RobinJadoul: Register index scaling is off again.
We should find a more fool-proof way of specifying those.

RobinJadoul: We also need a better way to type and cast the inputs to memw, being an array.

RobinJadoul (changes requested):
> Currently, it is possible to have a 2^64-length recursion by starting an ECALL[64] with count = 0.
> We could prevent this from happening, but doing so would probably add a column. Is this a problem we really want to prevent?

I _think_ a write with `count = 0` is a valid thing to do, and this would render it unprovable.
From `man 2 write`: 
> If  count is zero and fd refers to a regular file, then write() may return a failure status if one of the errors below is detected.  If no errors are detected, or error detection is not performed, 0 will be  returned  without causing  any  other effect.  If count is zero and fd refers to a file other than a regular file, the results are not specified.

What about turning last into `end`, representing `count = 0`, where we enforce `μ = 0 => end = 1` and performing both CNB and COMMIT with multiplicity `1 - end`?

spec/src/commit.toml
erik-3milabs: > Register index scaling is off again.

Ahh, the 2x. 

> We should find a more fool-proof way of specifying those.

I agree. As long as we are developing this spec, things might go alright, but if at any point this is transferred to somebody else, this is going to give problems.

spec/ecall.typ
erik-3milabs: I'll list it as a possible extension to consider later.

spec/book.typ
erik-3milabs: Why? The footnote-exponent-mixup should be fixed for both the PDF and the web pages, right?

spec/ecall.typ
erik-3milabs: Yeah, that'd be even better.

spec/book.typ
RobinJadoul: Just the line that applies it when not is-shiroa
Keep that identify function as-is

RobinJadoul: See 2d5513c for what I meant

spec/src/commit.toml
RobinJadoul: The casting inputs into memw is getting handled by #295 and we'll get to actually doing it here too when we get to checking signatures in the tooling.
For the registers, my best idea so far would be to introduce it as a `REG` template, but it doesn't strictly prevent calling into MEMW in another way (maybe not needed if the rest of the register accesses are all clearly going through REG)

spec/src/commit.toml
```diff
@@ -0,0 +1,210 @@
+name = "COMMIT"
+
+# Variables
+
+[[variables.input]]
+name = "index"
+type = "RowIndex"
+desc = "index of value being committed"
+precomputed = 1
```
RobinJadoul: I just realized that this will also behave weirdly with rows for count = 0: you end up skipping indices based on that, so that may break my suggestion to deal with those rows without extra column :/

erik-3milabs:
```suggestion
precomputed = true
```

erik-3milabs: I'd go even further and say that if we want to support `count=0` calls and make sure they do NOT use up an `index`, then we'd have to introduce some sort of intermediary chip (i.e., the kernel chip we've been discussing) that already reads `x10` and simply does not engage the `COMMIT` chip when `count=0`.

erik-3milabs: The above statement hinges on the assumption that `index` must be precomputed. When we let go of that assumption, the problem can be bypassed by adding a couple extra columns.

spec/src/commit.toml
erik-3milabs: Perhaps we should also add a `RAM` template. While overkill, this would allow making the complicated `MEMW`-interface private, completely shielding everybody from mistakes. Moreover, having the `RAM` and `REG` template names will make it much easier to discern whether a read/write is to a register or RAM (instead of distinguishing a 0/1)

RobinJadoul: Potentially, we even want to split those into READ and WRITE variants?

RobinJadoul: (though sometimes we do a combo read and write, e.g. for the pc)

RobinJadoul: Created #313 for this

spec/ecall.typ
```diff
@@ -45,13 +54,105 @@ This prevents any other operation involving memory from being executed hereafter
...
+- `A2` contains `count`, and
+- the written count should be written to `A0`.
+
+@commit:c:read_address reads `address` from `x11` (=`A1`) and @commit:c:read_count reads `count` from `x12` (=`A2`).
+Since we only support writing to `stdout` (which corresponds to $#`fd` = 1$
+#footnote([The Open Group Standard for Information Technology --- Portable Operating System Interface (POSIX) Base Specifications, `unistd.h`; The Open Group, issue 8, #link("https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/unistd.h.html")[[src]]]))
+we assert that `x10` contains $1$ in @commit:c:read_fd_write_count.
+Note that this constraint _also_ writes `count` to `A0`; 
+in this VM it is impossible for a commit to be interrupted or fail.
+Lastly, the `index` is read from `x254`#footnote([In this VM, register 254 is reserved for containing the commitment index.]); in the same operation, $#`index` + #`count`$ is written back to this location by @commit:c:read_index.
```
RobinJadoul, line 111: We should maybe have a separate list of register meanings somewhere. Maybe in memory.typ, or actually, the future `REG` template?

spec/src/commit.toml
```diff
@@ -0,0 +1,223 @@
...
+
+[[constraint_groups]]
+name = "incr_decr"
+
+[[constraints.incr_decr]]
+kind = "template"
+tag = "ADD"
+input = ["address", ["cast", 1, "DWordWL"]]
+output = ["cast", "address_incr", "DWordWL"]
+cond = "μ"
```
RobinJadoul: I suppose this cond isn't really needed? (other than for padding?) It essentially only ups the degree of the constraint

spec/src/commit.toml
```diff
@@ -0,0 +1,223 @@
...
+iter = ["i", 0, 3]
+multiplicity = "μ"
+ref = "commit:c:range_address_incr"
+
+[[constraints.incr_decr]]
+kind = "template"
+tag = "SUB"
+input = ["count", ["cast", 1, "DWordWL"]]
+output = ["cast", "count_decr", "DWordWL"]
+cond = "μ"
```
RobinJadoul: Same as for the ADD

spec/ecall.typ
```diff
@@ -45,13 +54,105 @@ This prevents any other operation involving memory from being executed hereafter
...
+Next, we read the `value` located at buffer address `address` and commit to it under the given `index`.
+This is only performed when we have not yet reached the `end` of the commit sequence.
+#render_constraint_table(chip, config, groups: "commit")
+
+In parallel, we compute $#`address_incr` = #`address` + 1$ (@commit:c:address_incr) as address of the next byte to commit, and $#`count_decr` = #`count` - 1$ (@commit:c:count_decr) as the number of bytes that still has to be committed after committing this byte.
+Both of these constraints are only required when $#`μ` = 1$, to permit a simpler padding configuration.
+@commit:c:range_address_incr and @commit:c:range_count_decr are included to satisfy @add:a:sum respectively @add:a:rhs.
+#render_constraint_table(chip, config, groups: "incr_decr")
+
+When `count` hits $0$, we should performing further recursive calls.
```
RobinJadoul:
```suggestion
When `count` hits $0$, we should stop performing further recursive calls.
```

spec/ecall.typ
```diff
@@ -45,13 +54,105 @@ This prevents any other operation involving memory from being executed hereafter
...
+We use the `end` bit to indicate these circumstances.
+
+#render_constraint_table(chip, config, groups: "end")
+
+*Note*: 
++ Rather than setting $#`end` = 1$ when $#`count` = 0$, we do so $#`count_decr` = -1$.
```
RobinJadoul:
```suggestion
+ Rather than setting $#`end` = 1$ when $#`count` = 0$, we do so when $#`count_decr` = -1$.
```

spec/src/commit.toml
```diff
@@ -0,0 +1,223 @@
...
+tag = "CNB"
+input = ["timestamp", ["+", "index", 1], ["cast", "address_incr", "DWordWL"], "count_decr"]
+multiplicity = ["-", "μ", "end"]
+ref = "commit:c:send_commit_next_byte"
+
+[[constraints.lookups]]
+kind = "interaction"
+tag = "CNB"
+input = ["timestamp", "index", "address", "count"]
+multiplicity = ["-", "first", "μ"]
```
RobinJadoul: I think `-(μ - first)` may be more readable

spec/src/commit.toml
```diff
@@ -0,0 +1,223 @@
...
+pad = 0
+
+[[variables.multiplicity]]
+name = "μ"
+type = "Bit"
+desc = ""
+pad = 0
+
+# Assumptions
+
```
RobinJadoul, line 66: We're assuming (kinda) that `address` and `count` are WL-checked

spec/ecall.typ
```diff
@@ -45,13 +54,105 @@ This prevents any other operation involving memory from being executed hereafter
...
+*Note*: the observant reader will notice that @commit:c:read_index casts `count` to a `BaseField`, potentiallly losing information.
+This is indeed correct.
+However, since it is practically impossible to commit more than $2^64-2^32$ bytes in a single VM execution, it was decided to permit this.
+
+Next, we read the `value` located at buffer address `address` and commit to it under the given `index`.
+This is only performed when we have not yet reached the `end` of the commit sequence.
+#render_constraint_table(chip, config, groups: "commit")
+
+In parallel, we compute $#`address_incr` = #`address` + 1$ (@commit:c:address_incr) as address of the next byte to commit, and $#`count_decr` = #`count` - 1$ (@commit:c:count_decr) as the number of bytes that still has to be committed after committing this byte.
+Both of these constraints are only required when $#`μ` = 1$, to permit a simpler padding configuration.
```
RobinJadoul: Ah, I see now... 
It _is_ the only degree three constraint, so we could get rid of it and make the padding slightly "worse"

RobinJadoul (commented):
Nice, seems like this should work.
I'm thinking that we may not really need `μ`, as first and end may be enough.
You get a weird thing about the ZERO for end needing a multiplicity depending on end, but it could still work out due to interaction balancing.

spec/src/commit.toml
erik-3milabs: I think that our core assumptions are technically
1. registers store proper WLs, and
2. no chip outside `COMMIT` executes `CNB`.

spec/ecall.typ
erik-3milabs: Yup, that `REG` template page sounds like a great location :ok_hand:

erik-3milabs:
@greptileai

spec/src/commit.toml
RobinJadoul: And the call to `ECALL` is a proper WL

erik-3milabs: the `syscallnr`, you mean?

RobinJadoul: No, just a monetarily forgetting about arguments no longer being passed through there at all...

RobinJadoul: No, just a monetarily forgetting about arguments no longer being passed through there at all...

spec/src/commit.toml
```diff
@@ -0,0 +1,221 @@
...
+[[constraint_groups]]
+name = "end"
+
+[[constraints.end]]
+kind = "interaction"
+tag = "ZERO"
+input = [["+", ["-", 0xFFFF, ["idx", "count_decr", 0]], ["-", 0xFFFF, ["idx", "count_decr", 1]], ["-", 0xFFFF, ["idx", "count_decr", 2]], ["-", 0xFFFF, ["idx", "count_decr", 3]]]]
+output = "end"
+multiplicity = "μ"
+ref = "commit:c:end"
```
RobinJadoul, line 177: We could save an interaction by turning this into `!end => count_decr = -1`; which would require a full `2^64` extra commits to reach the next possibility to end and balance the logup sum otherwise.
That would even be impossible to prove due to addresses being read twice at the same timestamp.
If we replace `count` by a BaseField, this argument would weaken to only `p` extra "iterations" being impossible to prove.

spec/src/commit.toml
```diff
@@ -0,0 +1,221 @@
...
+name = "count"
+type = "DWordWL"
+desc = "number of bytes to commit"
+pad = ["arr", 1, 0, 0, 0]
+
+[[variables.auxiliary]]
+name = "count_decr"
+type = "DWordHL"
+desc = "$#`count` - 1$"
+pad = ["arr", 0, 0, 0, 0]
```
RobinJadoul, line 39: We could get rid of this at the cost of one extra column for count as basefield that is populated conditionally from `CNB` or recomposed from the DWordWL count that came from the register.

RobinJadoul (approved):
A couple ideas for potential optimizations, but maybe we should just go for "works" for now, and work towards something better with the kernel.
I'd still write down our assumptions to be explicit, but otherwise this should be fine.

I'm starting to think now of "overcommitting" and having the ECALL return some "out of space" error, but let's maybe keep this for when we do a kernel-centric redesign.
For now, we don't need _full_ genericity.

spec/src/commit.toml
erik-3milabs: How would that work? What about committing a sequence of 8 bytes. The first seven bytes will have `end = 0`, but we definitely don't want  `count_decr = -1` in those cases.

Perhaps you're misinterpreting what `count_decr` means? Note that `count_decr := count - 1`; it is not the value by which we decrease `count`, i.e., `next_count := count - count_decr` or smth.

spec/src/commit.toml
erik-3milabs: am I correct in understanding that this proposal hinges on the assumption that `count < p`?

spec/src/commit.toml
RobinJadoul: I may have flipped the condition. The idea was that one direction of the ZERO implication didn't matter as much because it makes the overall system unprovable due to size

spec/src/commit.toml
RobinJadoul: Hmm, yeah, that would also have to add an LT enforced with multiplicity first, I suppose

spec/src/commit.toml
erik-3milabs: I think we would need `end => count = 0`; one cannot set the flag when `count != 0`, and if you "forget" when it is 0, you have to another round of 2^64 commits. Great observation!

PR #285, opened 2026-02-04
Add trace segmentation support for large executions

gabrielbosio:
Adds infrastructure for splitting execution traces into segments when trace length exceeds limits.

## Changes

- `SegmentBoundary` struct for state continuity across segments (timestamps, memory, registers)
- `SegmentConfig` for segment-specific behavior (is_final flag)
- `Traces::from_logs_segmented()` for segmented trace generation
- `generate_dummy_halt_trace()` for non-final segments
- Validation to reject ECALL in intermediate segments (fails early instead of bus imbalance later)

## Memory model

- Traces (~200MB) can be freed after each segment is proved
- Only `SegmentBoundary` (KB) carried forward between segments
- Rust ownership handles this automatically

## Deferred

- Proving logic (each segment would re-prove static tables like BITWISE)
- Proof aggregation / cross-segment LogUp balance

### DIFF
--- a/prover/src/lib.rs
+++ b/prover/src/lib.rs
@@ -49,8 +49,6 @@ pub enum Error {
     MissingInstruction(u64),
     /// Program does not contain an ECALL (halt) instruction
     MissingHaltEcall,
-    /// ECALL found in intermediate segment (should only be in final segment)
-    UnexpectedEcallInSegment,
     /// Executor failed (setup or runtime error)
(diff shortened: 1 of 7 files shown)


gabrielbosio:
Pausing development of this PR since, to make it work as intended, we have to consider things like how to reference a table `n_0 < n` in table segment `n`.

PR #289, opened 2026-02-05
feat(math/gpu): add Metal GPU FFT with Bowers algorithm

diegokingston:
Implement Metal-accelerated FFT for the Goldilocks field on macOS, using the Bowers G network algorithm with 2-layer fusion.

Key features:
- Metal shaders for Goldilocks field arithmetic (add, sub, mul, reduce128)
- Bowers FFT with 2-layer butterfly fusion for reduced memory traffic
- LayerTwiddles computation on GPU for cache-friendly access
- Batch FFT for SoA layouts (multiple polynomials)
- In-place bit-reversal permutation

Performance characteristics:
- GPU parallel execution of butterfly operations
- Shared memory optimization via threadgroup buffers
- Adaptive thread dispatch based on pipeline limits

Requires macOS with Metal-capable GPU. Enable with --features "metal".


jotabulacios (commented):
sorry, made a comment by mistake

jotabulacios (commented):
-

jotabulacios:
@greptileai please review

jotabulacios:
@claude please review

jotabulacios:
@claude

jotabulacios:
@greptile

jotabulacios:
@greptile

jotabulacios:
@claude please review

jotabulacios:
@greptileai please review

Dead error variant (CommandQueueCreation): device.new_command_queue() in the metal crate (v0.29) returns
CommandQueue directly, not Option or Result — there's no way to detect failure. We can remove the unused variant in a follow-up cleanup.

Allocation errors ignored:
Same situation — new_buffer in metal v0.29 returns Buffer directly and panics internally on OOM. There's no fallible API to check against. We can remove BufferAllocation along with CommandQueueCreation in a cleanup PR.

CPU oracle panics: Good catch, added assert!(input.len().is_power_of_two()) at the start of cpu_fft.

Doc/behavior mismatch:
This is correct as-is. fft() produces bit-reversed output (Bowers network), then fft_natural_order() applies one bit-reversal to get natural order. There's no double bit-reversal. The 19 differential tests against the CPU FFT confirm the ordering is correct.

Silent GPU errors: metal crate v0.29 doesn't expose command_buffer.error() on CommandBufferRef. status() is the only inspection available. We'd need to upgrade the crate or use raw Objective-C bindings to surface the underlying error.

jotabulacios:
@greptile

jotabulacios:
@claude please review

jotabulacios:
@greptileai 

Root of unity
This is correct. The loop for _ in order..32 iterates exactly 32 - order times, which is identical to the CPU implementation in [traits.rs (lines 92-93)](app://-/index.html#): (0..TWO_ADICITY - order).fold(root, |acc, _| acc.square()). Both start from the same 2^32-th primitive root (1753635133440165772) and square it 32 - order times, yielding a primitive 2^order-th root. The unit test at line 716 confirms: root^(2^order) = 1 for order=10.

FFT ordering mismatch
No double bit-reversal. fft() produces bit-reversed output (Bowers G-network), then fft_natural_order() applies a single bitrev_permutation_inplace() to get natural order. This is one reversal, not two. The 19 differential tests against the CPU FFT confirm the ordering is correct.

Twiddle sizing OOB
The sizing is correct. For layer l, the buffer has n >> (l+1) elements. The fused kernel sets quarter = block_size >> 2 = n >> (l+2) and j < quarter.

Max index into twiddles_l0 is j + quarter ≤ 2*quarter - 1 = (n >> (l+1)) - 1, which is exactly the last valid index.
Max index into twiddles_l1 is j ≤ quarter - 1 = (n >> (l+2)) - 1, also exactly the last valid index.
No OOB is possible.

 MAX_FFT_ORDER docs
The docs are consistent. Line 8: “Maximum FFT size is 2^31 elements”. Line 20–21: MAX_FFT_ORDER = 31 with comment “not 32, because 2^32 overflows u32 in shader parameters”. The test at line 736 verifies that order 32 is rejected. All three say the same thing.

 Ungated test
The entire metal module is behind #[cfg(feature = "metal")] in [mod.rs (lines 4-5)](app://-/index.html#). The test file is further gated with #[cfg(all(test, feature = "alloc"))] in [mod.rs (line 30)](app://-/index.html#). If the metal feature is not enabled, none of this code compiles. CI won’t break.

32-bit shift
Already addressed. The code uses let n: u64 = 1u64 << 32, not usize. This is valid on all platforms.

jotabulacios:
@greptileai what's the new confidence score?

jotabulacios:
@greptileai please review

PR #292, opened 2026-02-05
Fix big-endian serialization of extension field types

gabrielbosio:
`to_bytes_be` and `from_bytes_be` for multi-component field extensions kept the same component order as `to_bytes_le`, only reversing bytes within each component. Now they also reverse component order so that `to_bytes_be` is the true byte-reversal of `to_bytes_le`.

Fixed types: Goldilocks Fp3, BabyBear Fp4, BabyBear U32 Fp4, BLS12-381 Fp2, BN254 Fp2. Also removed `x_rev` workarounds in BLS12-381/BN254 G2 point compression that were compensating for this bug.


gabrielbosio:
superseded by #358

PR #295, opened 2026-02-06
spec: Introduce array expressions

### LINKED ISSUE #135: spec: Padding specifications for array types
Building upon #133, we should be able to specify the padding value for array-typed columns through iters or individually.
This is analogous to rendering of virtual column definitions, but would likely be simpler.

RobinJadoul:
Closes #135 

To be merged after #271

### DIFF
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -52,9 +52,9 @@
 #let render_chip_padding_table(chip, config) = {
   // Whether `var` is a preprocessed variable.
   let is_preprocessed(var) = {
-    config.variables.types
+    let type = config.variables.types
     .filter(t => t.label == var.type)
-    .all(t => t.at("preprocessed", default: false))
+    type.len() > 0 and type.all(t => t.at("preprocessed", default: false))
   }
 
   let instantiated_vars = config.variables.categories.instantiated.map(c => chip.variables.at(c, default: ())).flatten()
--- a/spec/tooling/chip.py
+++ b/spec/tooling/chip.py
@@ -62,6 +62,7 @@ DEFAULT_TYPE: Type = Range.const(0)
 type Expr = (
     LitExpr
     | VarExpr
+    | ArrExpr
     | IdxExpr
     | CastExpr
     | MulExpr
@@ -303,6 +304,8 @@ def build_expr(config: Optional["Config"], data: object) -> Expr:
                 x.isidentifier(), f"Invalid identifier name for variable {x!r}"
             )
             return VarExpr(x)
+        case ["arr", *elems]:
+            return ArrExpr([build_expr(config, e) for e in elems])
         case ["idx", x, y]:
             return IdxExpr(build_expr(config, x), build_expr(config, y))
         case ["cast", x, t]:
--- a/spec/expr.typ
+++ b/spec/expr.typ
@@ -28,6 +28,7 @@
 // <expr> ::= ()                           ; ""
 //          | var                          ; str(var)
 //          | int                          ; int
+//          | ["arr", expr, ...]           ; [expr, ...]
 //          | ["idx", expr1, expr2]        ; expr1[expr2]
 //          | ["not", expr]                ; !expr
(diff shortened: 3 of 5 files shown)


RobinJadoul:
Everything compiles fine in typst, and mwrap isn't needed since arrays come with their own non-paren delimiters.
@greptileai

spec/tooling/chip.py
```diff
@@ -98,6 +99,14 @@ def typecheck(self, env: Environment) -> Type:
         return DEFAULT_TYPE
 
 
+@dataclass
+class ArrExpr:
+    elems: list[Expr]
+
+    def typecheck(self, env: Environment) -> Type:
+        return [e.typecheck(env) for e in self.elems]
```
erik-3milabs, line 115: do we not wish to enforce that `len(elems) > 0`? (though perhaps already on `__init__`)

spec/tooling/chip.py
```diff
@@ -821,6 +836,7 @@ def __init__(self, config: Config, data: dict):
             self.output = build_expr(config, data["output"])
         else:
             self.output = None
+        assert "multiplicity" in data, data
```
erik-3milabs: 1. why is this in this PR? (and not in PR271)
2. should the error message not be a bit more explicit?
3. why is this a regular assert (and not an `reporter.asserts`)?

spec/tooling/chip.py
```diff
@@ -928,12 +944,13 @@ def typecheck(self) -> Iterable[TemplateSignature | InteractionSignature]:
...
     config = Config.from_file(sys.argv[1])
+    signatures = sys.argv[2] # Later
     if reporter.reported:
         sys.exit(1)
     reported = False
     chips: list[Chip] = []
-    for file in sys.argv[2:]:
-        if file == sys.argv[1]:
+    for file in sys.argv[3:]:
+        if file in sys.argv[1:3]:
```
erik-3milabs: while OK, these changes do not belong in this PR

spec/chip.typ
```diff
@@ -52,9 +52,9 @@
 #let render_chip_padding_table(chip, config) = {
   // Whether `var` is a preprocessed variable.
   let is_preprocessed(var) = {
-    config.variables.types
+    let type = config.variables.types
     .filter(t => t.label == var.type)
-    .all(t => t.at("preprocessed", default: false))
+    type.len() > 0 and type.all(t => t.at("preprocessed", default: false))
   }
```
erik-3milabs, line 58: this should probably be moved to 271 as well

spec/tooling/chip.py
```diff
@@ -402,10 +413,14 @@ def __init__(self, default_name: str, lookup: Callable[[str], Type], data: dict)
...
-            if not isinstance(start, int) and not (isinstance(start, str) and start.isdigit()):
+            if not isinstance(start, int) and not (
+                isinstance(start, str) and start.isdigit()
+            ):
                 reporter.error(f"Range start not an int: {data!r}")
                 start = 0
-            if not isinstance(stop, int) and not (isinstance(stop, str) and stop.isdigit()):
+            if not isinstance(stop, int) and not (
+                isinstance(stop, str) and stop.isdigit()
+            ):
```
erik-3milabs, line 429: This should probably also be part of 271

erik-3milabs (approved):
I would move some of the code (e.g. 7ef1970) to PR271, that is all about type checking, to keep this PR clean & focused on just arrays.
Aside from that, LGTM

RobinJadoul:
I must have messed up with the update-refs, because that was absolutely meant to move onto the typechecking only

RobinJadoul:
Hmm, but https://github.com/yetanotherco/lambda_vm/commit/7ef197062860e57211d8c3fa81607437ffff948c definitely appears as commit on typecheck, so I can only guess it's a problem in github's UI

spec/tooling/chip.py
erik-3milabs: suddenly, this code was moved to #271?

PR #297, opened 2026-02-06
spec: rv2 as DWordBL in CPU due to STORE

RobinJadoul:
This is a (temporary, for now) change to the CPU to have `rv2` available as `DWordBL` so that the `MEMW` interaction on the `STORE` path can work correctly.

It already builds on top of the array expressions to clean up the signatures for all MEMW calls from the CPU along the way.

### DIFF
--- a/spec/src/cpu.toml
+++ b/spec/src/cpu.toml
@@ -533,6 +533,13 @@ multiplicity = 1
 [[constraints.range]]
 kind = "interaction"
 tag = "IS_BYTE"
+input = [["idx", "rv2", "i"]]
+iter = ["i", 0, 7]
+multiplicity = 1
+
+[[constraints.range]]
+kind = "interaction"
+tag = "IS_BYTE"
 input = [["idx", "arg1", "i"]]
 iter = ["i", 0, 7]
 multiplicity = 1
--- a/spec/src/cpu.toml
+++ b/spec/src/cpu.toml
@@ -231,7 +231,7 @@ pad = 0
 
 [[variables.auxiliary]]
 name = "rv2"
-type = "DWordWHH"
+type = "DWordBL"
 desc = "The value of register `rs2`"
 pad = 0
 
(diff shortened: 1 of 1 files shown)


spec/src/cpu.toml
```diff
@@ -643,12 +650,11 @@ multiplicity = "DIVREM"
...
 prefix = "M"
 
-# TODO: no types available, so no casting yet
 [[constraints.mem]]
 kind = "interaction"
 tag = "MEMW"
-input = [1, ["*", 2, "rs1"], "rv1", ["+", "timestamp", ["cast", 0, "DWordWL"]], 1, 0, 0]
-output = "rv1"
+input = [1, ["*", 2, "rs1"], ["arr", ["idx", ["cast", "rv1", "DWordWL"], 0], ["idx", "rv1", 2], 0, 0, 0, 0, 0, 0], ["+", "timestamp", ["cast", 0, "DWordWL"]], 1, 0, 0]
+output = ["arr", ["idx", ["cast", "rv1", "DWordWL"], 0], ["idx", "rv1", 2], 0, 0, 0, 0, 0, 0]
```
erik-3milabs, line 657: While I agree that pushing `rv1` into the array this way is correct, I'd be inclined to apply `cast` to the second appearance of "rv1" in either array as well, just to make it easier to read.

spec/src/cpu.toml
```diff
@@ -747,20 +749,20 @@ poly = ["-", ["idx", ["cast", "arg1", "DWordWL"], 1], ["*", ["not", "word_instr"
...
 poly = ["-", ["idx", ["cast", "arg2", "DWordWL"], 0], ["*", ["-", 1, "STORE", "LOAD"], ["idx", ["cast", "rv2", "DWordWL"], 0]], ["*", ["-", 1, "BEQ", "BLT"], ["idx", "imm", 0]]]
 
 [[constraints.ext]]
 kind = "arith"
-constraint = "$#`arg2[4:]` = (1 - #`STORE` - #`LOAD`) dot ((1 - #`word_instr`) dot #`rv2[2]` + #`signed` dot #`arg2_sign_bit` dot (2^(32) - 1)) + (1 - #`BEQ` - #`BLT`) dot #`imm[1]`$"
-poly = ["-", ["idx", ["cast", "arg2", "DWordWL"], 1], ["*", ["-", 1, "STORE", "LOAD"], ["not", "word_instr"], ["idx", "rv2", 2]], ["*", ["-", 1, "STORE", "LOAD"], "signed", "arg2_sign_bit", ["-", ["^", 2, 32], 1]], ["*", ["-", 1, "BEQ", "BLT"], ["idx", "imm", 1]]]
+constraint = "$#`arg2[4:]` = (1 - #`STORE` - #`LOAD`) dot ((1 - #`word_instr`) dot #`rv2[4:]` + #`signed` dot #`arg2_sign_bit` dot (2^(32) - 1)) + (1 - #`BEQ` - #`BLT`) dot #`imm[1]`$"
+poly = ["-", ["idx", ["cast", "arg2", "DWordWL"], 1], ["*", ["-", 1, "STORE", "LOAD"], ["not", "word_instr"], ["idx", ["cast", "rv2", "DWordWL"], 1]], ["*", ["-", 1, "STORE", "LOAD"], "signed", "arg2_sign_bit", ["-", ["^", 2, 32], 1]], ["*", ["-", 1, "BEQ", "BLT"], ["idx", "imm", 1]]]
```
erik-3milabs, line 765: I believe we decided to move away from slicing. Shall we use this opportunity to remove the slices in these two constraints?

erik-3milabs (commented):
there are some changes in here that are not expected given the title of the PR...

spec/src/cpu.toml
RobinJadoul: I mostly mind them in the "formal" polys, less so in the descriptive constraint

erik-3milabs:
I take it this can be closed, as it is superseded by #308, right?

PR #296, opened 2026-02-06
Add hyperfine benchmarks for CLI prove command

gabrielbosio:
Prove benchmarking infrastructure mirroring the existing execute benchmarks. Local script (`scripts/bench_prove.sh`), CI workflow triggered by `bench-prove` label or manual dispatch, and `bench-prove` Makefile target. Only comments on the PR when a regression >5% is detected.

PR #299, opened 2026-02-06
Process tables sequentially to reduce peak memory

gabrielbosio:
Instead of collecting all LDE evaluations upfront across all tables, process each table sequentially: recompute LDE evaluations from polynomials on demand and free them after the table's proof is generated. Verifier ordering updated to match.


gabrielbosio:
we should merge #296 to see hyperfine's results

gabrielbosio:
**Method:** Poll `/proc/PID/status` VmHWM (kernel-tracked peak RSS) every 5 seconds while the CLI prover runs.

**Machine:** `admin@vm-tests-avx512` — Debian 12, x86_64, 62 GB RAM, 16 cores, AVX-512.

Results (2^19 instructions, bench_512k):
- `main`: 50.4 GB peak RSS, ~6 min
- `seq-proving-memory-opt`: 40.0 GB peak RSS, ~5.4 min

`seq-proving-memory-opt` saves ~10.4 GB (20% reduction). 9 AIR tables, 247 total trace columns.

MauroToscano:
With bigger programs there is an improvement:

image

MauroToscano (changes requested):
This extra computation needs to be further measured, it seems like with big programs it will get a big penalty.

gabrielbosio:
superseded by #349

PR #312, opened 2026-02-10
spec: Typecheck signatures and make all chips pass

RobinJadoul:


spec/src/halt.toml
```diff
@@ -17,31 +17,31 @@ name = "all"
 [[constraints.all]]
 kind = "interaction"
 tag = "MEMW"
-input = [1, ["*", 2, "i"], 0, ["-", ["^", 2, 64], 1], 1, 0, 0]
+input = [1, ["cast", ["*", 2, "i"], "DWordWL"], ["cast", 0, ["Byte", 8]], ["cast", ["-", ["^", 2, 64], 1], "DWordWL"], 1, 0, 0]
```
erik-3milabs: why do we cast to `["Byte", 8]` here? this is writing a 0 to a register, so I'd be expecting `["Word",8]` or smth.

RobinJadoul: Technically, it'd probably be even better to cast to ["BaseField", 8] even, but same difference, it's all zero
```suggestion
input = [1, ["cast", ["*", 2, "i"], "DWordWL"], ["cast", 0, ["BaseField", 8]], ["cast", ["-", ["^", 2, 64], 1], "DWordWL"], 1, 0, 0]
```

spec/src/halt.toml
```diff
@@ -17,31 +17,31 @@ name = "all"
...
 multiplicity = 1
 ref = "halt:c:zeroize_registers_lo"
 
 [[constraints.all]]
 kind = "interaction"
 tag = "MEMW"
-input = [1, ["*", 2, 10], 0, ["-", ["^", 2, 64], 1], 1, 0, 0]
-output = 0
+input = [1, ["cast", ["*", 2, 10], "DWordWL"], ["cast", 0, ["Byte", 8]], ["cast", ["-", ["^", 2, 64], 1], "DWordWL"], 1, 0, 0]
+output = ["cast", 0, ["Byte", 8]]
```
RobinJadoul:
```suggestion
input = [1, ["cast", ["*", 2, 10], "DWordWL"], ["cast", 0, ["BaseField", 8]], ["cast", ["-", ["^", 2, 64], 1], "DWordWL"], 1, 0, 0]
output = ["cast", 0, ["BaseField", 8]]
```

spec/src/halt.toml
```diff
@@ -17,31 +17,31 @@ name = "all"
...
 multiplicity = 1
 ref = "halt:c:read_zero_exit_code"
 
 [[constraints.all]]
 kind = "interaction"
 tag = "MEMW"
-input = [1, ["*", 2, "i"], 0, ["-", ["^", 2, 64], 1], 1, 0, 0]
+input = [1, ["cast", ["*", 2, "i"], "DWordWL"], ["cast", 0, ["Byte", 8]], ["cast", ["-", ["^", 2, 64], 1], "DWordWL"], 1, 0, 0]
```
RobinJadoul:
```suggestion
input = [1, ["cast", ["*", 2, "i"], "DWordWL"], ["cast", 0, ["BaseField", 8]], ["cast", ["-", ["^", 2, 64], 1], "DWordWL"], 1, 0, 0]
```

spec/src/halt.toml
erik-3milabs: My question is mostly: why were you allowed to put in "Byte" here, rather than "BaseField"? Why does the type checker not ERROR on this? The signature is quite clearly, `BaseField`, no?

RobinJadoul: Because for similar reasons to before, we can't actually check the range.
Signatures are brought back to (arrays of) ranges, so we can only check that it makes structural sense.
Even if we did have more capabilities, Byte would/should be assignable to BaseField, so that would still be entirely valid.

spec/src/cpu.toml
```diff
@@ -629,33 +627,30 @@ input = ["pc", ["*", ["+", ["*", 2, "c_type_instruction"], ["*", 4, ["not", "c_t
...
 name = "mem"
 prefix = "M"
 
-# TODO: no types available, so no casting yet
 [[constraints.mem]]
 kind = "interaction"
 tag = "MEMW"
-input = [1, ["*", 2, "rs1"], "rv1", ["+", "timestamp", ["cast", 0, "DWordWL"]], 1, 0, 0]
-output = "rv1"
+input = [1, ["*", ["cast", 2, "DWordWL"], "rs1"], ["arr", ["idx", ["cast", "rv1", "DWordWL"], 0], ["idx", ["cast", "rv1", "DWordWL"], 1], 0, 0, 0, 0, 0, 0], ["+", "timestamp", ["cast", 0, "DWordWL"]], 1, 0, 0]
```
erik-3milabs, line 652: While I still don't think this is the neatest way to write this down, I'll accept letting this be for now and move on. Before we do, I do wish we ensure the following:

1.  Somewhere in the document we specify
    - the "cast" operation, and
    - scalar-array multiplication.
2. In `tooling/chip.py` we enforce that casting a constant to an array type requires the constant to fit in the first limb of the array.

spec/src/memw.toml
```diff
@@ -127,29 +127,29 @@ poly = ["*", "w2", ["not", "μ_sum"]]
...
-input = ["base_address", ["+", "i", 1]]
+input = ["base_address", ["cast", ["+", "i", 1], "DWordWL"]]
 output = ["cast", ["idx", "address_add", "i"], "DWordWL"]
 iter = ["i", 3, 6]
 cond = "write8"
 
 [[constraints.consistency]]
 kind = "interaction"
-tag = "IS_HALFWORD"
+tag = "IS_HALF"
```
erik-3milabs, line 152: one of the assumptions in SHIFT still uses `IS_HALFWORD` rather than `IS_HALF`

erik-3milabs: The future optimizations list of `MEMW` also mentions `IS_HALFWORD` rather than `IS_HALF`

spec/src/cpu.toml
RobinJadoul: re 1: I would argue we even want a full on explicit description of the language somewhere, in particular for array and casting behavior, but generally good to have.

RobinJadoul: Created #325 to track this

erik-3milabs (approved):

PR #315, opened 2026-02-10
Remove proof_options from prove/verify API

gabrielbosio:
Hardcode production-secure proof options (100-bit provable security) inside `prove`, `verify`, and `prove_and_verify`. Callers no longer pass `ProofOptions`.

- Add `Default` impl for `ProofOptions` with production values (104 FRI queries, grinding 20, blowup 4)
- Remove `proof_options` from `ProofBundle`
- Remove `default_test_options()` — all tests use `ProofOptions::default()`


MauroToscano:
We can re do it later, and keep both versions, but use on

PR #322, opened 2026-02-11
Eliminate columns2rows transposition copy in Merkle commitment

gabrielbosio:
Adds `MerkleTree::build_from_columns` that hashes virtual rows by striding into column-major data, avoiding the full `columns2rows` transposition allocation. Removes the `columns2rows` function entirely.

### DIFF
--- a/crypto/crypto/src/merkle_tree/backends/field_element_vector.rs
+++ b/crypto/crypto/src/merkle_tree/backends/field_element_vector.rs
@@ -1,6 +1,7 @@
 use core::marker::PhantomData;
 
 use crate::hash::poseidon::Poseidon;
+use crate::merkle_tree::merkle::MerkleTree;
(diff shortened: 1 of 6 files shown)


gabrielbosio:
**Method:** Poll `/proc/PID/status` VmHWM (kernel-tracked peak RSS) every 5 seconds while the CLI prover runs.

**Machine:** `admin@vm-tests-avx512` — Debian 12, x86_64, 62 GB RAM, 16 cores, AVX-512.

Results (2^19 instructions, bench_512k):
- `main`: 50.4 GB peak RSS, ~5.6 min
- `feat/m3-eliminate-columns2rows-copy`: 43.0 GB peak RSS, ~5.2 min

Saves ~7.4 GB (14.7% reduction). No proving time regression.

Issue #325, opened 2026-02-12
spec: Document expression language semantics


RobinJadoul:
- Focus on arrays and casting behaviour
- Clarify scalar-array multiplications (see #312)
- {un,}conditional template syntax (assume `cond` can be an arbitrary degree 1 `BaseField`, unless otherwise noted)
- Probably integrate everything in what is currently the "Variables" chapter

PR #336, opened 2026-02-13
spec: Cleanup, uniformize chapters, make colors work better on web.

RobinJadoul:
The "meaning" of template notation (including `cond`, the absence of `cond` and that unless otherwise noted, `cond` can be any `BaseField` element) is moved to #325

### DIFF
--- a/spec/add.typ
+++ b/spec/add.typ
@@ -10,13 +10,13 @@
 #let add = raw(chip.name)
 #let sub = raw("SUB")
 
-#add is a constraint template that is used to assert that $#`sum` = #`lhs` + #`rhs` mod 2^64$, under the condition that `cond` is non-zero.
+#add is a constraint template that is used to assert that $#`sum` equiv #`lhs` + #`rhs` (mod 2^64)$, under the condition that `cond` is non-zero.
 For ease of notation, we moreover introduce the #sub constraint template
 $
-#`SUB<diff; lhs, rhs>` colon.eq #`ADD<lhs; rhs, diff>`,
+#`SUB<diff; lhs, rhs>` := #`ADD<lhs; rhs, diff>`,
 $
 in both conditional and unconditional versions.
-It constrains that $#`diff` = #`lhs` - #`rhs` mod 2^64$ when the expression `cond` is non-zero.
+It constrains that $#`diff` equiv #`lhs` - #`rhs` (mod 2^64)$ when the expression `cond` is non-zero.
 
 = Variables
 #render_chip_column_table(chip, config)
--- a/spec/add.typ
+++ b/spec/add.typ
@@ -8,33 +8,15 @@
 #show: book-page(chip.name)
 
 #let add = raw(chip.name)
-
-#let highlighted_code(code) = {
-  box(
-    inset: (left: 4pt, right: 4pt), 
-    outset: (top: 4pt, bottom: 4pt), 
-    radius: 2pt,
-    fill: luma(230), 
-    raw(code))
-}
+#let sub = raw("SUB")
 
 #add is a constraint template that is used to assert that $#`sum` = #`lhs` + #`rhs` mod 2^64$, under the condition that `cond` is non-zero.
-
-= Notation
-The #add constraint template has the following interface:
-#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => ADD<sum; lhs, rhs>"))
-where `cond` is any value described by an expression _of degree at most $1$_.
-#highlighted_code("ADD<sum; lhs, rhs>") can be used to denote the _unconditional_ application of the #add template to `lhs`, `rhs`, and `sum`.
-
-#let sub = raw("SUB")
-== #sub
-For ease of notation, we moreover introduce the #sub constraint template.
-Its interface
-#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => SUB<diff; lhs, rhs>"))
-maps onto the #add template as 
-#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => ADD<lhs; rhs, diff>"))
+For ease of notation, we moreover introduce the #sub constraint template
+$
+#`SUB<diff; lhs, rhs>` colon.eq #`ADD<lhs; rhs, diff>`,
+$
+in both conditional and unconditional versions.
 It constrains that $#`diff` = #`lhs` - #`rhs` mod 2^64$ when the expression `cond` is non-zero.
-As with #add, #highlighted_code("SUB<diff; lhs, rhs>") can be used to denote the _unconditional_ application of the template.
 
 = Variables
 #render_chip_column_table(chip, config)
--- a/spec/bitwise.typ
+++ b/spec/bitwise.typ
@@ -15,6 +15,9 @@
 
 #show: book-page(chip.name)
 
+The bitwise chips deal with precomputed lookup tables for bitwise boolean operations
+and convenience functionalities over small domains.
+
 = Columns
 #let nr_variables = total_nr_variables(chip)
 #let nr_columns = total_nr_instantiated_columns(chip, config)
--- a/spec/bitwise.typ
+++ b/spec/bitwise.typ
@@ -14,8 +14,9 @@
 #let bitwise = raw(chip.name)
 
 #show: book-page(chip.name)
+#let bitwise = raw(chip.name)
 
-The bitwise chips deal with precomputed lookup tables for bitwise boolean operations
+The #bitwise chips deal with precomputed lookup tables for bitwise boolean operations
 and convenience functionalities over small domains.
 
 = Columns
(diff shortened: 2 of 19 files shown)


spec/add.typ
```diff
@@ -8,33 +8,15 @@
...
-#let sub = raw("SUB")
-== #sub
-For ease of notation, we moreover introduce the #sub constraint template.
-Its interface
-#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => SUB<diff; lhs, rhs>"))
-maps onto the #add template as 
-#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => ADD<lhs; rhs, diff>"))
+For ease of notation, we moreover introduce the #sub constraint template
+$
+#`SUB<diff; lhs, rhs>` colon.eq #`ADD<lhs; rhs, diff>`,
```
erik-3milabs: Why the `colon.eq` rather than just `:=`? As far as I can see, they are rendered the same, and `:=` in the raw text is more readable imo.

spec/add.typ
```diff
@@ -8,33 +8,15 @@
...
-  box(
-    inset: (left: 4pt, right: 4pt), 
-    outset: (top: 4pt, bottom: 4pt), 
-    radius: 2pt,
-    fill: luma(230), 
-    raw(code))
-}
+#let sub = raw("SUB")
 
 #add is a constraint template that is used to assert that $#`sum` = #`lhs` + #`rhs` mod 2^64$, under the condition that `cond` is non-zero.
```
erik-3milabs: should this technically be an `equiv` rather than `=`? (same applies to SUB below)

spec/bitwise.typ
```diff
@@ -15,6 +15,9 @@
 
 #show: book-page(chip.name)
 
+The bitwise chips deal with precomputed lookup tables for bitwise boolean operations
```
erik-3milabs:
```suggestion
#let bitwise = raw(chip.name)

The #bitwise chips deal with precomputed lookup tables for bitwise boolean operations
```

spec/branch.typ
```diff
@@ -14,6 +14,8 @@
 
 #show: book-page(chip.name)
 
+The BRANCH chip computes the target address of a branching instruction.
```
erik-3milabs:
```suggestion
#let branch = raw(chip.name)

The #branch chip computes the target address of a branching instruction.
```

spec/cpu.typ
```diff
@@ -14,6 +14,9 @@
 
 #show: book-page(chip.name)
 
+The CPU chip is coordinates memory accesses and dispatches to other chips for arithmetic and logical operations.
```
```suggestion
#let cpu = raw(chip.name)

The #cpu chip coordinates memory accesses and dispatches to other chips for arithmetic and logical operations.
```

spec/load.typ
```diff
@@ -14,6 +14,9 @@
 
 #show: book-page(chip.name)
 
+The LOAD chip provides functionality to read values from memory and sign-extend them where appropriate.
```
erik-3milabs:
```suggestion
#let load = raw(chip.name)

The #load chip provides functionality to read values from memory and sign-extend them where appropriate.
```

spec/load.typ
```diff
@@ -14,6 +14,9 @@
...
+It delegates to the MEMW chip (@memw) for lower level memory handling.
```
erik-3milabs: I think this is a more readable ordering of the sentence.

```suggestion
It delegates low-level memory handling to the `MEMW` chip (@memw).
```

spec/lt.typ
```diff
@@ -14,6 +14,8 @@
 
 #show: book-page(chip.name)
 
+The LT chip constrains an indicator bit for the less-than relation, signed or unsigned.
```
erik-3milabs:
```suggestion
#let lt = raw(chip.name)

The #lt chip constrains an indicator bit for the less-than relation, signed or unsigned.
```

spec/memory.typ
```diff
@@ -107,7 +107,8 @@ This raises the question of how to represent timestamps and cleanly perform this
 as over a finite field the “less than” relation is ill-defined
 (though it is common and natural to consider it as the less than relation over the natural lift of the field into the integers).
 We choose to represent timestamps as machine words, using the existing `LT` chip (@lt) functionality for comparisons.
-#rj[Properly link/refer to the LT chip]
+The full implementation of the timestamp system can be seen in the `timestamp` column of the CPU (@cpu) and MEMW chips (@memw).
+The CPU merely passes in the current timestamp, while MEMW can recall the previously written timestamp and constrain the correct sequencing.
```
erik-3milabs:
```suggestion
The full implementation of the timestamp system can be seen in the `timestamp` column of the `CPU` (@cpu) and `MEMW` chips (@memw).
The `CPU` merely passes in the current timestamp, while `MEMW` can recall the previously written timestamp and constrain the correct sequencing.
```

spec/memory.typ
```diff
@@ -213,11 +212,10 @@ and hence doesn't need a column, nor a range check.
 
 == Register initialization/finalization
 
-#rj[Properly link/reference ECALL/HALT chip]
 The initial and final state of registers can be entirely known by
 the verifier, since the relevant initialization values are either zero,
 or embedded in the ELF, and the final values can be set to a known value
-by the HALT ecall.
+by the HALT ecall (@ecall).
```
erik-3milabs:
```suggestion
by the `HALT` ecall (@ecall).
```

spec/memw.typ
```diff
@@ -13,6 +13,11 @@
 
 #show: book-page(chip.name)
 
+The MEMW chip is used to read and write memory locations (both RAM and registers)
```
erik-3milabs:
```suggestion
#let memw = raw(chip.name)

The #memw chip is used to read and write memory locations (both RAM and registers)
```

spec/neg.typ
```diff
@@ -8,22 +8,8 @@
...
-}
-
 #neg is a constraint template that is used to assert that $#`neg` = -#`x`$, under the condition that `cond` is non-zero.
-
-= Notation
-The #neg constraint template has the following interface:
-#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => NEG<neg; x>"))
-where `cond` is a bit value (i.e., lies in ${0, 1}$)  described by an expression _of degree at most $1$_.
-#highlighted_code("NEG<neg; x>") can be used to denote the _unconditional_ application of the #neg template to `x` and `neg` (which is equivalent to $#`cond` = 1$).
+It requires `cond` to be a bit.
```
erik-3milabs, line 12: why drop the part about it being an expression of degree at most 1?

erik-3milabs (commented):
The common denominator in my review is whether a chip's name should be printed as `raw` or not. In my experience, using `raw` sets a chip name apart from other all-caps names such as RAM, PC, ALU, ISA, etc.

If we decide to be consistent in applying raw to chip names, we should apply this consistently in the spec.

spec/neg.typ
erik-3milabs: o, you want to do this in the discussion on templates (#325). I see...

spec/add.typ
RobinJadoul:
```suggestion
#add is a constraint template that is used to assert that $#`sum` equiv #`lhs` + #`rhs` (mod 2^64)$, under the condition that `cond` is non-zero.
```

spec/add.typ
RobinJadoul:
```suggestion
#`SUB<diff; lhs, rhs>` := #`ADD<lhs; rhs, diff>`,
```

spec/add.typ
```diff
@@ -8,33 +8,15 @@
...
-Its interface
-#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => SUB<diff; lhs, rhs>"))
-maps onto the #add template as 
-#block(radius: 5pt, width: 100%, inset: 1.5em, fill: luma(230), raw("cond => ADD<lhs; rhs, diff>"))
+For ease of notation, we moreover introduce the #sub constraint template
+$
+#`SUB<diff; lhs, rhs>` colon.eq #`ADD<lhs; rhs, diff>`,
+$
+in both conditional and unconditional versions.
 It constrains that $#`diff` = #`lhs` - #`rhs` mod 2^64$ when the expression `cond` is non-zero.
```
RobinJadoul:
```suggestion
It constrains that $#`diff` equiv #`lhs` - #`rhs` (mod 2^64)$ when the expression `cond` is non-zero.
```

spec/cpu.typ
```diff
@@ -13,6 +13,10 @@
 #let chip = load_chip("src/cpu.toml", config)
 
 #show: book-page(chip.name)
+#let cpu = raw(chip.name)
+
+The #cpu chip coordinates memory accesses and dispatches to other chips for arithmetic and logical operations.
+It bases its decisions on the entry of the DECODE table (@decode) corresponding the the current program counter (PC).
```
```suggestion
It bases its decisions on the entry of the `DECODE` table (@decode) corresponding the the current program counter (PC).
```

erik-3milabs (approved):
lgtm

PR #288, opened 2026-02-05
feat(math/fft): add Bowers FFT with SoA optimization

diegokingston:
Implement the Bowers G network FFT algorithm with Structure of Arrays (SoA) layout for improved cache utilization and memory access patterns.

Key features:
- LayerTwiddles: Pre-computed twiddles per layer for O(N) sequential access instead of O(N log N) strided access (10-30% speedup on large inputs)
- 2-layer butterfly fusion: Keeps intermediate values in registers to reduce memory traffic
- FftMatrix: SoA layout for batch FFT operations
- Parallel execution via rayon with adaptive threshold (>= 64 blocks)

Includes comprehensive tests comparing against naive DFT and existing Cooley-Tukey implementation.

Ported from lambdaworks with adaptations for lambda_vm's field types.

PR #348, opened 2026-02-20
Integrate Bowers FFT into prover pipeline

gabrielbosio:
Benchmark-only PR to measure end-to-end proving impact of Bowers FFT integration. Based on feat/bowers-fft-cpu (#288) with polynomial.rs wired to use Bowers instead of Cooley-Tukey.


gabrielbosio:
Superseded by #349

PR #355, opened 2026-02-24
Remove unused crypto primitives

gabrielbosio:
Remove cryptographic primitives not used by the Goldilocks-based zkVM.

- Remove BLS12-381 and BN-254 curve implementations
- Remove Stark252 prime field
- Remove binary field implementation
- Remove elliptic curve abstractions (short Weierstrass, traits, point types)
- Remove Pedersen hash parameters and Poseidon/Starknet constants
- Remove `StoneProverTranscript` and Stone compatibility tests
- Remove MSM (multi-scalar multiplication)
- Migrate remaining tests to use Goldilocks field

- It also removes `U64PrimeField` and uses `U64Field` instead which is defined in the `test_fields` directory.
- Removes some borrows when doing operations over fields in certain tests now that the field they use is `FieldElement` over Goldilocks which implements `Copy`.
- Now that tests that used `DefaultTranscript::<Stark252PrimeField>` now use `DefaultTranscript::<GoldilocksField>` so the PR also implements `HasDefaultTranscript` for `GoldilocksField`.


diegokingston (approved):

MauroToscano:
Notice this PR is also doing the following:

(1) migrating prover/src/utils.rs to Goldilocks, 
(2) removing StoneCompatibleSerializer, 
(3) removing public type aliases, and 
(4) removing the test_fiat_shamir feature

MauroToscano (approved):

PR #360, opened 2026-02-25
Opt/higher arity fri

diegokingston:


MauroToscano:
Stale, superseeded by another

PR #363, opened 2026-02-25
Experimental/fri optimizations

jotabulacios:
Benchmark-only PR to measure end-to-end proving impact of FRI optimizations

Optimizations made:
- Early stopping: stop folding when the polynomial degree drops below a configurable bound (`last_layer_degree_bound`), sending the remaining coefficients directly instead of committing more Merkle layers.
- Higher folding factor: fold multiple times per FRI round (configurable power-of-2 `folding_factor`), reducing the number of committed layers and Merkle tree operations.

### DIFF
--- a/crypto/stark/benches/profile_prover.rs
+++ b/crypto/stark/benches/profile_prover.rs
@@ -21,6 +21,8 @@ fn main() {
         fri_number_of_queries: 100,
         coset_offset: 3,
         grinding_factor: 0,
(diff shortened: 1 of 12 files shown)


jotabulacios:
The FRI optimization provides ~20% speedup in isolation, but the end-to-end impact is negligible. Need to check whether it behaves as originally intended.

gabrielbosio:

PR #366, opened 2026-02-26
Perf/allocation cleanup and fri

diegokingston:


diegokingston:
Seems some changes result in regression in performance, needs more investigation. Some FRI cleaning makes sense

PR #372, opened 2026-02-27
spec: SHA256 accelerator

RobinJadoul:


spec/sha256.typ
```diff
@@ -0,0 +1,180 @@
...
+
+== Constraints
+
+First, we gather the dependencies from earlier in the message schedule.
+
+#render_constraint_table(sha256msgschedchip, config, groups: "lookback")
+
+Then, we calculate the result.
+It suffices to check that the overflow of adding four range-checked words
+into a range-checked word is not too big, following the logic from @add.
```
cdesaintguilhem, line 99: This `@add` label doesn't compile for me.

RobinJadoul: Are you compiling ebook.typ or sha256.typ?

spec/src/sha256.toml
```diff
@@ -0,0 +1,293 @@
...
+name = "m_addr"
+type = ["DWordHL", 7]
+desc = "The addresses of the next doublewords of `m`: `m_base_addr + 8 * (i + 1)`"
+pad = ["arr", 8, 16, 24, 32, 40, 48, 56]
+
+[[variables.output]]
+name = "out"
+type = ["Byte", 32]
+desc = "The new state."
+pad = 0
```
erik-3milabs, line 37: this seems to be lacking range check. It definitely is not implicit in C18, and C10's MEMW assumes it is done on the calling side.

spec/src/sha256.toml
```diff
@@ -0,0 +1,293 @@
...
+[[variables.input]]
+name = "m_base_addr"
+type = "DWordWL"
+desc = "The starting address of `m`."
+pad = 0
+
+[[variables.input]]
+name = "m_addr"
+type = ["DWordHL", 7]
+desc = "The addresses of the next doublewords of `m`: `m_base_addr + 8 * (i + 1)`"
```
erik-3milabs:
```suggestion
desc = "The addresses of the next doublewords of `m`: $#`m_base_addr` + 8 dot (i + 1)$"
```

spec/src/sha256.toml
```diff
@@ -0,0 +1,293 @@
...
+[[variables.input]]
+name = "h_base_addr"
+type = "DWordWL"
+desc = "The starting address of `h`."
+pad = 0
+
+[[variables.input]]
+name = "h_addr"
+type = ["DWordHL", 3]
+desc = "The addresses of the next doublewords of `h`: `h_base_addr + 8 * (i + 1)`"
```
erik-3milabs:
```suggestion
desc = "The addresses of the next doublewords of `h`: $#`h_base_addr` + 8 dot (i + 1)$"
```

spec/src/sha256.toml
```diff
@@ -0,0 +1,293 @@
...
+pad = ["arr", 8, 16, 24]
```
erik-3milabs: Could we not apply the same trick here as we did in MEMW? i.e., rather than storing the `h_addr` in full, have the prover provide the carry bits and bank on the fact that an incorrect carry bit will knock the memory argument out of balance.

spec/src/sha256msgsched.toml
```diff
@@ -0,0 +1,155 @@
...
+[[variables.auxiliary]]
+name = "back7"
+type = "Word"
+desc = "`w[index - 7]`"
+pad = 0
+
+[[variables.auxiliary]]
+name = "s0"
+type = "Word"
+desc = "`ROTXOR[back15, 7, 18, 3, 0]`"
```
erik-3milabs: This notation suggests that this is its definition, rather than its description. Consider switching to `back15 >>> 7 XOR back15 >>> 18 XOR back15 >>> 3` or something similar.

If you insist on sticking with ROTXOR: this is probably some an old ROTXOR interface (?); it no longer aligns with the actual ROTXOR interface anymore.

The same applies to s1 below

spec/src/sha256msgsched.toml
```diff
@@ -0,0 +1,155 @@
...
+name = "s1"
+type = "Word"
+desc = "`ROTXOR[back2, 17, 19, 10, 0]`"
+pad = 0
+
+[[variables.virtual]]
+name = "carry"
+type = "Byte"
+desc = "The overflow of computing `out`"
+def = ["*", ["^", 2, -32], ["-", ["cast", "out", "Word"], "back16", "s0", "back7", "s1"]]
```
erik-3milabs: I believe that we have to reverse the order of the elements inside the "-" here as negative values of carry will not pass the IS_BYTE check. This ordering seems to be wrong in several places inside the SHA256 chip-suite, so have a close look at all of them.

Also, this being called `carry` and the description talking about "overflow" is confusing. Consider sticking with a single term.

spec/src/sha256msgsched.toml
```diff
@@ -0,0 +1,155 @@
...
+name = "back16"
+type = "Word"
+desc = "`w[index - 16]`"
+pad = 0
+
+[[variables.auxiliary]]
+name = "back7"
+type = "Word"
+desc = "`w[index - 7]`"
+pad = 0
```
erik-3milabs, line 48: While I understand that this is the order in which they're used, I think readability will improve by presenting them in sorted fashion:


```suggestion
[[variables.auxiliary]]
name = "back2"
type = "Word"
desc = "`w[index - 2]`"
pad = 0

[[variables.auxiliary]]
name = "back7"
type = "Word"
desc = "`w[index - 7]`"
pad = 0

[[variables.auxiliary]]
name = "back15"
type = "Word"
desc = "`w[index - 15]`"
pad = 0

[[variables.auxiliary]]
name = "back16"
type = "Word"
desc = "`w[index - 16]`"
pad = 0
```

spec/src/sha256msgsched.toml
```diff
@@ -0,0 +1,155 @@
+name = "SHA256MSGSCHED"
+
+[[variables.input]]
+name = "timestamp"
+type = "DWordWL"
+desc = "The timestamp/identifier for this execution of the "
```
erik-3milabs: todo: complete sentence

spec/sha256.typ
```diff
@@ -0,0 +1,180 @@
...
+
+#render_chip_assumptions(sha256msgschedchip, config)
+
+== Constraints
+
+First, we gather the dependencies from earlier in the message schedule.
+
+#render_constraint_table(sha256msgschedchip, config, groups: "lookback")
+
+Then, we calculate the result.
```
erik-3milabs, line 97: Let's highlight here that `out` is constrained through the definition of `carry` and the `IS_BYTE` constraint on it (and the range check); it is a nifty trick and took me some time to understand what is going on.

spec/src/sha256msgsched.toml
```diff
@@ -0,0 +1,155 @@
...
+kind = "interaction"
+tag = "SHA256_M"
+input = ["timestamp", ["-", "index", 16], "back16"]
+multiplicity = "μ"
+
+[[constraints.lookback]]
+kind = "interaction"
+tag = "SHA256_M"
+input = ["timestamp", ["-", "index", 7], "back7"]
+multiplicity = "μ"
```
erik-3milabs, line 113: I propose a reorder here as well:


```suggestion
[[constraints.lookback]]
kind = "interaction"
tag = "SHA256_M"
input = ["timestamp", ["-", "index", 2], "back2"]
multiplicity = "μ"

[[constraints.lookback]]
kind = "interaction"
tag = "SHA256_M"
input = ["timestamp", ["-", "index", 7], "back7"]
multiplicity = "μ"

[[constraints.lookback]]
kind = "interaction"
tag = "SHA256_M"
input = ["timestamp", ["-", "index", 15], "back15"]
multiplicity = "μ"

[[constraints.lookback]]
kind = "interaction"
tag = "SHA256_M"
input = ["timestamp", ["-", "index", 16], "back16"]
multiplicity = "μ"
```

spec/src/sha256.toml
```diff
@@ -0,0 +1,293 @@
...
+[[constraints.sched]]
+kind = "interaction"
+tag = "SHA256_M"
+input = ["timestamp", "i", ["+",
+  ["*", ["^", 2, 0], ["idx", "m", ["+", ["*", 4, "i"], 3]]],
+  ["*", ["^", 2, 8], ["idx", "m", ["+", ["*", 4, "i"], 2]]],
+  ["*", ["^", 2, 16], ["idx", "m", ["+", ["*", 4, "i"], 1]]],
+  ["*", ["^", 2, 24], ["idx", "m", ["+", ["*", 4, "i"], 0]]],
+]]
+multiplicity = ["*", -4, "μ"]
```
erik-3milabs: 1. I believe $-4\mu$ should not be constant for [0, 15]. Rather, I think that it it should be

$i = 0: -\mu$
$i \in [1, 8]: -2\mu$
$i \in [9, 13]: -3\mu$
$i \in [14, 15]: -4\mu$

since w[0] is only looked back at when constructing w[16], w[1] only when constructing w[16] and w[17], etc.

2. these (and other) SHA256_M values are completely consumed during message scheduling (and thus cannot be ingested during compression). As such, the output multiplicity should be increased by 1 for the [0, 61] range compared to what it is now (but not for 62 and 63, as those are never consumed during message scheduling).

In other words: the multiplicities need to be looked at.

spec/src/sha256.toml
```diff
@@ -0,0 +1,293 @@
...
+  ["*", ["^", 2, 16], ["idx", "h", ["+", ["*", 4, "i"], 1]]],
+  ["*", ["^", 2, 24], ["idx", "h", ["+", ["*", 4, "i"], 0]]],
+]]]
+output = ["arr", 0, ["+",
+  ["*", ["^", 2, 0], ["idx", "out", ["+", ["*", 4, "i"], 3]]],
+  ["*", ["^", 2, 8], ["idx", "out", ["+", ["*", 4, "i"], 2]]],
+  ["*", ["^", 2, 16], ["idx", "out", ["+", ["*", 4, "i"], 1]]],
+  ["*", ["^", 2, 24], ["idx", "out", ["+", ["*", 4, "i"], 0]]],
+]]
+iter = ["i", 0, 7]
```
erik-3milabs, line 256: Question: this uses `ADD` to perform addition mod 2^32, while on the MSGSCHED chip this is done by computing the `carry` and asserting `IS_BYTE[carry]`. Is there a reason for using either in either place? If not, perhaps we should decide on a canonical approach, rather than having separate approaches (where ADD would probably be preferable as it does not use the `IS_BYTE` interaction?)

spec/src/rotxor.toml
```diff
@@ -0,0 +1,225 @@
...
+kind = "interaction"
+tag = "HWSL"
+input = [["idx", "a", "i"], ["-", 16, "r0_low"]]
+output = ["idx", "a0_left", "i"]
+iter = ["i", 0, 1]
+multiplicity = "μ"
+
+[[constraints.shift]]
+kind = "interaction"
+tag = "HWSLC"
```
erik-3milabs: HWSLC has been merged into HWSL. Please update the PR to reflect this change.

spec/src/rotxor.toml
```diff
@@ -0,0 +1,225 @@
...
+input = [["idx", "a", "i"], ["-", 16, "r2_low"]]
+output = ["idx", "a2_right", "i"]
+iter = ["i", 0, 1]
+multiplicity = "μ"
+
+[[constraints.shift]]
+kind = "arith"
+constraint = "$#`a0[i]` = (1 - #`r0_high`) * (#`a0_left[1 - i]` + #`a0_right[i]`) + #`r0_high` * (#`a0_left[i]` + #`a0_right[i - i]`)$"
```
erik-3milabs: 1. Consider using `dot` rather than `*`
2. The `i-i` should be `1-i`.

spec/src/rotxor.toml
```diff
@@ -0,0 +1,225 @@
...
+poly = ["-", ["idx", ["cast", "a0", "WordHL"], "i"], ["*", ["not", "r0_high"], ["+", ["idx", "a0_left", ["-", 1, "i"]], ["idx", "a0_right", "i"]]], ["*", "r0_high", ["+", ["idx", "a0_left", "i"], ["idx", "a0_right", ["-", 1, "i"]]]]]
+iter = ["i", 0, 1]
+
+[[constraints.shift]]
+kind = "arith"
+constraint = "$#`a1[i]` = (1 - #`r1_high`) * (#`a1_left[1 - i]` + #`a1_right[i]`) + #`r1_high` * (#`a1_left[i]` + #`a1_right[i - i]`)$"
```
erik-3milabs: See comments on `a0`

spec/src/rotxor.toml
```diff
@@ -0,0 +1,225 @@
...
+name = "r1_low"
+type = "Byte"
+desc = "The second amount of rotation, low nibble"
+pad = 0
+
+[[variables.input]]
+name = "r1_high"
+type = "Bit"
+desc = "The second amount of rotation, high bit"
+pad = 0
```
erik-3milabs: By computing `a1` based on `a0` rather than `a`, things can be rearranged in such a way that `r1` can be dropped.

spec/src/rotxor.toml
```diff
@@ -0,0 +1,225 @@
...
+name = "r0_low"
+type = "Byte"
+desc = "The first amount of rotation, low nibble"
+pad = 0
+
+[[variables.input]]
+name = "r0_high"
+type = "Bit"
+desc = "The first amount of rotation, high bit"
+pad = 0
```
erik-3milabs: Given that XOR is commutative, we're allowed to chose the order in which to pass the shifts into ROTXOR. In particular, we can do
s0 = rotxor(w[i-15], 18, 7, 3, false)
s1 = rotxor(w[i-2], 19, 17, 10, false)
S1 = rotxor(e, 25, 11, 6, true)
S0 = rotxor(a, 22, 13, 2, true)

Now, observe that these can be written as
s0 = rotxor(w[i-15], 16+2, 7, 3, false)
s1 = rotxor(w[i-2], 16+3, 17, 10, false)
S1 = rotxor(e, 16+9, 11, 6, true)
S0 = rotxor(a, 16+6, 13, 2, true)

and if we compute a1 based on a0 rather than a, we can do
s0 = rotxor(w[i-15], 16+2, -11, 3, false)
s1 = rotxor(w[i-2], 16+3, -2, 10, false)
S1 = rotxor(e, 16+9, -14, 6, true)
S0 = rotxor(a, 16+6, -9, 2, true)

Hence, if we modify ROTXOR such that r0_high = 1 is hardcoded and it computes the second value as a _left_-rotation of a0, we don't need any high-bits.

spec/src/rotxor.toml
```diff
@@ -0,0 +1,225 @@
...
+type = "Bit"
+desc = ""
+pad = 0
+
+
+[[assumptions]]
+desc = "$1 <= #`r0_low`, #`r1_low`, #`r2_low` < 16$"
+
+[[assumptions]]
+desc = "$0 <= #`r0_high`, #`r1_high` < 2$"
```
erik-3milabs: I find $\in$ to be more readable with comma's than $\leq$

spec/src/sha256round.toml
```diff
@@ -0,0 +1,286 @@
...
+name = "carry_a"
+type = "Byte"
+desc = "The carry from `out_a`"
+def = ["*", ["^", 2, -32], ["-", ["cast", "out_a", "Word"], "temp1", "temp2"]]
+
+[[variables.virtual]]
+name = "carry_e"
+type = "Byte"
+desc = "The carry from `out_e`"
+def = ["*", ["^", 2, -32], ["-", ["cast", "out_e", "Word"], "d", "temp1"]]
```
erik-3milabs: the order of the elements in the "-" should be inverted; otherwise the carry may not be a Byte.
Also applies to `carry_a`

spec/src/sha256round.toml
```diff
@@ -0,0 +1,286 @@
...
+output = ["idx", "not_e_and_g", "i"]
+multiplicity = "μ"
+iter = ["i", 0, 3]
+
+[[constraints.value]]
+kind = "interaction"
+tag = "SHA256_K"
+input = ["index"]
+output = "kval"
+multiplicity = "μ"
```
erik-3milabs, line 223: It is not made explicit where/how these constants are supposed to be added into the logup

spec/src/sha256round.toml
```diff
@@ -0,0 +1,286 @@
...
+name = "e_and_f"
+type = "WordBL"
+desc = "Part of `ch`"
+pad = 0
+
+[[variables.auxiliary]]
+name = "not_e_and_g"
+type = "WordBL"
+desc = "Part of `ch`"
+pad = 0
```
erik-3milabs, line 91: I propose we expand the description to include a description of what the variable is supposed to contain (and not just that it is part of `ch`).

```suggestion
[[variables.auxiliary]]
name = "not_e_and_g"
type = "WordBL"
desc = "$(not e) math.amp g$. Part of `ch`"
pad = 0
```

This is particularly important for this variable, since it is unclear whether this is `(not e) and g` or `not (e and g)`.

I propose we apply this here and for the other parts of `ch` and `maj`.

spec/sha256.typ
```diff
@@ -0,0 +1,180 @@
...
+It suffices to check that the overflow of adding four range-checked words
+into a range-checked word is not too big, following the logic from @add.
+#render_constraint_table(sha256msgschedchip, config, groups: "calc")
+
+Finally, we contribute to the LogUp.
+#render_constraint_table(sha256msgschedchip, config, groups: "output")
+
+= #sha256round chip
```
erik-3milabs, line 107: This chapter leans heavily on the naming scheme used on the wiki page. I propose we (at the very least) add a link that page, or even provide some pseudo-code here to clarify what all of these one or two-letter variables represent.

spec/sha256.typ
```diff
@@ -0,0 +1,180 @@
...
+by distribution.
+Additionally, since for this form, $(a bitand b)$ and $(a xor b)$ are disjoint, so are $(a bitand b)$ and $(c bitand (a xor b))$,
+and hence we can replace that top-level XOR with a field addition to compute $(a bitand b) + (c bitand (a xor b))$,
+needing fewer intermediate columns.
+Similarly, `ch` can be written as $(e bitand f) + ((2^32 - 1 - e) bitand g)$.
+#render_constraint_table(sha256roundchip, config, groups: "value")
+
+Then we constrain the addition for the new state.
+Since `out_e` is the range-checked sum of three range-checked words, we
+can constrain `carry_e` to be 0, 1 or 2 with a degree 3 constraint instead of a lookup.
```
erik-3milabs: I believe this reasoning to be incorrect:  `tmp1 = h + S1 + ch + kval + wval` is not range checked, so we are talking about six rather than 3 range-checked Words. Worst case, carry can be equal to 5.

spec/src/sha256round.toml
```diff
@@ -0,0 +1,286 @@
...
+name = "out_e"
+type = "WordHL"
+desc = "d + temp1"
+pad = 0
+
+[[variables.auxiliary]]
+name = "S1"
+type = "Word"
+desc = "Transformation of `e`"
+pad = 0
```
erik-3milabs, line 79: I propose to rearrange the aux vars, preferably in a semi-alphebetically-sorted fashion.

spec/src/sha256round.toml
```diff
@@ -0,0 +1,286 @@
...
+desc = "`k[index]`"
+pad = 0
+
+[[variables.auxiliary]]
+name = "wval"
+type = "Word"
+desc = "`w[index]`"
+pad = 0
+
+[[variables.virtual]]
```
erik-3milabs, line 141: I propose to rearrange the virtual variables, preferably in a semi-alphabetically-sorted fashion.

spec/src/sha256round.toml
```diff
@@ -0,0 +1,286 @@
...
+type = "Bit"
+desc = ""
+pad = 0
+
+[[assumptions]]
+desc = "All state values are valid words"
+
+
+[[constraint_groups]]
+name = "value"
```
erik-3milabs, line 176: the constraints in this group would benefit from a slight rearrange.

spec/sha256.typ
erik-3milabs, line 1: Am I correct in understanding that this chip does _not_ perform the i) initialization and ii) preprocessing steps? I.e., the caller is supposed to deal with that themselves?

erik-3milabs (changes requested):
TL;DR: I believe that
1. the readability of some things (variables, constraints) would benefit from a rearrange,
2. some carries are computed incorrectly (the order of the sum elements should be inverted),
3. there are some tiny bugs/missing checks,
4. some descriptions are worth updating a bit.

spec/src/sha256.toml
RobinJadoul: Technically, maybe.
It's a bit more involved, since we'd be willfully ignoring the contract (assumptions) of the MEMW chip, which could become quite fragile.
The alternative is to do that, and hit the memory table directly, bypassing MEMW.

spec/src/sha256.toml
RobinJadoul: The `IS_BYTE` trick is so that we're able to add multiple DWords at the same time, without having to store range-checked intermediate results.
In this case, we're only adding two DWords, and we already need to materialize and range-check the output, so the regular `ADD` can apply without having to materialize the bits or rely on an `IS_BYTE` lookup

spec/src/rotxor.toml
RobinJadoul: Yep, I had a local update for that after HWSLC disappeared, but it didn't make it into a commit yet.

spec/src/rotxor.toml
RobinJadoul: I looked at doing something like that, but I believe that the re-used across message schedule and compression prevented it.
Will check again.

spec/sha256.typ
RobinJadoul: Correct, we're speeding up a single compression in the Merkle-Damgaard construction, initializing the state with the prescribed constants, padding, and chunking are to be done by regular RISC-V code.
There's some tradeoff between granularity of the accelerator and possible speedup, but I think that these parts wouldn't benefit in particular from acceleration, and would introduce even more complexity.

spec/src/rotxor.toml
```diff
@@ -0,0 +1,225 @@
...
+name = "a01"
+type = "WordBL"
+desc = "$a_0 xor a_1$"
+pad = 0
+
+[[variables.multiplicity]]
+name = "μ"
+type = "Bit"
+desc = ""
+pad = 0
```
erik-3milabs, line 103: Should we not be able to make this a fixed-size table when we have `mu: BaseField` here? After all, this only takes a Byte (256 possible values) and six different rot sets, so this should fit in a 2^11 length table.

RobinJadoul: It acts on a Word, doesn't it?

erik-3milabs: Good point. nvm

spec/src/rotxor.toml
RobinJadoul:
```suggestion
[[assumptions]]
desc = "$#`r0_low`, #`r1_low`, #`r2_low` in [1, 15]$"

[[assumptions]]
desc = "$#`r0_high`, #`r1_high` in [0, 1]$"
```

spec/src/rotxor.toml
RobinJadoul:
```suggestion
constraint = "$#`a0[i]` = (1 - #`r0_high`) dot (#`a0_left[1 - i]` + #`a0_right[i]`) + #`r0_high` dot (#`a0_left[i]` + #`a0_right[1 - i]`)$"
```

spec/src/rotxor.toml
RobinJadoul:
```suggestion
constraint = "$#`a1[i]` = (1 - #`r1_high`) dot (#`a1_left[1 - i]` + #`a1_right[i]`) + #`r1_high` dot (#`a1_left[i]` + #`a1_right[1 - i]`)$"
```

spec/src/sha256msgsched.toml
RobinJadoul:
```suggestion
desc = "The timestamp/identifier for this execution of the message schedule"
```

spec/src/sha256msgsched.toml
RobinJadoul:
```suggestion
desc = "$#`back15` >>> 7 xor #`back15` >>> 18 xor #`back15` >> 3$"
```

spec/src/sha256msgsched.toml
```diff
@@ -0,0 +1,155 @@
...
+[[variables.auxiliary]]
+name = "s0"
+type = "Word"
+desc = "`ROTXOR[back15, 7, 18, 3, 0]`"
+pad = 0
+
+[[variables.auxiliary]]
+name = "s1"
+type = "Word"
+desc = "`ROTXOR[back2, 17, 19, 10, 0]`"
```
RobinJadoul:
```suggestion
desc = "$#`back2` >>> 17 xor #`back2` >>> 19 xor #`back2` >> 10$"
```

spec/src/sha256msgsched.toml
```diff
@@ -0,0 +1,155 @@
...
+pad = 0
+
+[[variables.virtual]]
+name = "carry"
+type = "Byte"
+desc = "The overflow of computing `out`"
```
RobinJadoul:
```suggestion
desc = "The carry of computing `out`"
```

spec/src/sha256msgsched.toml
RobinJadoul:
```suggestion
def = ["*", ["^", 2, -32], ["-", ["+", "back16", "s0", "back7", "s1"], ["cast", "out", "Word"]]
```

spec/src/sha256round.toml
RobinJadoul:
```suggestion
[[variables.auxiliary]]
name = "not_e_and_g"
type = "WordBL"
desc = "$(not e) class(\"binary\", amp) g$. Part of `ch`"
pad = 0
```

spec/src/sha256round.toml
RobinJadoul:
```suggestion
[[variables.virtual]]
name = "carry_a"
type = "Byte"
desc = "The carry from `out_a`"
def = ["*", ["^", 2, -32], ["-", ["+", "temp1", "temp2"], ["cast", "out_a", "Word"]]]

[[variables.virtual]]
name = "carry_e"
type = "Byte"
desc = "The carry from `out_e`"
def = ["*", ["^", 2, -32], ["-", ["+", "d", "temp1"], ["cast", "out_e", "Word"]]]
```

RobinJadoul (commented):
quick pass to prepare suggestions

spec/sha256.typ
```diff
@@ -0,0 +1,180 @@
...
+#render_chip_assumptions(sha256msgschedchip, config)
+
+== Constraints
+
+First, we gather the dependencies from earlier in the message schedule.
+
+#render_constraint_table(sha256msgschedchip, config, groups: "lookback")
+
+Then, we calculate the result.
+It suffices to check that the overflow of adding four range-checked words
```
RobinJadoul:
```suggestion
It suffices to check that the carry of adding four range-checked words
```

spec/src/sha256round.toml
RobinJadoul: There's a short sentence about it in the accompanying typst:
> It could either be instantiated with a (short) precomputed table, or through hardcoded LogUp contributions in this chip.

It could be worthwhile to add this as its own table similar to the precomputed bitwise ones. Integrating it into the driving table (or future kernel chip) will probably just end up objectively worse.

spec/sha256.typ
RobinJadoul: I... have no idea anymore where I saw a sum with only three terms :thinking: 
(Maybe read tmp1 as tmp2?)

spec/sha256.typ
```diff
@@ -28,15 +28,17 @@ to perform the message schedule and the compression rounds, respectively.
...
   cases(
-    (a >>> r_0) xor (a >>> r_1) xor (a >>> r_2) quad "if" #`last_rot`,
-    (a >>> r_0) xor (a >>> r_1) xor (a >> r_2) quad "if" #`!last_rot`
+    (a >>> (16 + r_0)) xor (a >>> (16 + r_0 - r_1)) xor (a >>> r_2) quad "if" #`last_rot`,
+    (a >>> (16 + r_0)) xor (a >>> (16 + r_0 - r_1)) xor (a >> r_2) quad "if" #`!last_rot`
   ),
 $
 where we let $>>>$ denote right rotation and $>>$ logical shift right.
 
+Most of the structure and variable naming follows the pseudocode at #link("https://en.wikipedia.org/wiki/SHA-2#Pseudocode").
```
erik-3milabs: I propose to use an internet archive link here. Also, I'd rewrite this as


```suggestion
Most of the structure and variable naming follows #link("https://en.wikipedia.org/wiki/SHA-2#Pseudocode", "this pseudocode").
```

where we should add some color to the link

spec/src/sha256msgsched.toml
```diff
@@ -24,56 +24,53 @@ type = "WordHL"
...
 name = "μ"
 type = "Bit"
 desc = ""
 pad = 0
 
-[[assumptions]]
-desc = "$#`index` >= 16$"
-
 [[assumptions]]
 desc = "#`IS_WORD[SHA256_M[i]]` for $0 <= i < #`index`$"
```
erik-3milabs: This feels a touch too informal/inconsistent, given that this does not at all match the SHA256_M signature. Though, I get that it is a bit weird. Hmmm, what about

for all `i in <range>: SHA_256[timestamp, i, X] => IS_WORD[X]`?

spec/src/sha256round.toml
```diff
@@ -73,51 +73,51 @@ desc = "d + temp1"
 pad = 0
 
 [[variables.auxiliary]]
-name = "S1"
-type = "Word"
-desc = "Transformation of `e`"
+name = "a_and_b"
+type = "WordBL"
+desc = "$a class(\"binary\", amp) b$. Part of `maj`"
```
erik-3milabs: I've thusfar used `code notation` to refer to other variables in descriptions. In this case, we'd get

```suggestion
name = "a_and_b"
type = "WordBL"
desc = "$#`a` class(\"binary\", amp) #`b`$. Part of `maj`"
```

I propose we apply this consistently for all variables in all SHA chips

spec/sha256.typ
erik-3milabs: I believe this to be outdated / incomplete. Perhaps we should add a brief overview of the weird stuff we do up front here.

spec/src/rotxor.toml
erik-3milabs: I believe we can make this a BaseField? If you insist on keeping it a bit, `IS_BIT<mu>` should be added.

spec/sha256.typ
erik-3milabs: Excellent. We should clarify this at the start of the SHA256 spec. It is always a good idea to set proper expectations

spec/sha256.typ
```diff
@@ -0,0 +1,198 @@
...
+#render_constraint_table(rotxorchip, config, groups: "output")
+
+== Padding
+
+#render_chip_padding_table(rotxorchip, config)
+
+= Constant lookup
+
+#let sha256_kchip = load_chip("src/sha256consts.toml", config)
+#let sha256_k = sha256_kchip.name
```
erik-3milabs:
```suggestion
#let sha256_k = raw(sha256_kchip.name)
```

spec/src/signatures.toml
```diff
@@ -187,3 +187,26 @@ output = ["Half", 2]
...
+[[signatures]]
+tag = "SHA256_K"
+kind = "interaction"
+input = ["BaseField"]
+output = "Word"
+
+[[signatures]]
+tag = "SHA256_M"
+kind = "interaction"
+input = ["DWordWL", "BaseField", "Word"]
```
erik-3milabs: Out of curiosity: what is your reason for not having `wval` as an `output` in this interaction?

I know the concept of "output" is rather arbitrary, but we are still using it (if only to improve the readability). I would've expected this to be an output

erik-3milabs (changes requested):
some minor things I've found now. I plan to do another thorough readthrough tomorrow

spec/sha256.typ
RobinJadoul: Agree that the archive link may be useful; I'm in general not a huge fan of having href's without readable/copyable url in a pdf, so maybe making it into a footnote is a better option here.

spec/sha256.typ
RobinJadoul: Hmm, true, we should get rid of the outdated info.
The weird stuff, I've tried to preempt by describing what we compute in ROTXOR at the start, in terms of purely the values in the signature, and not thinking about any "ellided" information.

spec/src/signatures.toml
RobinJadoul: Hmm, not really any reason behind it that I can remember anymore; especially seeing SHA256_K doing it with an output

RobinJadoul: That would also solve some of the weirdness with the assumption

spec/sha256.typ
```diff
@@ -0,0 +1,203 @@
...
+#let sha256roundchip = load_chip("src/sha256round.toml", config)
+#let rotxorchip = load_chip("src/rotxor.toml", config)
+#let sha256 = raw(sha256chip.name)
+#let sha256msgsched = raw(sha256msgschedchip.name)
+#let sha256round = raw(sha256roundchip.name)
+#let rotxor = raw(rotxorchip.name)
+
+The purpose of this set of chips is to provide an accelerator for the SHA256 compression function.
+They do not intend to handle the requisite repeated invocations of this compression,
+nor with padding or the preparation of the initial state.
```
erik-3milabs:
```suggestion
The purpose of this set of chips is to provide an accelerator for SHA256's compression function _only_;
they do not pad the input, prepare the initial state or handle the requisite repeated invocations of this compression as required by the SHA256 hash function.
```

spec/src/sha256round.toml
```diff
@@ -0,0 +1,288 @@
...
+name = "temp1"
+type = "BaseField"
+desc = "temp1 value"
+def = ["+", "h", "S1", "ch", "kval", "wval"]
+
+[[variables.virtual]]
+name = "temp2"
+type = "BaseField"
+desc = "temp2 value"
+def = ["+", "S0", "maj"]
```
erik-3milabs, line 163:
```suggestion
[[variables.virtual]]
name = "temp1"
type = "BaseField"
desc = "`temp1` value"
def = ["+", "h", "S1", "ch", "kval", "wval"]

[[variables.virtual]]
name = "temp2"
type = "BaseField"
desc = "`temp2` value"
def = ["+", "S0", "maj"]
```

spec/sha256.typ
```diff
@@ -38,7 +41,7 @@ $
 $
 where we let $>>>$ denote right rotation and $>>$ logical shift right.
```
erik-3milabs: Given that the ROTXOR chip is (currently) internal to the SHA256 chip, I don't see the need for stating the interface here already. Rather I'd move this to the rotxor subsection. This would allow the chip's definition to be closer to the spec, which I prefer.

spec/sha256.typ
```diff
@@ -0,0 +1,203 @@
...
+
+The #sha256 chip leverages #nr_variables variables, spanning #nr_columns columns:
+#render_chip_column_table(sha256chip, config)
+
+== Constraints
+
+The first responsibility of the chip is to read the current state and message chunk from memory,
+passed as arguments through pointers.
+Since the memory ranges could overlap, we read the chunk first, before reading and writing the state at the next timestamp.
+The state is passed in argument `A0 = x10`, and the chunk as `A1 = x11`.
```
erik-3milabs:
```suggestion
References to the addresses containing the state and the current chunk are passed in as arguments `A0 = x10` and `A1 = x11`, respectively.
```

spec/src/sha256.toml
```diff
@@ -0,0 +1,312 @@
...
+name = "h_base_addr"
+type = "DWordWL"
+desc = "The starting address of `h`."
+pad = 0
+
+[[variables.input]]
+name = "h_addr"
+type = ["DWordHL", 3]
+desc = "The addresses of the next doublewords of `h`: $#`h_base_addr` + 8 dot (i + 1)$"
+pad = ["arr", 8, 16, 24]
```
erik-3milabs: At the cost of 2x2 columns, I propose we merge h_base_addr/h_addr (resp. m_base_addr/m_addr).

With this chip already spanning 180 columns, adding four more isn't going to change performance much. Moreover, we're planning on replacing it with a kernel chip anyway.

The upside would be that constraints C4-C5, and C9-C10 can be merged in a way that significantly improves readability.

I would be OK with listing reverting to WL + 3xHL as a potential improvement.

spec/src/sha256.toml
```diff
@@ -0,0 +1,312 @@
...
+input = [1, ["cast", ["*", 2, 11], "DWordWL"], ["arr", ["idx", "m_base_addr", 0], ["idx", "m_base_addr", 1], 0, 0, 0, 0, 0, 0], "timestamp", 1, 0, 0]
+output = ["arr", ["idx", "m_base_addr", 0], ["idx", "m_base_addr", 1], 0, 0, 0, 0, 0, 0]
+multiplicity = "μ"
+
+[[constraints.memory]]
+kind = "interaction"
+tag = "MEMW"
+input = [0, "m_base_addr", ["arr", ["idx", "m", 3], ["idx", "m", 2], ["idx", "m", 1], ["idx", "m", 0], ["idx", "m", 7], ["idx", "m", 6], ["idx", "m", 5], ["idx", "m", 4]], "timestamp", 0, 0, 1]
+output = ["arr", ["idx", "m", 3], ["idx", "m", 2], ["idx", "m", 1], ["idx", "m", 0], ["idx", "m", 7], ["idx", "m", 6], ["idx", "m", 5], ["idx", "m", 4]]
+multiplicity = "μ"
```
erik-3milabs: to improve the flow of reading these constraints, I propose we order them as
1. read *_base_addr
2. compute the other x addresses from base_addr
3. range checks required by addition (could also be swapped with 2)
4. load the data from *_addr

applies to both m and h

spec/src/sha256.toml
```diff
@@ -0,0 +1,312 @@
...
+name = "last_round_out"
+type = ["Word", 8]
+desc = "The output from the last compression round"
+pad = 0
+
+[[variables.multiplicity]]
+name = "μ"
+type = "Bit"
+desc = ""
+pad = 0
```
erik-3milabs, line 49: mu being a bit is not enforced by a constraint. Technically it is not necessary, since mu>1 would mess up the logup, but might still be nice to include it. just to be safe

spec/src/sha256round.toml
```diff
@@ -0,0 +1,288 @@
+name = "SHA256ROUND"
+
+[[variables.input]]
+name = "timestamp"
+type = "DWordWL"
+desc = "The timestamp/identifier for this execution of the "
```
erik-3milabs: this sentence is not complete

erik-3milabs (approved):
Some last, comments. Aside from that, lgtm.

spec/sha256.typ
RobinJadoul: Hmm, I think the reason I moved it up, was to have an idea what the thing is doing when you're using it in the message schedule and round function, otherwise it's very much "what is going on, the constants don't even match"

spec/sha256.typ
RobinJadoul: Not convinced; it's a good intro of purpose, even when considered without the limitations.

spec/sha256.typ
RobinJadoul:
```suggestion
The addresses containing the state and the current chunk are passed in as arguments `A0 = x10` and `A1 = x11`, respectively.
```

spec/src/sha256.toml
RobinJadoul: Hmm, maybe, but I don't love that we then need `["idx", ["cast", ["idx", "m_addr", 0], "DWordWL"], 0]` for the register reads

RobinJadoul: (kept it in a hopefully easy-to-revert commit, see how it looks)

spec/src/sha256.toml
```diff
@@ -0,0 +1,268 @@
...
+input = [1, ["cast", ["*", 2, 11], "DWordWL"], ["arr", ["idx", ["cast", ["idx", "m_addr", 0], "DWordWL"], 0], ["idx", ["cast", ["idx", "m_addr", 0], "DWordWL"], 1], 0, 0, 0, 0, 0, 0], "timestamp", 1, 0, 0]
+output = ["arr", ["idx", ["cast", ["idx", "m_addr", 0], "DWordWL"], 0], ["idx", ["cast", ["idx", "m_addr", 0], "DWordWL"], 0], 0, 0, 0, 0, 0, 0]
+multiplicity = "μ"
+
+[[constraints.memory]]
+kind = "interaction"
+tag = "IS_HALF"
+input = [["idx", ["idx", "m_addr", "i"], "j"]]
+multiplicity = "μ"
+iters = [["i", 0, 6], ["j", 0, 3]]
```
erik-3milabs:
```suggestion
[[constraints.memory]]
kind = "interaction"
tag = "IS_HALF"
input = [["idx", ["idx", "m_addr", "i"], "j"]]
multiplicity = "μ"
iters = [["i", 0, 7], ["j", 0, 3]]
```

note: we don't need to range check `m_addr[0]` since every time it is used, it is cast to a `DWordWL`, so this could also be changed to `[1,7]`.

spec/src/sha256.toml
```diff
@@ -0,0 +1,268 @@
...
+input = [1, ["cast", ["*", 2, 10], "DWordWL"], ["arr", ["idx", ["cast", ["idx", "h_addr", 0], "DWordWL"], 0], ["idx", ["cast", ["idx", "h_addr", 0], "DWordWL"], 1], 0, 0, 0, 0, 0, 0], "timestamp", 1, 0, 0]
+output = ["arr", ["idx", ["cast", ["idx", "h_addr", 0], "DWordWL"], 0], ["idx", ["cast", ["idx", "h_addr", 0], "DWordWL"], 1], 0, 0, 0, 0, 0, 0]
+multiplicity = "μ"
+
+[[constraints.memory]]
+kind = "interaction"
+tag = "IS_HALF"
+input = [["idx", ["idx", "h_addr", "i"], "j"]]
+multiplicity = "μ"
+iters = [["i", 0, 3], ["j", 0, 3]]
```
erik-3milabs, line 114: Technically, we don't need to range check `h_addr[0]` since every time it is used, it is cast to a `DWordWL`, so this could also be changed to `i=[1,3]`. Still, might be nice to keep it for peace of mind.

spec/src/sha256.toml
```diff
@@ -0,0 +1,268 @@
...
+output = ["arr",
+  ["idx", "h", ["+", ["*", "i", 8], 3]],
+  ["idx", "h", ["+", ["*", "i", 8], 2]],
+  ["idx", "h", ["+", ["*", "i", 8], 1]],
+  ["idx", "h", ["+", ["*", "i", 8], 0]],
+  ["idx", "h", ["+", ["*", "i", 8], 7]],
+  ["idx", "h", ["+", ["*", "i", 8], 6]],
+  ["idx", "h", ["+", ["*", "i", 8], 5]],
+  ["idx", "h", ["+", ["*", "i", 8], 4]],
+]
```
erik-3milabs, line 150: By reordering these elements, the rendering shrinks them from `i * 8` to `8i`, improving readability:

```suggestion
[[constraints.memory]]
kind = "interaction"
tag = "MEMW"
input = [
  0,
  ["cast", ["idx", "h_addr", "i"], "DWordWL"],
  ["arr",
      ["idx", "out", ["+", ["*", 8, "i"], 3]],
      ["idx", "out", ["+", ["*", 8, "i"], 2]],
      ["idx", "out", ["+", ["*", 8, "i"], 1]],
      ["idx", "out", ["+", ["*", 8, "i"], 0]],
      ["idx", "out", ["+", ["*", 8, "i"], 7]],
      ["idx", "out", ["+", ["*", 8, "i"], 6]],
      ["idx", "out", ["+", ["*", 8, "i"], 5]],
      ["idx", "out", ["+", ["*", 8, "i"], 4]]],
  ["+", "timestamp", ["cast", 1, "DWordWL"]],
  0, 0, 1
]
output = ["arr",
  ["idx", "h", ["+", ["*", 8, "i"], 3]],
  ["idx", "h", ["+", ["*", 8, "i"], 2]],
  ["idx", "h", ["+", ["*", 8, "i"], 1]],
  ["idx", "h", ["+", ["*", 8, "i"], 0]],
  ["idx", "h", ["+", ["*", 8, "i"], 7]],
  ["idx", "h", ["+", ["*", 8, "i"], 6]],
  ["idx", "h", ["+", ["*", 8, "i"], 5]],
  ["idx", "h", ["+", ["*", 8, "i"], 4]],
]
```

spec/src/sha256.toml
```diff
@@ -0,0 +1,268 @@
...
+input = [["idx", ["idx", "h_addr", "i"], "j"]]
+multiplicity = "μ"
+iters = [["i", 0, 3], ["j", 0, 3]]
+
+[[constraints.memory]]
+kind = "template"
+tag = "ADD"
+input = [["cast", ["idx", "h_addr", 0], "DWordWL"], ["*", 8, ["cast", ["+", "i", 1], "DWordWL"]]]
+output = ["cast", ["idx", "h_addr", "i"], "DWordWL"]
+iter = ["i", 1, 3]
```
erik-3milabs, line 121: this offset should be removed:

```suggestion
[[constraints.memory]]
kind = "template"
tag = "ADD"
input = [["cast", ["idx", "h_addr", 0], "DWordWL"], ["*", 8, ["cast", "i", "DWordWL"]]]
output = ["cast", ["idx", "h_addr", "i"], "DWordWL"]
iter = ["i", 1, 3]
```

spec/sha256.typ
```diff
@@ -0,0 +1,203 @@
...
+#let nr_columns = total_nr_instantiated_columns(sha256chip, config)
+
+The #sha256 chip leverages #nr_variables variables, spanning #nr_columns columns:
+#render_chip_column_table(sha256chip, config)
+
+== Constraints
+
+The first responsibility of the chip is to read the current state and message chunk from memory,
+passed as arguments through pointers.
+Since the memory ranges could overlap, we read the chunk first, before reading and writing the state at the next timestamp.
```
erik-3milabs: I propose we add references to C4 and C8 to clarify which is which.

spec/sha256.typ
erik-3milabs: > what is going on, the constants don't even match!?

To the best of my understanding, ROTXOR is not an all-that-wellknown primitive (or that ROTXOR is a common name for it); I don't expect people to catch the round constants not "matching up" immediately.
And even if they do, I don't think that's a problem: confusion invites the reader to go to the definition and thoroughly read through it.
Third, it is not like ROTXOR is the only interaction that is used before the definition of the corresponding table: SHA256_M, SHA256_K and SHA256ROUND are also defined later (though I will admit that their interface is easier to guess based on the provided parameters). I don't see why ROTXOR deserves a special treatment.

I'm still of the opinion that this definition is a better fit for the ROTXOR subsection.

spec/src/sha256.toml
erik-3milabs: I really like the readability improvement this gives. 

> Hmm, maybe, but I don't love that we then need ["idx", ["cast", ["idx", "m_addr", 0], "DWordWL"], 0] for the register reads

I agree this isn't the prettiest thing I've seen today. Still, I do think the drawback of adding this is outweighed by the benefit of the concise description it allows.

erik-3milabs (changes requested):

spec/sha256.typ
erik-3milabs: I'm OK with my proposal not being accepted, but I do think we need to do something about these two sentences, for three reasons:

1. I believe the sentence

> They do not intend to handle the requisite repeated invocations of this compression,
nor with padding or the preparation of the initial state.

to be grammatically incorrect. In particular, the use of "nor" requires that _"They do not intend ... with padding..."_ be proper English, which I don't believe to be the case.

2. The phrase

> They do not intend

suggests that these chips have a will of their own -- which they do not. I believe the proper phrase would be "They are not intended ..."

3. With the way the sentence is currently written, very little emphasis is placed on the difference between "SHA256" (the hash function) and "the SHA256 compression function" (its primary subroutine). To not confuse the reader, I believe this emphasis should be added. I tried to in the form of adding "_only_" as well as the "'s" at the end of the first sentence, suggesting that what came before (SHA256 compression function) it is one of several components of what comes later (the SHA256 hash function)

Another attempt:

```suggestion
The purpose of this set of chips is to provide an accelerator for SHA256's compression function only;
they are not intended to handle the requisite repeated invocations of this compression,
padding, or the preparation of the initial state.
```

spec/src/sha256.toml
erik-3milabs: leaving this open for posterity

spec/sha256.typ
RobinJadoul:
```suggestion
The purpose of this set of chips is to provide an accelerator for the SHA256 compression function.
They do not handle the requisite repeated invocations of this compression when used as part of a full SHA256 hash function,
nor deal with padding or the preparation of the initial state.
```
maybe?

erik-3milabs: Some more remarks:
1. "this set": "this" usually refers back to something said earlier. That cannot be the case here, as this is the first sentence in the section.
2. "set of chips" is singular. As such, I believe we cannot refer to it as "they" at the start of the second sentence.
3. "The purpose of this set of chips is to provide an accelerator": we introduce these chips to _be_ the accelerator, not provide one.
4. "They do not handle ... when used as part of a full SHA256 hash function, ..."
This sentence is hard to read and has some contextual flaws in my opinion.
5. I find the "not handle ... nor ... or ..." hard to read.

Counter proposal:

```suggestion
The following chips constitute an accelerator for the SHA256 compression function; 
other aspects of SHA256 hashing (such as repeated compression invocation, 
input padding and state initialization) fall outside the scope of this accelerator.
```

erik-3milabs (approved):

PR #375, opened 2026-02-28
SIMD Goldilocks FFT butterflies (NEON + AVX2)

gabrielbosio:
TypeId dispatch in the FFT butterfly helpers (`process_fused_block`, `process_single_layer_block`, `process_ifft_single_layer_block`). When `F=E=GoldilocksField` on a supported arch, casts `&mut [FieldElement<E>]` to `&mut [u64]` via `#[repr(transparent)]` and calls SIMD kernels. The TypeId comparison is a compile-time constant after monomorphization — zero runtime cost for non-Goldilocks fields.

- **NEON (aarch64)**: 2-wide `PackedGoldilocks2` — processes 2 butterflies per iteration
- **AVX2 (x86_64)**: 4-wide `PackedGoldilocks4` — processes 4 butterflies per iteration
- Scalar tail handles remainders when `half_block` is not a multiple of the SIMD width
- Both parallel and sequential FFT paths get SIMD (fallback loops now route through the helper functions)
- `'static` bounds propagated through the FFT call chain (all concrete callers already use `'static` types)

### DIFF
--- a/crypto/math/src/fft/cpu/bowers_fft.rs
+++ b/crypto/math/src/fft/cpu/bowers_fft.rs
@@ -320,20 +320,18 @@ fn process_fused_block<F, E>(
     #[cfg(target_arch = "x86_64")]
     if core::any::TypeId::of::<F>() == core::any::TypeId::of::<GoldilocksField>()
         && core::any::TypeId::of::<E>() == core::any::TypeId::of::<GoldilocksField>()
+        && is_x86_feature_detected!("avx2")
     {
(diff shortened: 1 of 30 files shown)


gabrielbosio:
Running in my M3 Pro with 18 GB of RAM also show no improvements, I will close this PR and look for other bottleneck. Also, it seems that NEON won't show improvements over Apple Silicon so I will go with AVX only in the next PR for now, we can add NEON later to confirm.

PR #378, opened 2026-03-02
Add AVX2 vectorized constraint accumulation

gabrielbosio:
Vectorizes the `sum += eval[i] * beta[i]` accumulation in `evaluate_transitions` using AVX2 across 4 LDE points. Adds `square4`, `inv4`, and cubic extension SIMD ops to `goldilocks_avx2.rs`. Evaluator changed from per-point to `par_chunks_mut(4)` iteration calling `IsField::batch_dot_product_4`.

Depends on #358.


gabrielbosio:
No improvements. We can tackle this later.

PR #382, opened 2026-03-03
Parallelize auxiliary trace building (fingerprints, batch inversion, …

diegokingston:
…prefix sum)

Add row-level parallelism within each LogUp interaction:
- parallel_batch_inverse: chunked Montgomery's trick via par_chunks_mut(1024)
- Fingerprint loop: par_chunks_mut across rows (each row reads immutable main cols)
- Accumulated column: 3-phase parallel prefix sum (local sums → combine → fold offsets)
- Term computation: parallel zip of multiplicities × fingerprint inverses

Following Plonky3's chunked parallelism patterns. Small tables (<1024 rows) fall through to sequential path with no regression.


gabrielbosio:

PR #405, opened 2026-03-10
Refactor/dvrm template alignment

diegokingston:


ColoCarletti (approved):

MauroToscano (changes requested):
If we prefer this way to handle constraints, we can move this app to a shared module as the buses, and apply it to all. If we only apply it in div rem we have an inconsistency on how for example sign is used in this table and others

PR #412, opened 2026-03-11
Feat/logup micro optimizations

diegokingston:


diegokingston:
image
In PR413, we showed that using sparse results in 4.4% improvement

PR #428, opened 2026-03-16
IFFT fusion + flat twiddles

jotabulacios:
This PR introduces two performance-oriented changes to `bowers_fft.rs`: 2-layer fusion in the iFFT path, and a flat contiguous layout for `LayerTwiddles`.

The iFFT is a hot path in the STARK prover, running once per column during LDE, and each butterfly layer performs a full pass over the buffer. By fusing two layers into one pass, intermediates stay in registers/L1 cache instead of round-tripping through memory. By flattening `LayerTwiddles` from `Vec<Vec<_>>` into a single allocation with arithmetic offsets, twiddle access becomes more cache-friendly and avoids pointer chasing.

### DIFF
--- a/crypto/math/src/fft/cpu/bowers_fft.rs
+++ b/crypto/math/src/fft/cpu/bowers_fft.rs
@@ -242,25 +242,22 @@ where
     while layer + 1 < log_n {
         let block_size = n >> layer;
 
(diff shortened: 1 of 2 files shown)


MauroToscano:
Superseeded

PR #435, opened 2026-03-17
spec: Fix CPU sign bit constraints for `word_instr`

RobinJadoul:


spec/src/cpu.toml
```diff
@@ -744,11 +738,10 @@ constraint = "$#`arg1[4:]` = #`rv1[2]` dot (1 - #`word_instr`) + (2^(32) - 1) do
...
 
 [[constraints.ext]]
-kind = "interaction"
-tag = "MSB16"
-input = [["idx", "rv2", 1]]
+kind = "template"
+tag = "SIGN"
+input = [["idx", "rv2", 1], "word_instr"]
 output = "arg2_sign_bit"
-multiplicity = "word_instr"
```
erik-3milabs, line 751: I believe this should have `arg2` rather than `rv2` as input.

The two things suggesting this to me:
1. the name mismatch between the input (`rv2`) and the output (`arg2_sign_bit`)
2. The `SIGN` template assumes `IS_HALF[rv2[1]]`. While CE61 + CR33 does this for _some_ cases, it does not cover all.

erik-3milabs: Having another look at the constraints, I believe that the above is not actually correct. Perhaps renaming `arg2_sign_bit` to `rv2_sign_bit` would be a better solution.

RobinJadoul: Hmm, I think that may even be more generally a problem, that rv2 isn't always properly range checked (and maybe rv1 too)
Though the range check would have to be on arg{1,2} for most uses, it could be dangerous

RobinJadoul: Double checked that the range checks are fine; since rv1 and rv2 are just bridging between range-checked memory and range-checked arg1 and arg2

erik-3milabs (approved):
lgtm

PR #448, opened 2026-03-18
opt: parallelize FRI fold with rayon

MauroToscano:
## Summary

Parallelize `fold_evaluations_in_place` in the FRI commit phase using `rayon::par_iter`. The fold loop was previously sequential over `N/2` extension field elements. Since the in-place fold has aliasing (`evals[j]` reads from `evals[2*j]`), the parallel version uses a temporary buffer + `clone_from_slice`.

## Benchmark (Apple Silicon M3 Pro, `PARALLEL_TABLES=1`, `fib_iterative_1M`)

| Metric | Before | After | Delta |
|--------|--------|-------|-------|
| Prove time (median, 3 samples) | 62.6s | 60.3s | **-3.7%** |
| CV | - | 0.4% | |
| Heap | 23,300 MB | 23,336 MB | +0.2% |
| Verification | - | PASS | |

The temp buffer for the parallel version is ~36MB (N/2 extension elements) — small vs the 23GB working set.

## Test plan
- [x] `cargo test --release -p stark` (all passed)
- [x] Proof verified against baseline verifier binary
- [ ] `/bench` on CI runner


MauroToscano:
Closing this — superseded by #597, which is the newer take on the same FRI fold parallelization and adds a size threshold to avoid Rayon overhead on the small final layers. #597 still needs a more thorough review (rebase onto current main, fmt fix, and a real `/bench` on the runner), but it's the version we'll carry forward rather than this one.

PR #457, opened 2026-03-23
spec/MEMW_R: register access fast path

erik-3milabs:
This PR introduces a 10-column wide fast path for register interactions.

Also defines padding for `MEMW_A`


spec/memw.typ
```diff
@@ -87,9 +88,63 @@ The #aligned chip only needs #nr_variables variables, expressed through #nr_colu
...
+The #reg chip provides a fast-path for accessing registers.
+This fast-path leverages that registers
++ can be addressed using a `Byte`, rather than a full `DWord`,
++ are constantly accessed, i.e., $#`timestamp` - #`old_timestamp`$ is small, and
++ have a fixed access pattern
+to achieve a footprint that is significantly smaller than both #memw and #aligned.
+
+Note: as a result of hard optimization, this chip can only be used for register accesses for which 
++ $#`timestamp` - #`old_timestamp` in [1, 2^16]$, and
++ $#`timestamp` mod 2^ > #`old_timestamp`_0$
```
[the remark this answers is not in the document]
erik-3milabs: fixed

spec/src/memw_register.toml
```diff
@@ -0,0 +1,145 @@
...
+[[constraint_groups]]
+name = "multiplicities"
+
+[[constraints.multiplicities]]
+kind = "template"
+tag = "IS_BIT"
+input = ["μ_read"]
+ref = "regw:c:μ_read_is_bit"
+
+[[constraints.multiplicities]]
```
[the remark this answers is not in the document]
erik-3milabs: fixed

spec/memw.typ
```diff
@@ -87,9 +88,63 @@ The #aligned chip only needs #nr_variables variables, expressed through #nr_colu
 #render_chip_assumptions(alignedchip, config)
 #render_constraint_table(alignedchip, config)
 
+== Padding
+The table can be padded to the next power of two with the following value assignments:
```
erik-3milabs, line 96: @RobinJadoul do you know why MEMW_A (and MEMW, for that matter) don't have padding tables?

RobinJadoul: Nope, no idea :sweat_smile:

spec/src/memw_register.toml
```diff
@@ -0,0 +1,145 @@
...
+name = "old_timestamp_lo"
+type = "Word"
+desc = "the lower limb of `old_timestamp`"
+pad = 0
+
+[[variables.virtual]]
+name = "old_timestamp"
+type = "DWordWL"
+desc = "timestamp at which this register was last accessed"
+def = ["cast", ["arr", "old_timestamp_lo", ["idx", "timestamp", 1]], "DWordWL"]
```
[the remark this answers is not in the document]
RobinJadoul: This is not an assumption, but pure functioning of the chip, since old_timestamp is never surfaced to any caller, so no caller can violate the assumption.
It does matter for "routing", but neither assumption, nor assumption-adjacent comment is the place for that.

spec/src/memw_register.toml
```diff
@@ -0,0 +1,145 @@
...
+
+[[constraint_groups]]
+name = "interactions"
+
+[[constraints.interactions]]
+kind = "interaction"
+tag = "memory"
+input = [1, ["cast", ["arr", ["cast", ["+", ["*", 2, "address"], "i"], "Word"], 0], "DWordWL"], "old_timestamp", ["idx", "old", "i"]]
+iter = ["i", 0, 1]
+multiplicity = "μ_sum"
```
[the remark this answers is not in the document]
RobinJadoul: This is regular MEMW behaviour already, and consistent with the description from the memory chapter.

spec/memw.typ
```diff
@@ -87,9 +88,63 @@ The #aligned chip only needs #nr_variables variables, expressed through #nr_colu
...
+
+The #reg chip provides a fast-path for accessing registers.
+This fast-path leverages that registers
++ can be addressed using a `Byte`, rather than a full `DWord`,
++ are constantly accessed, i.e., $#`timestamp` - #`old_timestamp`$ is small, and
++ have a fixed access pattern
+to achieve a footprint that is significantly smaller than both #memw and #aligned.
+
+Note: as a result of hard optimization, this chip can only be used for register accesses for which 
++ $#`timestamp` - #`old_timestamp` in [1, 2^16]$, and
```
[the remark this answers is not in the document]
RobinJadoul: The condition should remain independent from the construction details like the fixing of timestamp[1], as that is more of an implementation/optimization detail.

spec/src/memw_register.toml
```diff
@@ -0,0 +1,145 @@
...
+desc = "Whether we are performing a write (and hence not return `out`)"
+pad = 0
+
+
+
+# Assumptions
+
+[[assumptions]]
+desc = "`IS_BYTE[address]`"
+ref = "regw:a:address"
```
RobinJadoul: `address` is not directly exposed to any callers (since we have the scaling integration).
There are two sides I'm seeing:
- we do not check this and leave it implicit (not _really_ an assumption in the traditional sense, since the correctness of the chip doesn't depend on it in isolation)
- we check it here to detect and prevent any calling site forgetting to scale; leading to some extra cost on a highly-optimized chip

spec/memw.typ
```diff
@@ -87,9 +88,63 @@ The #aligned chip only needs #nr_variables variables, expressed through #nr_colu
...
+The #reg chip is comprised of #nr_variables variables that are expressed using #nr_columns columns:
+#render_chip_column_table(register_chip, config)
+
+== Assumptions
+The following range checks are assumed to be performed/enforced outside of this chip:
+#render_chip_assumptions(register_chip, config)
 
+== Constraints
+Since most registers are frequently accessed, the difference between `timestamp` and `old_timestamp` is small most of the times.
+Rather than storing their (nearly) identical upperlimbs twice, it is instead assumed that
```
RobinJadoul:
```suggestion
Rather than storing their (nearly) identical upper limbs twice, it is instead assumed that
```

spec/src/memw_register.toml
```diff
@@ -0,0 +1,145 @@
...
+ref = "regw:c:write_val"
+
+
+[[constraint_groups]]
+name = "output"
+
+[[constraints.output]]
+kind = "interaction"
+tag = "MEMW"
+input = [1, ["cast", ["arr", ["cast", "address", "Word"], 0], "DWordWL"], ["arr", ["idx", "val", 0], ["idx", "val", 1], 0, 0, 0, 0, 0, 0], "timestamp", 1, 0, 0]
```
RobinJadoul: To check correspondence with how we want to enforce byte constraints and scaling on `address`. If we don't scale here, the Byte assumption is wrong, and we have some danger of overlapping memory accesses (which _should_ mess with timestamps too, but still a little tricky).
I think I prefer scaling explicitly here (though depending on the Byte assumption question, that may or may not actually do much)

spec/src/memw_register.toml
RobinJadoul, line 39: Technically, this may not really need a cast, as the array and DWordWL are structurally the same

spec/src/memw_register.toml
```diff
@@ -0,0 +1,145 @@
...
+ref = "regw:a:timestamp"
+
+# Constraints
+
+[[constraint_groups]]
+name = "diff"
+
+[[constraints.diff]]
+kind = "interaction"
+tag = "IS_HALF"
```
RobinJadoul, line 80: We also have an IS_B20 table (or ZERO) to expand our range a bit more

spec/src/memw_register.toml
```diff
@@ -0,0 +1,145 @@
...
+input = ["μ_sum"]
+ref = "regw:c:μ_sum_is_bit"
+
+[[constraint_groups]]
+name = "interactions"
+
+[[constraints.interactions]]
+kind = "interaction"
+tag = "memory"
+input = [1, ["cast", ["arr", ["cast", ["+", ["*", 2, "address"], "i"], "Word"], 0], "DWordWL"], "old_timestamp", ["idx", "old", "i"]]
```
RobinJadoul: Here, and the next, no hard need to cast an array to DWordWL

RobinJadoul (commented):
Short and sweet 👍

spec/memw.typ
```diff
@@ -6,7 +6,11 @@
   total_nr_variables,
   total_nr_instantiated_columns,
   render_constraint_table,
+<<<<<<< spec/MEMW_R
   render_chip_padding_table
+=======
+  render_chip_padding_table,
+>>>>>>> spec/memw_update2
```
RobinJadoul: Whoops, conflict markers

spec/memw.typ
erik-3milabs: this was fixed in #459

spec/src/memw_register.toml
erik-3milabs: We settled on using 2*address in all interactions and just listing the assumption (i.e., not turning it into an actual check); we find this an acceptable balance between having a clear spec and a fully optimized chip

spec/src/memw_register.toml
erik-3milabs: I removed them

spec/src/memw_register.toml
erik-3milabs: I added this as a future optimization. With this chip being on the fast path & there being discussions of splitting the BYTE_OPS tables, I can imagine it being favourable for this to use the 2^16 lookup rather than a 2^20

spec/src/memw_register.toml
erik-3milabs: I removed it

spec/src/memw_register.toml
RobinJadoul: I would still tend to remove the assumption, or reword it to say `address ∈ 2 dot [0, 255]` or the like.
Currently, it's a bit confusing to have an IS_BYTE assumption and a Half column definition

RobinJadoul: The weird thing about having an assumption about `address` is that it doesn't come directly from the "incoming" interaction, due to the internal scaling; so maybe just removing is cleanest (and keeping the column def as Byte for semantics)

RobinJadoul (approved):
Just the last cleanup on `address` assumptions and column definition, but looking good

PR #469, opened 2026-03-26
spec: interaction counter

erik-3milabs:
This PR does three things:
- Add support for computing the number of interactions performed by a chip (including those issued by used templates).
- Print number of interactions for all (relevant) existing chips
- some cleanup

### DIFF
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -44,7 +44,16 @@
   .product(default: 1)
 }
 
-#let set_nr_interactions(chip) = {
+// Compute the number of interactions performed by `chip` and
+// store it as metadata under the `<interaction_count>` label
+// with tag `chip.name`. This tag is overwritten by `name` when specified.
+#let set_nr_interactions(chip, name: none) = {
+  let chip-name = if name != none {
+      name 
+    } else {
+      chip.name    
+    }
+
   let constraints = chip
     .constraints
     .values()
@@ -73,7 +82,9 @@
 
     let total-nr-interactions = nr-direct-interactions + nr-indirect-interactions
 
-    [#metadata((chip.name: total-nr-interactions)) <interaction_count>]
+    let entry = (:)
+    entry.insert(chip-name, total-nr-interactions)
+    [#metadata(entry) <interaction_count>]
   }
 }
 
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -25,7 +25,7 @@
 }
 
 // Given a constraint, compute the number of interactions it induces
-#let get_interaction_count(constraint) = {
+#let get_constraint_interaction_count(constraint) = {
   let iters = if "iters" in constraint {
     constraint.iters
   } else if "iter" in constraint {
@@ -37,7 +37,7 @@
   iters.map(i => {
     assert(
       i.len() == 3 and type(i.at(1)) == int and type(i.at(2)) == int,
-      message: "contains invalid iter: " + repr(i),
+      message: "invalid iter: " + repr(i),
     )
     i.at(2) - i.at(1) + 1
   })
@@ -48,10 +48,8 @@
 // store it as metadata under the `<interaction_count>` label
 // with tag `chip.name`. This tag is overwritten by `name` when specified.
 #let set_nr_interactions(chip, name: none) = {
-  let chip-name = if name != none {
-      name 
-    } else {
-      chip.name    
+  if name == none {
+      name = chip.name
     }
 
   let constraints = chip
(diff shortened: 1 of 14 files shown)


spec/chip.typ
```diff
@@ -24,6 +24,49 @@
...
+
+    [#metadata((chip.name: total-nr-interactions)) <interaction_count>]
+  }
+}
+
+#let get_nr_interactions(chip) = {
+  context {
+    query(<interaction_count>).map(c=>c.value).sum().at(chip.name)
+  }
+}
```
RobinJadoul: I would opt to just call `set_nr_interactions` from here?
That makes it less repetitive and easy-to-forget for the individual chapters

spec/chip.typ
```diff
@@ -24,6 +24,49 @@
...
+    .constraints
+    .values()
+    .flatten()
+
+  // nr. of direct interactions
+  let nr-direct-interactions = constraints
+    .filter(c=>c.kind == "interaction")
+    .map(c => {
+      let x = c.at("iter", default: (0, 0, 0))
+      x.at(2) - x.at(1) + 1
```
RobinJadoul: This doesn't deal with multiple iters (and the same for the templates)

spec/chip.typ
```diff
@@ -24,6 +24,49 @@
...
+  context {
+    let lookup-table = query(<interaction_count>).map(x => x.value).sum(default:(:))
+
+    // nr. of indirect interactions through templates
+    let nr-indirect-interactions = template-constraints
+      .map(c => {
+        let x = c.at("iter", default: (0, 0, 0))
+        let iter-size = x.at(2) - x.at(1) + 1
+
+        let template-interactions = lookup-table.at(c.tag, default: 0)
```
RobinJadoul: Should we really have a default here? If the tag doesn't exist or didn't have its number of interactions rendered anywhere, that's worthy of an error (or at least a warning), imo

spec/chip.typ
```diff
@@ -24,6 +24,49 @@
...
+        iter-size * template-interactions 
+      })
+      .sum(default: 0)
+
+    let total-nr-interactions = nr-direct-interactions + nr-indirect-interactions
+
+    [#metadata((chip.name: total-nr-interactions)) <interaction_count>]
```
RobinJadoul: An alternative to having a bunch of metadata elements with the same label is to dynamically generate the labels, based on the template/chip name

spec/is_bit.typ
```diff
@@ -1,12 +1,15 @@
...
-#import "/chip.typ": render_chip_column_table, render_constraint_table
+#import "/chip.typ": render_chip_column_table, render_constraint_table, set_nr_interactions, get_nr_interactions,
 
 #let config = load_config()
 #let chip = load_chip("src/is_bit.toml", config)
 
 #show: book-page(chip.name)
 
+#set_nr_interactions(chip)
+#let nr_interactions = get_nr_interactions(chip)
```
RobinJadoul: Unused (intentionally?)

spec/chip.typ
erik-3milabs: The problem we're facing here is that `set_nr_interactions` returns a `metadata` object that needs to hit the main document for the `get_*` to recall again. I have not found a way of merging `set_*` into `get_*` that does both does this placement and is less cumbersome that the system in place right now.

Do you have an idea on how to better approach this?

spec/chip.typ
erik-3milabs: What is the upside of taking that approach?

spec/chip.typ
erik-3milabs: no, good point

spec/chip.typ
```diff
@@ -24,6 +24,49 @@
...
+    .values()
+    .flatten()
+
+  // nr. of direct interactions
+  let nr-direct-interactions = constraints
+    .filter(c=>c.kind == "interaction")
+    .map(c => {
+      let x = c.at("iter", default: (0, 0, 0))
+      x.at(2) - x.at(1) + 1
+    })
```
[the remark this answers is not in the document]
erik-3milabs: removed

spec/chip.typ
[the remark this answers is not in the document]
erik-3milabs: removed

spec/chip.typ
```diff
@@ -24,6 +24,49 @@
...
+
+    let total-nr-interactions = nr-direct-interactions + nr-indirect-interactions
+
+    [#metadata((chip.name: total-nr-interactions)) <interaction_count>]
+  }
+}
+
+#let get_nr_interactions(chip) = {
+  context {
+    query(<interaction_count>).map(c=>c.value).sum().at(chip.name)
```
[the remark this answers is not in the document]
erik-3milabs: removed

spec/chip.typ
RobinJadoul: You never call `get_nr_interactions` somewhere without first calling the corresponding setting, so something like this should be just fine, right?
```suggestion
#let compute_nr_interactions(chip) = {
  set_nr_interactions(chip)
  context {
    query(<interaction_count>).map(c=>c.value).sum().at(chip.name)
  }
}
```

spec/chip.typ
RobinJadoul: Nothing much, other than a minor dislike for labels/ids that aren't a unique element; but it was mostly just an idle thought while reviewing

RobinJadoul (approved):

spec/chip.typ
erik-3milabs: Actually, yeah, that works... Weird! I thought that because `set_nr_interactions` is not the last expression in `compute_nr_interactions`, its return-value (which is the metadata tag) would not be propagated to the top of the calling tree, making the metadata-tag unusable. Do you see the flaw in my reasoning?

RobinJadoul: The multiple statements get concatenated as content, I believe, so both are in the output
(you can look at the output of some `repr(compute_nr_interactions(...))` and I expect that to be a sequence of two context elements or smth like that)

erik-3milabs: I see: `#let nr_interactions = compute_interactions(chip)` means `nr_interactions` is a sequence of the metadata tag and the final count (both in context). This approach now relies on the assumption that `nr_interactions` is actually included in the text; otherwise the metadata tag doesn't actually hit the document and the system breaks down when the computation of other chips are dependent on it.

:+1:

PR #474, opened 2026-04-01
spec: KECCAK accelerator

erik-3milabs:
This PR adds:
1. specification for a keccak accelerator
2. spec support for modulus notation with a constant modulus
3. spec support for multi-dimensional arrays


spec/src/keccak.toml
```diff
@@ -0,0 +1,102 @@
...
+multiplicity = "μ"
+
+[[constraints.mem]]
+kind = "interaction"
+tag = "MEMW"
+input = [0, ["cast", ["idx", ["idx", "state_ptr", "x"], "y"], "DWordWL"], ["idx", ["idx", "output_state", "x"], "y"], "timestamp", 0, 0, 1]
+output = ["idx", ["idx", "input_state", "x"], "y"]
+iters = [["x", 0, 4], ["y", 0, 4]]
+multiplicity = "μ"
+ref = "keccack:c:load_store_state"
```
RobinJadoul:
```suggestion
ref = "keccak:c:load_store_state"
```

spec/expr.typ
```diff
@@ -97,6 +104,13 @@
     "not": (pp, rec, e) => cwrap(rec(PREC.not, 1) + ` - ` + rec(PREC.not, e.at(1)), pp < PREC.not),
     "+": (pp, rec, e) => cwrap(e.slice(1).map(rec.with(PREC.add)).join(` + `), pp < PREC.add),
     "sum": (pp, rec, e) => assert(false, message: "sum is unsupported in code."),
+    "mod": (pp, rec, e) => {
+      assert(e.len() == 3 and type(e.at(2)) == int, message: "Invalid mod expr: " + repr(e))
+      cwrap(
+        rec(PREC.mod, e.at(1)) + `%` + rec(PREC.mod, e.at(2)),
```
RobinJadoul: Nit: probably looks nicer to have spaces around the `%` like for other operators

spec/signatures.typ
```diff
@@ -6,6 +6,16 @@
 #let config = load_config()
 #let signatures = load_signatures(config)
 
+// Render a type
```
RobinJadoul: We should probably reuse `type_to_code` here

spec/src/keccak.toml
```diff
@@ -0,0 +1,102 @@
...
+name = "state_ptr"
+type = [["DWordHL", 5], 5]
+desc = "memory addresses storing the entire state"
+pad = ["*", 8, ["arr", 
+        ["arr", 0, 1, 2, 3, 4],
+        ["arr", 5, 6, 7, 8, 9],
+        ["arr", 10, 11, 12, 13, 14],
+        ["arr", 15, 16, 17, 18, 19],
+        ["arr", 20, 21, 22, 23, 24]
+]]
```
RobinJadoul, line 37: Possible optimization: don't repeat `addr` in `state_ptr`
Gains: 4 columns and `IS_HALF` checks
Loses: sanity

spec/keccak.typ
```diff
@@ -0,0 +1,108 @@
...
+`iota` contains the updated lane.
+In the definition of `out`, the output of `chi` and `iota` is combined to construct the output of the permutation.
+#render_constraint_table(round_chip, config, groups: "iota")
+
+Lastly, the round chip contributes the following interactions to the lookup:
+#render_constraint_table(round_chip, config, groups: "io")
+
+== Optimizations
+- step $rho$ does not need to be applied to `state[0][0]`; its has a zero-shift. This saves 16 columns and 4 `HWSL` interactions.
+- $#`rc[2]` = #`rc[4]` = #`rc[5]` = #`rc[6]` = 0$. As such, those elements need not be stored in `rc`, and need not be XORed into the state in the $iota$-step. This saves 8 columns and 4 `XOR_BYTE` interactions.
```
RobinJadoul: To clarify: these are potential, but not implemented optimizations?

spec/src/keccak_round.toml
```diff
@@ -0,0 +1,299 @@
...
+
+[[variables.input]]
+name = "start"
+type = [[["Byte", 8], 5], 5]
+desc = "state at the start of executing the permutation"
+
+[[variables.auxiliary]]
+name = "Cxz"
+type = [[["Byte", 8], 4], 5]
+desc = "$xor_(i=0)^(y+2) #`start[x,i,z]`$"
```
RobinJadoul, line 22: To investigate: Would adding a 3-way (size 2^24) XOR precomputed table be a worthwhile addition to shrink a bunch of these sort things? (e.g. also helpful the the SHA256 ROTXOR)

spec/src/keccak.toml
```diff
@@ -0,0 +1,102 @@
...
+multiplicity = "μ"
+
+[[constraints.mem]]
+kind = "interaction"
+tag = "MEMW"
+input = [0, ["cast", ["idx", ["idx", "state_ptr", "x"], "y"], "DWordWL"], ["idx", ["idx", "output_state", "x"], "y"], "timestamp", 0, 0, 1]
+output = ["idx", ["idx", "input_state", "x"], "y"]
+iters = [["x", 0, 4], ["y", 0, 4]]
+multiplicity = "μ"
+ref = "keccak:c:load_store_state"
```
RobinJadoul, line 87: If I'm reading algorithm 10 (B.1) correctly, bytes are actually stored little-endian?

spec/src/keccak_round.toml
```diff
@@ -0,0 +1,299 @@
...
+
+[[variables.auxiliary]]
+name = "Cxz_right"
+type = [["Byte", 8], 5]
+desc = "the right-rotated component of `rotated_Cxz`"
+
+[[variables.auxiliary]]
+name = "Dxz"
+type = [["Byte", 8], 5]
+desc = "$#`Cxz[x,y,z]` xor #`rotated_Cxz[`\\(#`x` +1) mod 5#`,y,z]`$"
```
RobinJadoul:
```suggestion
desc = "$#`Cxz[`\\(#`x` - 1) mod 5#`,y,z]` xor #`rotated_Cxz[`\\(#`x` + 1) mod 5#`,y,z]`$"
```

spec/keccak.typ
```diff
@@ -0,0 +1,108 @@
...
+== Columns
+#let nr_variables = total_nr_variables(round_chip)
+#let nr_columns = total_nr_instantiated_columns(round_chip, config)
+
+The #keccak_rnd chip is comprised of #nr_variables variables that are expressed using #nr_columns columns:
+#render_chip_variable_table(round_chip, config)
+
+
+#strong("Note on " + raw("rnc") + " and " + raw("rbc") + ".")
+Rho rotates every lane by a rotation offset in $[0, 64)$.
```
RobinJadoul, line 76: > Equivalently, for each bit in the lane, the z coordinate is modified by adding the offset, modulo the lane size

So if the `(z - 1) mod w` from θ's `C` is a left rotation, `rho` should be a right rotation and hence have modified constants, right?

spec/tooling/chip.py
```diff
@@ -255,6 +255,30 @@ def typecheck(self, env: Environment) -> Type:
...
+    elt: Expr
+    modulus: Expr
+
+    def typecheck(self, env: Environment) -> Type:
+        elt = self.elt.typecheck(env)
+        modulus = self.modulus.typecheck(env)
+
+        if isinstance(modulus, list) or not modulus.is_const():
+            reporter.error(f"Invalid non-constant modulus: {self.modulus!r}")
+            return Range.const(1)
```
RobinJadoul: Any reason to default to `1` instead of `0`?

RobinJadoul (commented):
Some comments and questions.
Confidence in several of them is fairly low, so feel free to tell me if I'm wrong somewhere :)

spec/keccak.typ
erik-3milabs: potential, indeed. I've modified the header to clarify this

spec/src/keccak.toml
erik-3milabs: Given the loss of sanity, I don't think we should incorporate this. I will add it to the "potential optimizations" section for future consideration.

spec/src/keccak.toml
erik-3milabs: Ok, let's elaborate why this chip deals with endianness correctly:

- Let `F = [0xA3, 0x2E, ...]` be a sequence of bytes that we wish to hash
- According to B.1) `S[i] = F[i//8][i mod 8]` 
- According to sec 3.1.2: `A[x, y, z] = S[64*(5y + x) + z]`

Putting this together, we get that `A[x, y, z]` (the input state) should equal `F[8 * (5y + x) + z//8][z mod 8]`

Suppose now that the first byte of `F` is stored at `addr`. Looking at `input_state`, we see that:
```
input_state[x][y][z]: Byte[5][5][8] 
     = MEMW[state_ptr[x][y]][z] (C5)
     = MEMW[addr + 8*(5x+y)][z] (C3)
     = RAM[addr + 8*(5x+y) + z] (MEMW-CM10/12/14/16)
``` 
(there is actually a bug here!: C3 has `x` and `y` swapped! Note that this is now fixed. Let's ignore this for the remainder of this discussion)

Hence, we find that `input_state[x][y][z//8][z mod 8] = F[8*(5y + x) + z//8][z mod 8] = A[x, y, z]`, which is what we expected.

When we represent this with a diagram, we get:
Image
(where 
1. the content should be in the middle layer,  but that is impossible to draw, and
2. the lanes in this diagram are only 8 bits rather than 64 bits long, but I hope you get the gist.
)

With this in place, we can now see that for
1. theta: `C[x+1, z-1]` should be computed as `C[x+1, *] <<< 1`
2. rho: should be a left-rotation,
3. pi: does not care about endianness
4. chi: does not care about endianness, and
5. iota: the round constants just need to be hard-coded correctly.

Do you agree?

spec/src/keccak_round.toml
erik-3milabs: IIUC, your best bet would be to introduce a four-column-2^24-row table akin to BITWISE that handles this lookup. It would allow you to drop the number of lookups performed in C1 and C2 from 160 to 80. 

If a lookup is ~equal in cost to one column, this would start to make sense once you exceed ~2^15 keccak permutation calls. As such, I don't think it is worth it by itself. Perhaps ROTXOR might make it a better idea. Will write the idea down under the potential optimizations for now.

spec/tooling/chip.py
erik-3milabs: None that I can think of anymore. Updated it to be `0` (same holds for the one on line 273)

spec/keccak.typ
erik-3milabs: why would `rho` have to a right-rotation? I believe both rotations are in the same direction.

spec/src/keccak_round.toml
RobinJadoul: It could technically replace the entire XOR part of bitwise (since a xor b = a xor b xor 0), but I wasn't entirely convinced either yet

spec/keccak.typ
RobinJadoul: My confusion stems from the following seeming inconsistency:
- θ introduces a left rotation (by one)
- θ is described by _subtracting_ (one) from the `z` index
- ρ introduces a left rotation (by constant)
- ρ is described as _adding_ (constant) to the `z` index

spec/src/keccak_round.toml
RobinJadoul: It does also reduce the number of columns needed for Cxz, right?

spec/keccak.typ
RobinJadoul: Maybe this is the difference between "adding to the source" and "subtracting from the destination"?

spec/src/keccak.toml
RobinJadoul: I definitely agree the order of the bytes is fine.
And I suppose the intra-byte ordering is mostly just a matter of having the right interpretation, and "reading" from LSB to MSB, which I think indeed works

spec/src/keccak_round.toml
erik-3milabs: > It could technically replace the entire XOR part of bitwise (since a xor b = a xor b xor 0), but I wasn't entirely convinced either yet

Agreed. Yet, doing so would introduce the overhead of always having to prove a 2^24, even when you're not using keccak.



> It does also reduce the number of columns needed for Cxz, right?

You're right; it would induce the removal of 80 columns. I'll add it to the potential optimization section.

spec/keccak.typ
erik-3milabs: > Maybe this is the difference between "adding to the source" and "subtracting from the destination"?

yup, that's it!

RobinJadoul (approved):

PR #476, opened 2026-04-06
Reduce time regression caused by disk spilling

gabrielbosio:
Continues #466 work but reduces time regression while keeping heap peak so that we can still prove fib_iterative_64M on a 128GB RAM machine.


nicole-graus (approved):

gabrielbosio:
#466 now leaves LDE cached on RAM instead of disk. This makes this PR cause no impact on prove time nor heap peak. Closing.

PR #478, opened 2026-04-07
spec: update TOC

erik-3milabs:
This PR reformats the table of contents for the specification.

@RobinJadoul could you check to see whether these chapters can also be integrated into shiroa?


erik-3milabs (commented):
lgtm

RobinJadoul (approved):

PR #480, opened 2026-04-08
Feat/isolated pool parallelism

diegokingston:


gabrielbosio:
It seems this is affecting Rounds 2-4. We could also try applying this same technique for Round 1.

gabrielbosio:
> It seems this is affecting Rounds 2-4. We could also try applying this same technique for Round 1.

Just tested this, no changes.

PR #486, opened 2026-04-09
Disk spill with scaling improvements

gabrielbosio:
A merge of #476 with the first 5 commits of #483 to check how heap peak gets reduced.


jotabulacios (approved):

gabrielbosio:
Since #497 exists, I think it doesn't make sense to keep this PR open.

PR #487, opened 2026-04-09
Perf/batch byte range checks

jotabulacios:
This PR batches CPU byte range checks by unifying the IS_BYTE lookup to verify two bytes per interaction instead of one. Adjacent byte pairs in the CPU trace (e.g., ARG1[0] with ARG1[1]) are now checked together in a single bus interaction, reducing CPU bus interactions from 68 to 56 (-18%). The BITWISE chip already holds every (x, y) byte pair as part of its table, so the new lookup reuses existing rows without widening the table


diegokingston:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Reviewed: byte pairing logic correctness, BusId renumbering consistency, bus interaction value counts, soundness of dual-byte IS_BYTE[X, Y] lookups, trace builder padding alignment, and PAGE/BRANCH single-byte (Y=0) convention.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

diegokingston (approved):

diegokingston:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

**Notes (below threshold, informational only):**

- The IS_BYTE soundness concern raised in the existing comment is a false positive for this PR. IS_BYTE sends X and Y as **independent** `BusValue::Packed { packing: Direct }` entries, producing separate alpha-weighted fingerprint terms. The BITWISE table only has rows where X,Y in [0,255], so both bytes are individually constrained. The IS_HALF packing confusion (lo+256*hi) does not apply here.

- PR description says "68 to 56 (-18%)" but the code and test correctly assert 55 interactions. Minor text discrepancy.

- `single_byte` doc comment (bitwise.rs:507) still lists IS_BYTE as a "single-byte op" example. This is technically still true (you CAN use `single_byte(IsByte, x)` with y=0), but could be clarified since IS_BYTE now primarily operates in two-byte mode via `byte_op()`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

diegokingston (approved):

nicole-graus (approved):

PR #489, opened 2026-04-10
GKR LogUp with disk spill

gabrielbosio:


gabrielbosio:
This PR was meant to be a test on how both GKR and disk spill will affect heap peak and prove time. I will close this since we already posted benchmark results in this PR.

PR #492, opened 2026-04-13
Cache LDE from Round 1 to skip recomputation in Rounds 2-4

gabrielbosio:


diegokingston (approved):

MauroToscano (changes requested):
Pool shouldn't be needed now, since it's being emptied all the time. We can remove all the complexity of that code and only use the cache

gabrielbosio:
> Pool shouldn't be needed now, since it's being emptied all the time. We can remove all the complexity of that code and only use the cache

Done in #498

MauroToscano (approved):

PR #494, opened 2026-04-13
Cache LDE + Scaling improvements

gabrielbosio:
Merge of #492 with the first 5 commits of #483. Just to check bench numbers.


gabrielbosio:
Not a great difference

PR #497, opened 2026-04-14
Uniform table cap + Twiddle dedup + FRI fold parallelism

gabrielbosio:


gabrielbosio:
Capping 2^20 is not showing a difference against capping 2^19.

gabrielbosio:
#499 showed the same improvement and it's smaller.

PR #501, opened 2026-04-16
spec: Inline PC memory access into CPU

RobinJadoul:


spec/src/cpu.toml
```diff
@@ -691,12 +703,39 @@ tag = "MEMW"
...
+
+[[constraints.mem]]
+kind = "template"
+tag = "IS_BIT"
+input = ["prev_pc_timestamp_borrow"]
+
+[[constraints.mem]]
+kind = "arith"
+constraint = "$#`pc_double_read` => #`prev_pc_timestamp_borrow` = 0$"
+poly = ["*", ["-", 1, "pc_double_read"], "prev_pc_timestamp_borrow"]
```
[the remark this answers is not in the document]
RobinJadoul:
```suggestion
constraint = "$#`pc_double_read` => #`prev_pc_timestamp_borrow` = 0$"
poly = ["*", "pc_double_read", "prev_pc_timestamp_borrow"]
```

spec/src/cpu.toml
```diff
@@ -221,6 +221,18 @@ desc = "The value to (maybe) be written back to rvd"
...
+[[variables.auxiliary]]
+name = "prev_pc_timestamp_borrow"
+type = "Bit"
+desc = "The borrow bit for computing the previous timestamp the PC was accessed"
+pad = 0
+
+[[variables.auxiliary]]
+name = "pc_double_read"
+type = "Bit"
+desc = "Whether the PC is being read as a general purpose register (rs1) this cycle"
```
```suggestion
desc = "Whether the PC is being read as a general purpose register (`rs1`) this cycle"
```

spec/cpu.typ
```diff
@@ -60,9 +60,12 @@ The ALU functionality is then obtained through judicious dispatching to the corr
 
 The interactions with the memory, both for register loading and storing, as for `LOAD` and `STORE` instructions are handled.
 Note that since registers need no byte-addressing, we store them in the memory argument with `Word` limbs.
+The `pc` register behaves very predictably with respect to its timestamps and when it is being read,
+so for performance reasons, we inline its memory interactions directly into the #cpu chip.
 The timestamps are ensured to be disjoint for disjoint memory locations.
 One consequence of that is that `next_pc` is written at `timestamp + 1`
 to ensure the access is disjoint with the `pc` read into `rv1` as part of the `AUIPC` instruction.
+#rj[We can try to do a bit more, and move `double_pc_read` into decoding, so that `AUIPC` could set `read_register1 = 0`.]
```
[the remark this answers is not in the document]
erik-3milabs: The start of the message ("we can...") suggests that this message is _optional_. As such, I would propose to
1. implement this (if we know for sure that this will optimize things),
2. list it as a potential optimizations (if we don't know everything we need to determine this), or
3. drop the comment (if we know it is not actually an optimization)

spec/src/cpu.toml
```diff
@@ -691,12 +703,39 @@ tag = "MEMW"
 input = [0, ["cast", "res", "DWordWL"], ["cast", "arg2", ["Byte", 8]], ["+", "timestamp", ["cast", 1, "DWordWL"]], "memory_2bytes", "memory_4bytes", "memory_8bytes"]
 multiplicity = "STORE"
 
+[[constraints.mem]]
+kind = "arith"
+constraint = "$#`pc_double_read` => #`rs1` = 255$"
+poly = ["*", "pc_double_read", ["-", "rs1", 255]]
```
erik-3milabs: IIUC, `pc_double_read` should be set iff `rs1 = 255`. This constraint enforces only one direction. Am I correct in understanding that you're suggesting the opposite direction to follow from the lookup? i.e., if `rs1=255` but `pc_double_read=0`, there would be two reads of register 255 with `old_timestamp = timestamp-3` and thus the logup won't balance.

If so, please provide this reasoning as a comment in the cpu.typ file.

erik-3milabs (approved):
some minor stuff. Aside from that, lgtm

erik-3milabs:
[PR505](https://github.com/yetanotherco/lambda_vm/pull/505) already contains an attempt to implement this spec change.

spec/cpu.typ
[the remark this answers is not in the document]
RobinJadoul: I don't expect the impact to be huge, compared to the potential extra complexity.
But it's something I mostly wanted to keep in mind for when working on a followup PR that does bigger restructuring of the CPU, as that could affect how impactful or complex it is.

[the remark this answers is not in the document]
erik-3milabs: I'd post it under a "Notes/potential optimizations" subsection to maintain consistency

spec/cpu.typ
```diff
@@ -65,7 +65,11 @@ so for performance reasons, we inline its memory interactions directly into the
 The timestamps are ensured to be disjoint for disjoint memory locations.
 One consequence of that is that `next_pc` is written at `timestamp + 1`
 to ensure the access is disjoint with the `pc` read into `rv1` as part of the `AUIPC` instruction.
-#rj[We can try to do a bit more, and move `double_pc_read` into decoding, so that `AUIPC` could set `read_register1 = 0`.]
+Constraints on `pc_double_read` corresponding to an `AUIPC` instruction are not necessary,
+as regardless of its value, the old timestamp is guaranteed smaller than the new timestamp,
+and the integrity of the memory argument therefor ensures the correctness of this bit.
+
+#bold[Potential optimization: ] `double_pc_read` could be integrated into decoding, so that `AUIPC` could set `read_register1 = 0` and no extra MEMW access for `rv1` is needed at this point.
```
erik-3milabs: unknown variable: "bold"

RobinJadoul: Woopsie

spec/cpu.typ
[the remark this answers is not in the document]
RobinJadoul: Fair enough, I still prefer keeping it local to the `Memory` subsection (so maybe a level 3 heading) because the CPU is big enough that the locality would help in seeing exactly how/why it could be good.

spec/cpu.typ
```diff
@@ -60,12 +60,21 @@ The ALU functionality is then obtained through judicious dispatching to the corr
 
 The interactions with the memory, both for register loading and storing, as for `LOAD` and `STORE` instructions are handled.
 Note that since registers need no byte-addressing, we store them in the memory argument with `Word` limbs.
+The `pc` register behaves very predictably with respect to its timestamps and when it is being read,
+so for performance reasons, we inline its memory interactions directly into the #cpu chip.
```
erik-3milabs, line 64: I'd start a new paragraph here:
```suggestion
The `pc` register behaves very predictably with respect to its timestamps and when it is being read,
so for performance reasons, we inline its memory interactions directly into the #cpu chip.

```

spec/cpu.typ
```diff
@@ -60,12 +60,21 @@ The ALU functionality is then obtained through judicious dispatching to the corr
...
 The timestamps are ensured to be disjoint for disjoint memory locations.
 One consequence of that is that `next_pc` is written at `timestamp + 1`
 to ensure the access is disjoint with the `pc` read into `rv1` as part of the `AUIPC` instruction.
+Constraints on `pc_double_read` corresponding to an `AUIPC` instruction are not necessary,
+as regardless of its value, the old timestamp is guaranteed smaller than the new timestamp,
+and the integrity of the memory argument therefor ensures the correctness of this bit.
```
erik-3milabs: 1. This first sentence is weird. What do you think of writing

    > Disjoint memory locations are ensured to be accessed at disjoint timestamps.

2. I'd add a paragraph break before talking about `pc_double_read`, as this is a different topic.

3. I'd tag CM47 & sec 10.3.1 when talking about reading `pc` into `rv1`

4. > "constraints on pc_double_read corresponding to AUIPC" 

    how can a constraint correspond to an instruction? did you perhaps mean 
    > Constraining that `pc_double_read` corresponds to an `AUIPC` instruction
 
5. I believe is should be "therefore" rather than "therefor" (https://englishstudyonline.org/therefore-vs-therefor/)

RobinJadoul: > how can a constraint correspond to an instruction? did you perhaps mean

The intended parse was "constraints on (`pc_double_read` corresponding to an `AUIPC` instruction) are not"

erik-3milabs (changes requested):
.

erik-3milabs (approved):

PR #505, opened 2026-04-17
Feat/inline pc

diegokingston:


diegokingston:
Rebased onto current main and dropped the REGISTER-elimination commits — they had a bus-balance bug that caused 3 verification-regression tests to fail on CI. The branch now contains only the inline-PC core (4 commits) that matches the state of the last successful CI run on this PR (sha c99dd887).

Please run `/bench 3` to confirm the -11.3% prove time / -5.9% heap win from the April 17 bench is preserved.

ColoCarletti (approved):

jotabulacios (approved):

MauroToscano (approved):

PR #513, opened 2026-04-20
fix(ci): add rust program cache in ci

ColoCarletti:


gabrielbosio (changes requested):
I would replace the `.github/actions/setup-rust/action.yaml` with `Makefile` to force recompiling Rust programs when the Rust version set in the Makefile changes.

gabrielbosio (approved):

MauroToscano (approved):

PR #518, opened 2026-04-21
logup parallelism and twiddle deduplication

diegokingston:


MauroToscano (changes requested):
@nicole-graus is splitting it and checking the optimization one by one

diegokingston:
Pushed fixes for the two concrete findings to `triage/pr518-fixes` — two focused commits on top of the PR head. Cherry-pick or merge as you prefer.

**(H) `coset_offset` in domain cache key** — `c7f6aaef`
Adds `coset_offset` (u64) to the HashMap key tuple so that AIRs with same `(trace_length, blowup)` but different offsets don't silently alias onto the same `Domain`/`LdeTwiddles`. Zero runtime cost, 12-line diff in `prover.rs`. Current call sites thread one ProofOptions everywhere, so this is purely defensive — but the cache is in a generic crate and other callers shouldn't need to keep the invariant in their head.

**(M) eliminate 2N allocations in `build_accumulated_column_from_terms`** — `04602bb3`
Restructures from `row_sums → chunk_data → acc_col → trace.set_aux` (three N-sized live buffers) to `chunk_totals → chunk_data → trace.set_aux` (one N-sized live buffer, plus O(num_chunks) scalars). Peak transient storage drops from ~3N to ~N extension-field elements per table — roughly 670 MiB less allocator pressure on a 14-table proof at N=2^20.

Trade-off: one extra pass over `term_columns` in Pass 1 (just chunk totals, parallel). The final scatter into `trace.set_aux` is now sequential because `TraceTable` is row-major + `&mut self`-only; one extension-field add per row is ~10 ms/table — rounding error next to the Round 1 LDE FFTs.

Verified on `triage/pr518-fixes`:
- `cargo check --workspace` + `cargo check -p stark --no-default-features` — clean
- `make lint` — clean (fmt + clippy default + `debug-checks`)
- `cargo test -p stark --release --lib` — 121/121 pass
- `cargo test -p lambda-vm-prover --release --lib` non-ELF subset — 247/247 pass

Would be good to re-`/bench` once these land on the PR branch to confirm the −5.8% prove time from commit 95c77188 holds (expect at worst flat, at best a small additional heap win).

PR #519, opened 2026-04-21
feat(prover): Add Keccak precompile

jotabulacios:
This PR adds the Keccak precompile. The permutation is implemented in the executor (`keccak_f1600`, verified against `tiny-keccak`) and proven by three chips:

  1. **KECCAK core** (511 columns): ECALL receiver, one row per call, issues 25 combined `MEMW` read+write ops for the 200-byte state, packs the state into `Word4L` halves, and sends to the Keccak round bus.
  2. **KECCAK_RND round chip** (1,480 columns, 24 rows per call): runs one round per row. Theta / rho / pi / chi / iota are enforced through a mix of `BITWISE` lookups (`XOR_BYTE`, `AND_BYTE`, `HWSL`, `IS_BYTE`) and polynomial constraints (`IS_BIT` on the round-constant / rho-case selector bits, rnc equality, and the degree-3 `pi` mux).
  3. **KECCAK_RC table** (10 columns, preprocessed, 24 rows padded to 32): supplies the round constants via 1 bus interaction


erik-3milabs:
FYI: the spec is merged (#474) GL!

MauroToscano (approved):

nicole-graus (approved):

PR #524, opened 2026-04-22
Sequential column reads in commit_columns_bit_reversed

gabrielbosio:
Hash leaves in natural order (col_0[k] || col_1[k] || ...) inside parallel loop, then apply a single in_place_bit_reverse_permute over the resulting digest vector. Produces the same Merkle tree as reading at br(k) inside the loop, but trades ~2GB of scattered column reads for sequential, cache-friendly reads plus one ~64MB bit-reverse pass over 32-byte digests.


gabrielbosio:
No significant improvements on proving time.

PR #525, opened 2026-04-22
Strided barycentric OOD eval, skip per-column LDE pre-extraction

gabrielbosio:
Replace the per-column `Vec<FieldElement>` pre-extraction in `get_trace_evaluations_from_lde` with a strided variant of the barycentric interpolator that reads directly from the already column-major LDE storage with `stride = blowup_factor`. Same values, no intermediate clone.

On `fib_iterative_2M`: R3 OOD evaluation 5.66s _ 4.66s (-17.7%). Avoids cloning ~(`num_main` + `num_aux`) * `trace_size` field elements per proof.

### DIFF
--- a/crypto/math/src/polynomial/mod.rs
+++ b/crypto/math/src/polynomial/mod.rs
@@ -812,6 +812,75 @@ where
     &scalar * &(vanishing * &sum) // F × E → E
 }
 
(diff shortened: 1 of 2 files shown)


gabrielbosio:
No significant improvements on proving time.

PR #532, opened 2026-04-23
spec: `ARE_BYTES`

erik-3milabs:
i) introduces the `ARE_BYTES` lookup,
ii) introduces the `IS_BYTE<X, μ>` template that maps to `ARE_BYTES[0, X]/μ`
iii) removes the `IS_BYTE` lookup

### DIFF
--- a/spec/src/is_byte.toml
+++ b/spec/src/is_byte.toml
@@ -1,14 +1,14 @@
 name = "IS_BYTE"
 
+[[variables.condition]]
+name = "cond"
+type = "BaseField"
+desc = ""
+
 [[variables.input]]
 name = "X"
 type = "BaseField"
 desc = "Value for which to assert that it lies in the range $[0, 255]$."
-
-[[variables.multiplicity]]
-name = "μ"
-type = "BaseField"
-desc = ""
 
 [[constraint_groups]]
 name = "all"
@@ -17,5 +17,5 @@ name = "all"
 kind = "interaction"
 tag = "ARE_BYTES"
 input = [0, "X"]
-multiplicity = "μ"
+multiplicity = "cond"
 ref = "isbyte:c:isbyte"
--- a/spec/src/is_byte.toml
+++ b/spec/src/is_byte.toml
@@ -1,0 +1,21 @@
+name = "IS_BYTE"
+
+[[variables.input]]
+name = "X"
+type = "BaseField"
+desc = "Value for which to assert that it lies in the range $[0, 255]$."
+
+[[variables.multiplicity]]
+name = "μ"
+type = "BaseField"
+desc = ""
+
+[[constraint_groups]]
+name = "all"
+
+[[constraints.all]]
+kind = "interaction"
+tag = "ARE_BYTES"
+input = [0, "X"]
+multiplicity = "μ"
+ref = "isbyte:c:isbyte"
--- a/spec/bitwise.typ
+++ b/spec/bitwise.typ
@@ -37,8 +37,6 @@ This chip adds the following interactions to the lookup:
 
 = Notes/Optimizations
 The following ideas may prove to be optimizations for the #bitwise chip:
-+ Extend `IS_BYTE[X]` to `ARE_BYTES[X, Y]`, such that two bytes are range checked at once. 
-  When only a single check is required, one can still execute `IS_BYTE[X] := ARE_BYTES[X, 0]`.
 + Drop `MSB8` column, and instead define the `MSB8` lookup as `MSB8<X> := MSB16[256X]`.
   Note: currently, `MSB8` also implicity range checks the input `X` (the lookup fails if `X` is not a `Byte`).
   This optimization should only be executed when all chips leveraging `MSB8` do _not_ need this implicit range check.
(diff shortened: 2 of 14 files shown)


spec/src/is_byte.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[[variables.input]]
+name = "X"
+type = "BaseField"
+desc = "Value for which to assert that it lies in the range $[0, 255]$."
+
+[[variables.multiplicity]]
+name = "μ"
+type = "BaseField"
+desc = ""
```
RobinJadoul: This should probably be a `cond` instead?

RobinJadoul (commented):
Let's make an issue to manually add the batching to the spec once we're done with the rework of constraint numbering?

spec/src/is_byte.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[[constraint_groups]]
+name = "all"
+
+[[constraints.all]]
+kind = "interaction"
+tag = "ARE_BYTES"
+input = [0, "X"]
```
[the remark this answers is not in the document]
erik-3milabs: > that implementors should prefer pairing directly into ARE_BYTES[X, Y] whenever possible ...

This is exactly what is written in the IS_BYTE template description.

spec/src/is_byte.toml
erik-3milabs: During development, I named it `cond` initially, but then switched to `mu` to highlight that the multiplicity should be the second parameter: the `IS_BYTE<BaseField, BaseField>` interface will be OK with swapping `X` and `cond`

erik-3milabs: Alternatively, we could stick with it being a multiplicity, and have this be the first template which has a multiplicity (?)

erik-3milabs:
> Let's make an issue to manually add the batching to the spec once we're done with the rework of constraint numbering?

I don't think we should batch them in the spec:

1. **No added clarity**: Whoever reads the spec and decides to build a VM out of it, will have to read the `IS_BYTE` template section anyway to see how to expand the template. There they'll read about the optimization right away.
2. **Readability-optimization balance**: when batching the `IS_BYTE`s present in each chip, we'll inevitably (now or in the future) find cases where `IS_BYTE`s located in _different_ constraint sections should be combined for optimal performance. At that point, we'll have to decide to
    i. move one of the constraints to the other section, harming readability, or
    ii. leave it be, suggesting to the reader that this is an exceptional case where batching should NOT be performed (thus harming performance)
3. **Induced ordering**: batching inside the spec means hardcoding which columns should be combined. However, perhaps there might be implementation reasons to batch in a different manner. In that case, one would have to deviate from the spec to gain performance.
4. **Constraint blowup**: while batching will improve conciseness in some places, others (especially iterated interactions) will become less readable. E.g., `x∈[0, 3], y∈[0,7]: IS_BYTE<A[x][y]>` would be morphed into `x∈[0, 3], y∈[0,3]: ARE_BYTES<A[x][2y], A[x][2y+1]>`, which I find harder to parse.

spec/src/is_byte.toml
RobinJadoul: Why do we want to pass it as a parameter and not a template cond though?
It leading to an actual multiplicity seems no different from how NEG's cond leads to a multiplicity for ZERO to me. And just having `IS_BYTE<x>` vs `IS_BYTE<x, 1>` looks a lot more readable to me.

RobinJadoul:
I don't disagree with those arguments, my main reason to do it despite those would be to keep the clear correspondence between spec and implementation stricter.

spec/src/is_byte.toml
erik-3milabs: > And just having IS_BYTE<x> vs IS_BYTE<x, 1> looks a lot more readable to me.

I completely agree with you.

> It leading to an actual multiplicity seems no different from how NEG's cond leads to a multiplicity for ZERO to me.

There would be a difference: `NEG`'s `cond` is assumed to be a `Bit`, where that would not be the case here. This is what was stopping me previously, but I agree that `cond => IS_BIT<X>` is much easier on the eyes

erik-3milabs:
> my main reason to do it despite those would be to keep the clear correspondence between spec and implementation stricter.

Let us note that it would be to keep the _possibility_ of having a stricter correspondence between spec and implementation: the fact that we increase the level of detail to the spec does NOT necessarily mean that the devs will (eternally) ensure the implementation matches the spec. I would argue that sticking with `IS_BYTE`s in favor of `ARE_BYTES` will actually lead to an overall stricter correspondence between spec and implementation.

But I'll bite: how do you propose we resolve the dilemma I spelled out in the second point?

RobinJadoul:
I still agree with those points, and if indeed we'd end up saying the spec should reflect the batching, we'd have to make sacrifices.
I'm not necessarily trying to argue this is what we should do, just trying to bring it up for consideration.

I think we can probably table it for now, and keep it implicit. If at any point in the future we have reason to reconsider (e.g. we notice issues, or start extracting code from the spec that we want to keep simple) we can deal with the readability then.

erik-3milabs:
It is now listed as issue #571

RobinJadoul (approved):

PR #533, opened 2026-04-23
spec: shiroa fixes

RobinJadoul:
- Strip raw blocks from chapter titles for `project` argument
- Add an explicit description (based on chapter title) to chapters to avoid compilation issues when context appears early in the chapter
- Export interaction counts from the pdf version to use in shiroa, since otherwise we run into convergence issues that are hard to debug

### DIFF
--- a/spec/book.typ
+++ b/spec/book.typ
@@ -106,6 +106,7 @@
 
 // Invisibly include another chapter, so that its labels can be resolved
 #let xref-include(f) = {
+  show ref: none
   context {
     place(hide(box(width: auto, height: 0%, strip-all(include "/" + f))))
   }
@@ -170,13 +171,19 @@
         }
       })
       let cond() = _toplevel.final() == file
-      project.with(..args, title: context meta_sections.find(x => x.at(0) == _toplevel.final()).at(1), cond: cond)([
-        #show ref: it => context if _toplevel.final() == file {
-          xref(it)
-        }
+      show ref: it => context if _toplevel.final() == file {
+        xref(it)
+      }
+      let title = context {
+        // Strip raw, because shiroa already makes the title raw
+        show raw: it => it.text
+        meta_sections.find(x => x.at(0) == _toplevel.final()).at(1)
+      }
+      project.with(..args, title: title, description: plain-text(meta_sections.find(x => x.at(0) == file).at(1)), cond: cond)([
         #context _xref-included.final().pairs().map(((key, value)) => context if value and cond() {
           xref-include(key)
         }).join()
+        #metadata(json("interaction_count.json").sum(default: (:)))<interaction_count>
         #body
       ])
     }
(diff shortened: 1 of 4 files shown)


spec/book.typ
```diff
@@ -171,13 +171,19 @@
         }
       })
       let cond() = _toplevel.final() == file
-      project.with(..args, title: context meta_sections.find(x => x.at(0) == _toplevel.final()).at(1), cond: cond)([
-        #show ref: it => context if _toplevel.final() == file {
-          xref(it)
-        }
+      show ref: it => context if _toplevel.final() == file {
+        xref(it)
+      }
```
erik-3milabs:
```suggestion
      show ref: it => context if cond() { xref(it) }
```

spec/build_shiroa.sh
```diff
@@ -0,0 +1,18 @@
...
+rm -f interaction_count.json
+
+# Always clean up after ourselves
+trap 'rm -f interaction_count.json' EXIT
+
+# Query the ebook version for the proper counts
+typst query ebook.typ '<interaction_count>' --field value > interaction_count.json
+
+# And build
+shiroa build
```
erik-3milabs, line 27: Just as a safety precaution

```suggestion
RAYON_NUM_THREADS=1 shiroa build
```

Also, this stuff still freezes my PC due to RAM usage :(

erik-3milabs (approved):

spec/book.typ
```diff
@@ -170,13 +171,19 @@
...
+        // Strip raw, because shiroa already makes the title raw
+        show raw: it => it.text
+        meta_sections.find(x => x.at(0) == _toplevel.final()).at(1)
+      }
+      project.with(..args, title: title, description: plain-text(meta_sections.find(x => x.at(0) == file).at(1)), cond: cond)([
         #context _xref-included.final().pairs().map(((key, value)) => context if value and cond() {
           xref-include(key)
         }).join()
+        #metadata(json("interaction_count.json").sum(default: (:)))<interaction_count>
         #body
```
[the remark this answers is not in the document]
RobinJadoul: The suggested guard will only make things worse by breaking the shiroa build when `--input skip-interaction-count=1` is given, and not meaningfully improve discoverability

spec/build_shiroa.sh
RobinJadoul: idk if that should be a default in the build script, vs being set in your local env. It would also make the builds slower when the memory is not a restriction.
I could add a `build_shiroa_slow.sh` that adds the flag?

erik-3milabs: Perhaps a better solution would be to allow the number of threads to be passed as a parameter/have a flag to enable multi-threading, where in the script it is documented that this might be a tricky thing to do. And then print a message at start such as 

> _Starting shiroa build on X threads..._

when executing to indicate what is happening & suggest to people new to the script that they can multi-thread the process.

Given that (even with THREADS=1) this program often crashes my PC (which has a not unreasonably low 16GB of RAM), I think that we should be cautious to enable multi-threading by default.

PR #545, opened 2026-04-23
Misc optimizations

gabrielbosio:
Stacks five independent prover optimizations in R1 commit, trace build, and the FFT / batch-inverse pipelines.

1. Sequential reads in `commit_columns_bit_reversed`: replaces scattered `columns[col][br(row)]` reads (~2 GB) with sequential reads plus a 64 MB post bit-reverse on the digest vector.
2. `par_chunks` inside `chunk_and_generate`: runs the 10 trace-build Phase 5 generators in parallel on the idle rayon pool.
3. Skip the redundant bit-reverse pair in the R4 deep-composition LDE: adds `evaluate_fft_bit_reversed` to drop both cancelling permutes.
4. Parallel `in_place_bit_reverse_permute`: swap pairs `(i, br(i))` with `i < br(i)` are disjoint, so a Send/Sync raw-pointer wrapper lets rayon drive them; sequential fallback below N=16K.
5. Chunked parallel `inplace_batch_inverse`: chunks the Montgomery prefix product, trading K−1 extra inversions for K-way parallelism; sequential fallback below N=2^16.


gabrielbosio:
It only shows a speedup with low table parallelism. Closing.

gabrielbosio:
Got outdated, consider investigating again about possible optimizations with the current implementation on main

PR #547, opened 2026-04-23
Merkle cache reads and skip R4 permute

gabrielbosio:


gabrielbosio:
Reopening #545 which is a superset of this PR.

PR #554, opened 2026-04-24
spec: Fix `KECCAK`

### LINKED ISSUE #546: Keccak spec mismatches affecting verification
While implementing the Keccak precompile #519 , we found two spec-level mismatches that break end-to-end verification.

#### Mismatch 1 - Off-by-one carry offset in `rotated_Cxz` and `rho`

**Files / lines**

- `spec/src/keccak_round.toml:78` — `rotated_Cxz` virtual
- `spec/src/keccak_round.toml:102, 110, 118, 126` — four `rho` mux cases

**Root cause**

`HWSL(hw, z)` produces a 16-bit carry (`SLLC`), which occupies **2 bytes** in `Cxz_right`. The spec formulas propagate that carry with a 1-byte shift (`z-1`, `z-3`, `z-5`, `z-7`). It should be a 2-byte shift.

**Byte-level proof**

Take `Cxz = 0x8000_0000_0000_0000`, rotate left by 1, expected `0x0000_0000_0000_0001`.

After HWSL per halfword: `Cxz_right = [0, 0, 0, 0, 0, 0, 0x01, 0x00]` (SLLC₃ at bytes 6–7).

At `z=0`, the target output byte is `0x01`.

- Spec: `Cxz_right[(0-1) mod 8] = Cxz_right[7] = 0x00`  (bit dropped)
- Correct: `Cxz_right[(0-2) mod 8] = Cxz_right[6] = 0x01`

**Empirical failures** (applying spec literally):

- `test_keccak_round_trace_matches_f1600` diverges from `keccak_f1600` at round 2, lane (0,0), byte 6 (expected 32, got 48).
- `test_keccak_core_round_state_consistency` wrong at round 23.
- `test_prove_elfs_keccak` — STARK verifier rejects the proof.

**Proposed fix** — `spec/src/keccak_round.toml`:

| Line | Current | Fix |
|---|---|---|
| 78 (`rotated_Cxz`) | `["-", "z", 1]` | `["-", "z", 2]` |
| 102 (rho case 0) | `["-", "z", 1]` | `["-", "z", 2]` |
| 110 (rho case 1) | `["-", "z", 3]` | `["-", "z", 4]` |
| 118 (rho case 2) | `["-", "z", 5]` | `["-", "z", 6]` |
| 126 (rho case 3) | `["-", "z", 7]` | `["-", "z", 8]` (or `["mod", "z", 8]`) |

#### Mismatch 2 - ECALL payload field-ordering mismatch

**File / line**

- `spec/src/keccak.toml:51`

**Root cause**

The keccak ECALL payload uses `arr(hi32, lo32)`. Every other ECALL receiver (`halt.toml`, `commit.toml`) and every other `arr(...)` in the spec uses `[lo, hi]` per the `DWordWL` convention (`config.toml:69-74`). The shared CPU sender also emits `[lo, hi]`. Because LogUp fingerprints are order-sensitive, the bus doesn't balance.

**Current**

```toml
input = ["timestamp", ["arr", ["-", ["^", 2, 32], 1], ["-", ["^", 2, 32], 2]]]
# flattens to [ts_lo, ts_hi, 2^32-1, 2^32-2] = [ts_lo, ts_hi, hi32_of_-2, lo32_of_-2]
```

**Comparison**

| Chip | Spec | Flattened | Matches CPU sender `[ts_lo, ts_hi, lo, hi]`? |
|---|---|---|---|
| HALT | `cast(93, DWordWL)` | `[ts_lo, ts_hi, 93, 0]` | ✓ |
| COMMIT | `cast(64, DWordWL)` | `[ts_lo, ts_hi, 64, 0]` | ✓ |
| KECCAK | `arr(2^32-1, 2^32-2)` | `[ts_lo, ts_hi, hi, lo]` | ✗ |

**Empirical failure**

Applying the spec literally (receiver in `[hi, lo]` order while CPU sends `[lo, hi]`): `test_prove_elfs_keccak` fails with `LogUp bus does not balance`.

**Proposed fix** — `spec/src/keccak.toml:51`:

```toml
# Before
input = ["timestamp", ["arr", ["-", ["^", 2, 32], 1], ["-", ["^", 2, 32], 2]]]

# After
input = ["timestamp", ["cast", -2, "DWordWL"]]
```

`cast(-2, DWordWL)` produces `[lo32, hi32] = [2^32-2, 2^32-1]`, matching the CPU sender and HALT/COMMIT.

erik-3milabs:
Fixes two mistakes in the KECCAK chip spec:
1. off-by-one indexing errors for cyclic shifts,
2. ordering mistake constructing ECALL numbers.

Closes #546

### DIFF
--- a/spec/keccak.typ
+++ b/spec/keccak.typ
@@ -109,6 +109,7 @@ Lastly, the round chip contributes the following interactions to the lookup:
   Since `rnc` is constant, @keccak:c:rho_rotation makes those zero-bytes show up in `rot_left` and `rot_right` at constant locations.
   This means 96 columns can be removed from the chip at no cost.
   Likewise, 96 `IS_BYTE` interactions can be dropped from @keccak:c:range_rot_left and @keccak:c:range_rot_right.
+  - the shift-constants are equivalent to $1 mod 16$ for $(#`x`, #`y`) = (1, 0)$ and $-1 mod 16$ for $(2, 3)$. This means that for those lanes it suffices to constrain `rot_left`/`rot_right` as `Bit`s rather than `Byte`s, saving an additional 8 `IS_BYTE` interactions.
 - $#`rc[2]` = #`rc[4]` = #`rc[5]` = #`rc[6]` = 0$. As such, those elements need not be stored in `rc`, and need not be XORed into the state in the $iota$-step. This saves 8 columns and 4 `XOR_BYTE` interactions.
 - when executed in large volumnes, `KECCAK_RND` could benefit from having a three-way XOR lookup table. With this in place, the 80 interactions in @keccak:c:theta_cxz_start and @keccak:c:theta_cxz could be dropped.
   Likewise, 80 columns could be removed from the chip (a \~5% savings).
--- a/spec/keccak.typ
+++ b/spec/keccak.typ
@@ -105,6 +105,10 @@ Lastly, the round chip contributes the following interactions to the lookup:
 == Notes/potential optimizations
 - one does not have to repeat `addr` in `state_ptr`; this saves 4 columns and 4 `IS_HALF` checks.
 - step $rho$ does not need to be applied to `state[0][0]`; its has a zero-shift. This saves 16 columns and 4 `HWSL` interactions.
+- when the output of `HWSL` are `Byte`s mapped as `Half`s, we find that out of every four output bytes, at least one is zero. 
+  Since `rnc` is constant, @keccak:c:rho_rotation makes those zero-bytes show up in `rot_left` and `rot_right` at constant locations.
+  This means 96 columns can be removed from the chip at no cost.
+  Likewise, 96 `IS_BYTE` interactions can be dropped from @keccak:c:range_rot_left and @keccak:c:range_rot_right.
 - $#`rc[2]` = #`rc[4]` = #`rc[5]` = #`rc[6]` = 0$. As such, those elements need not be stored in `rc`, and need not be XORed into the state in the $iota$-step. This saves 8 columns and 4 `XOR_BYTE` interactions.
 - when executed in large volumnes, `KECCAK_RND` could benefit from having a three-way XOR lookup table. With this in place, the 80 interactions in @keccak:c:theta_cxz_start and @keccak:c:theta_cxz could be dropped.
   Likewise, 80 columns could be removed from the chip (a \~5% savings).
--- a/spec/src/keccak_round.toml
+++ b/spec/src/keccak_round.toml
@@ -75,7 +75,7 @@ desc = "state update following from step $iota$."
 name = "rotated_Cxz"
 type = [["Byte", 8], 5]
 desc = "$#`Cxz[x,`3#`,z]` <<< 1$"
-def = {iters=[["x", 0, 4], ["z", 0, 7]], poly=["+", ["idx", ["idx", "Cxz_left", "x"], "z"], ["idx", ["idx", "Cxz_right", "x"], ["mod", ["-", "z", 1], 8]]]}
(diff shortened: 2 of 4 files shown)


spec/src/keccak_round.toml
```diff
@@ -75,7 +75,7 @@ desc = "state update following from step $iota$."
 name = "rotated_Cxz"
 type = [["Byte", 8], 5]
 desc = "$#`Cxz[x,`3#`,z]` <<< 1$"
-def = {iters=[["x", 0, 4], ["z", 0, 7]], poly=["+", ["idx", ["idx", "Cxz_left", "x"], "z"], ["idx", ["idx", "Cxz_right", "x"], ["mod", ["-", "z", 1], 8]]]}
+def = {iters=[["x", 0, 4], ["z", 0, 7]], poly=["+", ["idx", ["idx", "Cxz_left", "x"], "z"], ["idx", ["idx", "Cxz_right", "x"], ["mod", ["-", "z", 2], 8]]]}
```
RobinJadoul: I think you only need this for half the elements (even offsets), and the odd offsets are just already "clean"? So this seems to imply we can shrink `Cxz_right` since only it's LSB matters
I think it's technically correct, but we may gain some columns and byte lookups here

spec/src/keccak_round.toml
RobinJadoul: typo spotted: "start state"

spec/src/keccak_round.toml
```diff
@@ -218,11 +221,11 @@ iters = [["x", 0, 4], ["z", 0, 7]]
...
 [[constraints.theta]]
-kind = "interaction"
-tag = "IS_BYTE"
+kind = "template"
+tag = "IS_BIT"
 input = [["idx", ["idx", "Cxz_right", "x"], "z"]]
-iters = [["x", 0, 4], ["z", 0, 7]]
-multiplicity = "μ"
+iters = [["x", 0, 4], ["z", 0, 3]]
+cond = "μ"
```
RobinJadoul: We can probably leave `count` out, since it will still work for padding rows, and it's the only higher-degree constraint, afaict

spec/keccak.typ
```diff
@@ -105,6 +105,10 @@ Lastly, the round chip contributes the following interactions to the lookup:
 == Notes/potential optimizations
 - one does not have to repeat `addr` in `state_ptr`; this saves 4 columns and 4 `IS_HALF` checks.
 - step $rho$ does not need to be applied to `state[0][0]`; its has a zero-shift. This saves 16 columns and 4 `HWSL` interactions.
+- when the output of `HWSL` are `Byte`s mapped as `Half`s, we find that out of every four output bytes, at least one is zero.
```
RobinJadoul, line 108: I think there's some `1 mod 16` parts in the round constants? So we can make a note of further removing byte checks in favor of bit checks there

spec/src/keccak_round.toml
```diff
@@ -75,7 +75,7 @@ desc = "state update following from step $iota$."
 name = "rotated_Cxz"
 type = [["Byte", 8], 5]
 desc = "$#`Cxz[x,`3#`,z]` <<< 1$"
-def = {iters=[["x", 0, 4], ["z", 0, 7]], poly=["+", ["idx", ["idx", "Cxz_left", "x"], "z"], ["idx", ["idx", "Cxz_right", "x"], ["mod", ["-", "z", 2], 8]]]}
+def = {iters=[["x", 0, 4], ["z", 0, 7]], poly=["+", ["idx", ["idx", "Cxz_left", "x"], "z"], ["*", ["-", 1, ["mod", "z", 2]], ["idx", ["idx", "Cxz_right", "x"], ["mod", ["-", ["/", "z", 2], 1], 4]]]]}
```
RobinJadoul: Tooling/CI is failing on the division
If this is not field division, we may even have to introduce something for floor division, because otherwise it's gonna act weird

erik-3milabs: As agreed, I've unrolled the definition.

RobinJadoul (approved):

PR #565, opened 2026-04-24
perf: sparse gather for LogUp aux trace when multiplicity is selector…

diegokingston:
…-gated

Many bus interactions are gated by mutually-exclusive selector columns (e.g., AND/OR/XOR op flags, ADD/SUB/MUL op flags): when the selector is zero, the row's contribution to the term column is zero regardless of the fingerprint. We now detect Column/Sum/Sum3 multiplicities and, when fewer than 7/8 of a chunk's rows are active, compute fingerprints only at the active rows before batch-inverting. Inactive rows stay at zero from the initial `vec![zero; trace_len]` allocation.

Soundness: `sparse_active_rows` marks a row as active whenever any contributing column is non-zero, which is an over-approximation of `m != 0` (cancellation can only cause us to process more rows, never fewer). Dense multiplicities (One, Negated, Diff, Linear) fall back to the unchanged contiguous path.

Applies to both single (`compute_logup_term_column`) and paired (`compute_logup_batched_term_column`) term-column builders. Verified via all 71 stark bus soundness tests and 265 prover lib tests.


MauroToscano:
Closing this one for now — the `/bench` delta (-0.6%) is within noise levels, so we can't justify carrying the extra complexity (sparse/dense/half-sparse paths) on the merits of the measurement alone.

To be clear this isn't a knock on the approach — the soundness reasoning (over-approximating `m != 0` via active columns) is correct. The main reason is that we plan to rework constraint evaluation more holistically later and monomorphize it, which should be the right place to capture this kind of win in a cleaner way. Let's revisit the sparse-gather idea in that context rather than bolting it onto the current path.

PR #570, opened 2026-04-28
spec: update constraint ID system

erik-3milabs:
This PR introduces content-derived constraint IDs, i.e., the ID of a constraint is derived from its content, rather than its index in the constraint list. This
1. permit addition, removing and reorganizing constraints without changing any IDs, and
2. enforces ID updates upon constraint modification

i.e., two constraints with the same ID are 100-ε% certain to be identical.

Note: the ID is derived in a way that updating a variable name does _not_ lead to a name update. This is at the expense of an ID update when variables are _reordered_.

Before:
image

After:
image

### DIFF
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -314,10 +314,11 @@
     
     let indices = (("",) + iters_of(constraint).map(it => it.at(0))).join(".")
 
-    let z-fill(s) = "0" * calc.max(2 - s.len(), 0) + s
-    let ref-tag(i) = raw(tag) + sub("/" + z-fill(str(i)))
+    let pad-width() = calc.max(calc.ceil(calc.log(counter(figure.where(kind: counter-kind)).final().at(0))), 2)
+    let z-pad(s) = context "0" * calc.max(pad-width() - s.len(), 0) + s
+    let ref-tag(i) = raw(tag) + sub("/" + z-pad(str(i)))
     return (
-      context super[#emph(z-fill(str(counter(figure.where(kind: counter-kind)).get().at(0) + 1)))],
+      context super[#emph(z-pad(str(counter(figure.where(kind: counter-kind)).get().at(0) + 1)))],
       [
         #show figure: (it) => align(left, raw(tag + indices))
         #cref(constraint)[#figure(kind: counter-kind, numbering: (i) => ref-tag(i), supplement: [], [])]
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -306,8 +306,9 @@
 
   /// Render the contraint's tag.
   let tag(constraint, group) = {
-    let counter-kind = chip.name + "constraint"
-    let tag = chip.name + "-" + constraint.id
+    let code = chip.at("code", default: chip.name)
+    let counter-kind = code + "constraint"
+    let tag = code + "-" + constraint.id
     
     let indices = (("",) + iters_of(constraint).map(it => it.at(0))).join(".")
 
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -269,10 +269,12 @@
 
 #let render_chip_assumptions(chip, config) = {
   let tag(assumption) = {
-    let with_index(x) = ((x,) + iters_of(assumption).map(it => it.at(0))).join(".")
-    let lbl = [#chip.name\-A]
-    show figure: (it) => align(left, block[#lbl#context with_index(it.counter.display())])
-    cref(assumption)[#figure(kind: chip.name + "assumption", numbering: (i) => [#lbl#i], supplement: [], [])]
+    let code = if "code" in chip {chip.code} else {chip.name}
+    let index = (("",) + iters_of(assumption).map(it => it.at(0))).join(`.`)
+    let lbl(idx) = raw(code + "-A" + str(idx))
+
+    show figure: (it) => align(left, block[#context lbl(it.counter.get().at(0))#index])
+    cref(assumption)[#figure(kind: code + "assumption", numbering: (i) => lbl(i), supplement: [], [])]
   }
 
   show figure: set block(breakable: true)
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -303,11 +303,8 @@
   assert(groups.all(group => group in all_groups), message: "unknown group: " + repr(groups))
   let selected_constraints = groups.map(g => ((g): chip.constraints.at(g))).join()
 
-  // Find the group definition in the constraint_groups
-  let lookup_group(name) = chip.constraint_groups.filter((g) => g.name == name).at(0, default: (name: name))
-
   /// Render the contraint's tag.
-  let tag(constraint, group) = {
+  let tag(constraint) = {
     let code = chip.at("code", default: chip.name)
     let counter-kind = code + "constraint"
     let tag = code + "-" + constraint.id
@@ -413,7 +410,7 @@
     ..for (group, group_constraints) in selected_constraints.pairs() {
       for constraint in group_constraints {
         (
-          ..tag(constraint, lookup_group(group)),
+          ..tag(constraint),
           [#iters(constraint)],
           [#repr_constraint(constraint)],
           [#expr_to_math(constraint.at("multiplicity", default: ""))],
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -389,7 +389,7 @@
 
   show figure: set block(breakable: true)
   figure(table(
-    columns: (auto, auto, auto, 1fr, auto),
+    columns: (auto, auto, if do_display_range {auto} else {0pt}, 1fr, if do_display_multiplicity {auto} else {0pt}),
     inset: 6pt,
     align: (top + left, top + left, top + left, top + left, top + center),
     stroke: none,
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -306,11 +306,20 @@
 
   /// Render the contraint's tag.
   let tag(constraint, group) = {
-    let with_index(x) = ((raw(x),) + iters_of(constraint).map(it => raw(it.at(0)))).join(".")
-    let prefix = if "prefix" in group { group.prefix }
-    let tag(x) = [#raw(chip.name + "-" + prefix + "" + constraint.id + "/")#with_index(x)]
-    show figure: (it) => align(left, context tag(it.counter.display()))
-    cref(constraint)[#figure(kind: chip.name + "constraint", numbering: (i) => tag(str(i)), supplement: [], [])]
+    let counter-kind = chip.name + "constraint"
+    let tag = chip.name + "-" + constraint.id
+    
+    let indices = (("",) + iters_of(constraint).map(it => it.at(0))).join(".")
+
+    let z-fill(s) = "0" * (2 - s.len()) + s
+    let ref-tag(i) = raw(tag) + sub("/" + z-fill(str(i)))
+    return (
+      context super[#emph(z-fill(str(counter(figure.where(kind: counter-kind)).get().at(0) + 1)))],
+      [
+        #show figure: (it) => align(left, raw(tag + indices))
+        #cref(constraint)[#figure(kind: counter-kind, numbering: (i) => ref-tag(i), supplement: [], [])]
+      ],
+    )
   }
 
   /// Generates a representation of `constraint`
@@ -351,13 +360,23 @@
     }
 
     (..for poly in polys {
-      (table.cell(align: right, colspan: 2, [_polynomial_]), $#expr_to_math(poly) = 0$, [])
+      (
+        [], 
+        table.cell(align: right, colspan: 2, [_polynomial_]), 
+        table.cell(align: left, colspan: 1, $#expr_to_math(poly) = 0$), 
+        []
+      )
     },)
   }
 
   // Rendering the additional "desc" field for arith constraints
   let render_extra_description(constraint) = {
-    (table.cell(align: right, colspan: 2, [_description_]), eval(constraint.desc, mode: "markup"), [])
+    (
+      [],
+      table.cell(align: right, colspan: 2, [_description_]), 
+      table.cell(align: left, colspan: 1, eval(constraint.desc, mode: "markup")), 
+      []
+    )
   }
 
   // Whether there is at least one constraint with a range
@@ -370,11 +389,12 @@
 
   show figure: set block(breakable: true)
   figure(table(
-    columns: (auto, auto, 1fr, auto),
+    columns: (auto, auto, auto, 1fr, auto),
     inset: 6pt,
-    align: (top + left, top + left, top + left, top + center),
+    align: (top + left, top + left, top + left, top + left, top + center),
     stroke: none,
     table.header(
+[],
       [*Tag*], 
       if do_display_range {[*Range*]} else {[]}, 
       [*Description*], 
@@ -384,7 +404,7 @@
     ..for (group, group_constraints) in selected_constraints.pairs() {
       for constraint in group_constraints {
         (
-          [#tag(constraint, lookup_group(group))],
+          ..tag(constraint, lookup_group(group)),
           [#iters(constraint)],
           [#repr_constraint(constraint)],
           [#expr_to_math(constraint.at("multiplicity", default: ""))],
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -24,7 +24,7 @@
     
     config.variables.types.filter(type => type.label == var_type).first().subtypes.len() * factor
   })
-  .sum()
+  .sum(default: 0)
 }
 
 // Given a constraint, compute the number of interactions it induces
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -314,7 +314,7 @@
     
     let indices = (("",) + iters_of(constraint).map(it => it.at(0))).join(".")
 
-    let z-fill(s) = "0" * (2 - s.len()) + s
+    let z-fill(s) = "0" * calc.max(2 - s.len(), 0) + s
     let ref-tag(i) = raw(tag) + sub("/" + z-fill(str(i)))
     return (
       context super[#emph(z-fill(str(counter(figure.where(kind: counter-kind)).get().at(0) + 1)))],
--- a/spec/chip.typ
+++ b/spec/chip.typ
@@ -269,7 +269,7 @@
 
 #let render_chip_assumptions(chip, config) = {
   let tag(assumption) = {
-    let code = if "code" in chip {chip.code} else {chip.name}
+    let code = chip.at("code", default: chip.name)
     let index = (("",) + iters_of(assumption).map(it => it.at(0))).join(`.`)
     let lbl(idx) = raw(code + "-A" + str(idx))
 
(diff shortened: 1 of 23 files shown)


spec/src.typ
```diff
@@ -115,12 +115,157 @@
...
 
+/// Fowler-Noll-Vo (FNV) hash function, version 1a
+/// Src: https://en.wikipedia.org/wiki/Fowler-Noll-Vo_hash_function
+/// 
+/// Note: this is a non-cryptographic hash function; it is optimized
+/// for speed at the expense of unpredictability.
+/// 
+/// This implementation operates on two 32-bit limbs, rather than a single 
+/// 64-bit limb, since Typst does not support u64s.
+#let FNV-1a(bytes) = {
```
RobinJadoul: Any reasoning for not using a typst package for hashing?
e.g. digestify should be running in wasm, so speed should not be a massive concern (probably)

spec/chip.typ
```diff
@@ -301,11 +303,21 @@
...
-    let lbl = [#chip.name\-C#prefix]
-    show figure: (it) => align(left, block[#lbl#context with_index(it.counter.display())])
-    cref(constraint)[#figure(kind: chip.name + "constraint", numbering: (i) => [#lbl#i], supplement: [], [])]
+    let code = chip.at("code", default: chip.name)
+    let counter-kind = code + "constraint"
+    let tag = code + "-" + constraint.id
+    
+    let indices = (("",) + iters_of(constraint).map(it => it.at(0))).join(".")
+
+    let z-fill(s) = "0" * calc.max(2 - s.len(), 0) + s
```
RobinJadoul: Appears a few times, maybe lift to the top level.

spec/chip.typ
```diff
@@ -301,11 +303,21 @@
...
+    let ref-tag(i) = raw(tag) + sub("/" + z-fill(str(i)))
+    return (
+      context super[#emph(z-fill(str(counter(figure.where(kind: counter-kind)).get().at(0) + 1)))],
```
RobinJadoul: `counter(...).final()` could give us the total count of constraints, to maybe be able to z-fill more than 2 digits if required

spec/src.typ
```diff
@@ -115,12 +115,157 @@
...
+    let lo = lo.bit-and(0xFFFFFFFF)
+    let hi = (hi + carry).bit-and(0xFFFFFFFF)
+    hash = (lo, hi)
+  }
+
+  hash.map(int.to-bytes).join()
+}
+
+/// Converts a byte array to a hexadecimal string
+#let bytes-to-hex(bytes) = {
```
RobinJadoul: Is there much need to carry hex around instead of going straight from bytes to base 32?

RobinJadoul (commented):
Very brief look only, so far; haven't properly looked at the canonicalization/serialization of the constraints yet

spec/src.typ
erik-3milabs: - `digestify` is available under the MIT license. When I skimmed the license last week, I thought that using it might require our spec to also be released under the MIT (or compatible) license. Having a closer look now, this might not actually be the case. Still, I'm not a legal expert, so I decided to avoid it just in case.
- The other hashing package (`jumble`) is slower than this implementation.
- It was fun to learn a bit about non-cryptographic hashing, and implement this technique here.

spec/chip.typ
erik-3milabs: It is used twice (in `src.typ` and `chip.typ`). Since those two files don't have any shared dependencies/don't depend on one another, I'd have to create a new file that both can depend on. I didn't feel like doing this just for this single one-line function.

spec/src.typ
erik-3milabs: not really. dropped it

spec/chip.typ
```diff
@@ -306,11 +303,22 @@
...
-    let lbl = [#chip.name\-C#prefix]
-    show figure: (it) => align(left, block[#lbl#context with_index(it.counter.display())])
-    cref(constraint)[#figure(kind: chip.name + "constraint", numbering: (i) => [#lbl#i], supplement: [], [])]
+    let code = chip.at("code", default: chip.name)
+    let counter-kind = code + "constraint"
+    let tag = code + "-" + constraint.id
+    
+    let indices = (("",) + iters_of(constraint).map(it => it.at(0))).join(".")
+
+    let pad-width() = calc.max(calc.ceil(calc.log(counter(figure.where(kind: counter-kind)).final().at(0))), 2)
```
RobinJadoul: I think we technically a `+ 1` inside of the log, since e.g. `ceil(log(10))` is still only 1

spec/chip.typ
```diff
@@ -269,10 +264,12 @@
 
 #let render_chip_assumptions(chip, config) = {
   let tag(assumption) = {
-    let with_index(x) = ((x,) + iters_of(assumption).map(it => it.at(0))).join(".")
-    let lbl = [#chip.name\-A]
-    show figure: (it) => align(left, block[#lbl#context with_index(it.counter.display())])
-    cref(assumption)[#figure(kind: chip.name + "assumption", numbering: (i) => [#lbl#i], supplement: [], [])]
+    let code = if "code" in chip {chip.code} else {chip.name}
```
RobinJadoul: Can do the same as further down

```suggestion
    let code = chip.at("code", default: chip.name)
```

spec/src.typ
RobinJadoul: MIT licensed is fine, it only comes in to play if we're redistributing the licensed thing itself
> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

One reason a standard hash function could be nice is if we want to be able to recompute the tags in another environment (e.g. python when doing some code extraction or even type checking).
On the other hand, then FNV shouldn't be too hard to implement elsewhere either, so your call.

spec/src.typ
RobinJadoul: And I suppose this function can disappear then too

spec/src.typ
```diff
@@ -115,12 +115,158 @@
...
+    /// Flatten the type of a variable into a string
+    let flatten_vartype(typ) = {
+      if type(typ) == array {
+        "(" + typ.map(flatten_vartype).join(",") + ")"
+      } else {
+        str(typ)
+      }
+    }
+    
+    let flattened_type = lower(flatten_vartype(var.type))
```
RobinJadoul: Why lowercase?

spec/src.typ
```diff
@@ -115,12 +115,158 @@
...
+    let flatten_vartype(typ) = {
+      if type(typ) == array {
+        "(" + typ.map(flatten_vartype).join(",") + ")"
+      } else {
+        str(typ)
+      }
+    }
+    
+    let flattened_type = lower(flatten_vartype(var.type))    
+    let input = (chip, group, str(idx), flattened_type).join("\x00")
```
RobinJadoul: Do we want this to be invariant under reordering or not?

spec/src.typ
```diff
@@ -115,12 +115,158 @@
...
+          (var.name: digest_variable(chip.name, group, idx, var))
+        }).sum(default: (:))
+    }).sum(default: (:))
+
+  // replace variable with ID in LISP
+  let replace_variable_with_ID(lisp) = {
+    if type(lisp) == array {
+      "(" + lisp.map(replace_variable_with_ID).join(",") + ")"
+    } else {
+      variable_to_ID.at(str(lisp), default: str(lisp))
```
RobinJadoul: This implicitly also converts ints to strings (and maybe more)?

spec/src.typ
```diff
@@ -115,12 +115,158 @@
...
+      .replace("\n", "")
+      .replace(" ", "")
+  }
+
+  // Add an ID to each constraint
+  chip.constraints = chip.at("constraints", default: (:))
+    .pairs()
+    .map(((group, constraints)) => {
+      (
+        str(group):
```
RobinJadoul: You can just do `(group)`

spec/src.typ
```diff
@@ -115,12 +115,158 @@
...
+      if type(typ) == array {
+        "(" + typ.map(flatten_vartype).join(",") + ")"
+      } else {
+        str(typ)
+      }
+    }
+    
+    let flattened_type = lower(flatten_vartype(var.type))    
+    let input = (chip, group, str(idx), flattened_type).join("\x00")
+    digest_to_id(nchf(input))
```
RobinJadoul: We _could_ carry the whole digest for a variable ID, since it never sees the outside

spec/src.typ
```diff
@@ -115,12 +115,158 @@
...
+    )
+
+    assert(c.kind in CONSTRAINT_CAT_TO_SCOPE)
+    let id_tagged = CONSTRAINT_CAT_TO_SCOPE
+      .at(c.kind)
+      .filter(cat => cat in c.keys())
+      .map(cat => (str(cat): replace_variable_with_ID(c.at(cat))))
+      .sum(default: (:))
+
+    repr(id_tagged)
```
RobinJadoul: This may depend a bit too much on typst internals for canonicalization
Maybe `json.encode` can work, but then there's still a potential ordering issue

spec/src.typ
```diff
@@ -115,12 +115,158 @@
...
+    if type(lisp) == array {
+      "(" + lisp.map(replace_variable_with_ID).join(",") + ")"
+    } else {
+      variable_to_ID.at(str(lisp), default: str(lisp))
+    }
+  }
+
+  // Replace variable names with their ID 
+  let digestable_constraint(c) = { 
+    let CONSTRAINT_CAT_TO_SCOPE = (
```
RobinJadoul: Is it more future-proof to include-list or exclude-list here?

spec/src.typ
```diff
@@ -115,12 +115,158 @@
...
+      "interaction": ("tag", "iter", "input", "output", "multiplicity"),
+      "template": ("tag", "iter", "input", "output", "cond"),
+      "arith": ("iter", "poly")
+    )
+
+    assert(c.kind in CONSTRAINT_CAT_TO_SCOPE)
+    let id_tagged = CONSTRAINT_CAT_TO_SCOPE
```
RobinJadoul: Let's include the `kind` here too

spec/src.typ
```diff
@@ -115,12 +115,158 @@
...
+      variable_to_ID.at(str(lisp), default: str(lisp))
+    }
+  }
+
+  // Replace variable names with their ID 
+  let digestable_constraint(c) = { 
+    let CONSTRAINT_CAT_TO_SCOPE = (
+      "interaction": ("tag", "iter", "input", "output", "multiplicity"),
+      "template": ("tag", "iter", "input", "output", "cond"),
+      "arith": ("iter", "poly")
```
RobinJadoul: Doesn't include `iters`

RobinJadoul: Also, allowing iter variables to be renamed without changing the ID could be nice

RobinJadoul: We also might want to deal with summations similarly

spec/src.typ
erik-3milabs: Indeed. Do you think there would be a problem in doing so?

spec/src.typ
erik-3milabs: I believe to exclude-list. Updated it accordingly.

spec/src.typ
erik-3milabs: Agreed. I reworked it. Let me know if you agree that it is better now.

spec/src.typ
erik-3milabs: I managed to make the iter variable renaming work; from an effort-reward perspective, I propose we push summations out of scope.

spec/src.typ
erik-3milabs: under reordering of what? this array?

spec/src.typ
erik-3milabs: > MIT licensed is fine,

Alright.

> so your call

Having given it some thought, I propose we stick with FNV: using a non-cryptographic hash here signals that there is no need for security here. (Silly example that clarifies my point: https://www.reddit.com/r/Jokes/comments/k53qdf/typical_camp_commander/)

spec/src/sha256round.toml
```diff
@@ -1,4 +1,5 @@
 name = "SHA256ROUND"
+code = "SHR"
```
RobinJadoul: Hmm, maybe dangerous to confuse with shift right?

spec/src.typ
RobinJadoul: variables. e.g. for #670 I had to re-order some (virtual) variables that depended on each other in SHA256. So I'm tempted to think that variables should maybe be tied to their name rather than position.
I think we discussed a while back that the De Bruijn index idea was originally mostly for iters and summations

But all that would need, I think, it just replacing the str(idx) by the name of the variable here, and the rest would keep working

spec/src.typ
RobinJadoul: Not sure, I think it just felt a bit weird as a canonicalization step

spec/src.typ
```diff
@@ -115,12 +115,154 @@
...
+  
+  // ID settings
+  let ID_CHAR_LEN = 4;
+  let ID_CHAR_SET = "123456789ABDEFGHJKLMNPQRSTUVWXYZ".codepoints()
+
+  // Constants
+  let LOG_ID_RADIX = 5
+  let RADIX = calc.pow(2, LOG_ID_RADIX)
+  assert(ID_CHAR_SET.len() == RADIX, message: "ID_CHAR_SET <> RADIX mismatch")
+  assert(ID_CHAR_LEN * LOG_ID_RADIX <= 64, message: "ID_CHAR_LEN and RADIX incompatible")
```
RobinJadoul:
```suggestion
  assert(ID_CHAR_LEN * LOG_ID_RADIX <= 64, message: "digest size too small for configured ID entropy")
```

spec/src.typ
```diff
@@ -115,12 +115,154 @@
...
+    let int = int.from-bytes(digest.slice(0, count: MIN_DIGEST_BYTES))
+    for _ in range(ID_CHAR_LEN) {
+      let idx = int.bit-and(31)
+      int = int.bit-rshift(5)
+      (ID_CHAR_SET.at(idx), )
+    }.sum()
+  }
+
+  // Recursively flatten nested object
+  let flatten(obj, terminator: str, braces: ("(", ")"), sep: ",") = {
```
RobinJadoul: Perhaps for naming
```suggestion
  let flatten(obj, leaf_transform: str, braces: ("(", ")"), sep: ",") = {
```

spec/src.typ
```diff
@@ -115,12 +115,154 @@
...
+    let var_map = c.at("iters", default: if "iter" in c {(c.iter,)} else {()})
+      .enumerate()
+      .map(((idx, iter)) => ((iter.at(0)): "iter" + str(idx)))
+      .sum(default: (:))
+
+    let terminator(x) = variable_IDs.at(str(x), default: var_map.at(str(x), default: str(x)))
+
+    flatten(filtered, terminator: terminator)
+      .replace("\n", "")
+      .replace(" ", "")
```
RobinJadoul: Are we expecting whitespace to be generated?

spec/src.typ
RobinJadoul: It's definitely better, the encoding for dicts feels a bit weird, but not necessarily problematically so (unless we suddenly could have vars named `:`)

erik-3milabs:
@claude

spec/src/sha256round.toml
erik-3milabs: good catch. Let's make it... `SHD`
```suggestion
code = "SHD"
```
(please apply if you agree)

spec/src.typ
erik-3milabs: I updated the encoding. Among others, I included domain separators. Let me know what you think!

spec/src.typ
erik-3milabs: I updated the encoding to operate on bytes rather than strings.

spec/src.typ
erik-3milabs: removed it

spec/src.typ
```diff
@@ -146,11 +146,33 @@
...
+    ("b", val)
+  } else if t == str {
+    ("s", leaf-transform(val))
+  } else if t == dictionary {
+    ("d", val.pairs().map(recurse).sum(default: _SEP))
+  } else if t == array {
+    ("a", val.map(recurse).join(_SEP, default: _SEP))
+  } else if t == int {
+    ("i", val.to-bytes())
+  }
```
RobinJadoul, line 158: Default case with `assert` to be sure?

spec/src.typ
```diff
@@ -146,11 +146,33 @@
...
+#let _SEP = bytes(",")
+#let to-bytes(leaf-transform, val) = {
+  let recurse = to-bytes.with(leaf-transform)
+  let t = type(val)
+  let (tag, val) = if t == bytes {
+    ("b", val)
+  } else if t == str {
+    ("s", leaf-transform(val))
+  } else if t == dictionary {
+    ("d", val.pairs().map(recurse).sum(default: _SEP))
```
RobinJadoul: Add explicit sorting to ensure order-independence

spec/src.typ
```diff
@@ -178,66 +200,32 @@
...
-  let digestable_constraint(c) = {
+  let _EXCLUDED_LABELS = ("desc", "ref", "constraint")
+  let digest_constraint(c) = {
     // filter out excluded fields
-    let filtered = c
-      .pairs()
-      .filter(((label, value)) => label not in EXCLUDED_LABELS)
-      .map(((label, value)) => ((label): value))
-      .sum(default: (:))
+    let filtered = c.keys().filter(k => k not in _EXCLUDED_LABELS).map(k => c.at(k))
```
RobinJadoul: Depends on field ordering?

erik-3milabs:
> ## Codex Code Review
> 
> * **High** — [spec/src.typ:143](/home/runner/work/lambda_vm/lambda_vm/spec/src.typ:143): `lo + hi.bit-lshift(32)` reconstructs an unsigned 64-bit FNV value in Typst’s signed 64-bit integer. Whenever `hi` has its top bit set, this overflows and aborts spec generation. Serialize the two 32-bit limbs separately instead.

This is not correct; the last operation in the loop truncates `hi` to its lower 32 bits. Applying `bit-lshift(32)` to such a number does not result in an error, even when the 31st bit is set.

> * **High** — [spec/src.typ:161](/home/runner/work/lambda_vm/lambda_vm/spec/src.typ:161): `val.to-bytes()` uses unsigned encoding by default, but many constraints contain negative integers (`multiplicity = -1`, negative constants/exponents). Loading those chips will fail while assigning IDs. Encode integers with `signed: true` or define an explicit sign-aware representation.

This is incorrect; `-1.to-bytes()` simply produces `(255,255,255,255,255,255,255,255)`.

erik-3milabs:
> ## Review summary
> 
> ...
> 
> * FNV-1a: the high 32-bit limb never influences the final ID (only the low 3 bytes of `lo` are consumed by `digest_to_id`, and `lo`'s recurrence doesn't read `hi`). Not a bug, but the high-limb work is effectively dead and `hi.bit-lshift(32)` relies on i64 shift-into-sign-bit wraparound. See inline comment — worth confirming the collision claim with the true 20-bit effective entropy in mind.

This is an interesting observation. I don't think this is a problem, but it does go to show that FNV-1a is more finicky than I thought it would be...

RobinJadoul (commented):
For the hash: http://www.isthe.com/chongo/tech/comp/fnv/index.html#xor-fold

spec/src.typ
```diff
@@ -138,14 +138,21 @@
...
   (lo + hi.bit-lshift(32)).to-bytes()
 }
 
+// 32-bit FNV, derived from FNV64-1a through XOR-folding.
+// src: http://www.isthe.com/chongo/tech/comp/fnv/index.html#xor-fold
+#let FNV32-1a(inp) = {
+  let a64 = array(FNV64-1a(inp))
+  return bytes(a64.zip(a64.slice(4)).map(((lhs,rhs)) => lhs.bit-xor(rhs)))
+}
+
```
RobinJadoul, line 140: Isn't FNV32 its own thing, and then you'd go from 32 to 24 through xor-folding?

erik-3milabs: you're right; I tried the lazy approach. Fixed it now :)

erik-3milabs:
@claude

RobinJadoul (approved):

PR #576, opened 2026-05-06
feat(cuda): math-cuda crate scaffolding — arith + NTT + coset LDE

ColoCarletti:
## Summary

  Introduces a new `math-cuda` crate that ships the foundation kernels for
  GPU-accelerated polynomial work: Goldilocks/ext3 field arithmetic,
  batched NTT (single-level + 8-level fused), and plain coset LDE entry
  points. Adds a `cuda` feature flag to `crypto/stark`, opt-in and disabled
  by default. No prover code consumes it yet — the kernels and dispatch
  surface land here so future PRs can wire them into Round 1 LDE/commit,
  Round 3 OOD evaluation, Round 4 DEEP/FRI, etc.

  The whole crate is gated. CPU-only consumers (`-p lambda-vm-prover`,
  `-p cli`) don't pull `math-cuda` in and don't require nvcc to build.

  ### What's in
  - New crate `crypto/math-cuda/` (~2.5k LoC):
    - `kernels/{goldilocks.cuh, ext3.cuh, arith.cu, ntt.cu}` — hand-written CUDA
    - `src/device.rs` — process-wide `Backend` singleton (32-stream pool,
      pinned host staging, twiddle cache, event-tracking disabled)
    - `src/ntt.rs` — host wrappers for forward/inverse NTT
    - `src/lde.rs` — `coset_lde_base`, `coset_lde_batch_base[_into]`,
      `coset_lde_batch_ext3_into`, `evaluate_poly_coset_batch_ext3_into[_keep]`
    - `GpuLdeBase` / `GpuLdeExt3` device-handle types — kept on device
      across calls so downstream consumers can read LDE buffers without
      re-paying PCIe transfer
  - 30 parity tests asserting **byte-identical u64 output** vs the CPU
    reference: arith, ext3, NTT round-trip, single-column LDE,
    batched LDE, ext3 LDE, coset evaluate.
  - Workspace wiring: `crypto/math-cuda` added to `[workspace] members`,
    `math-cuda = { path, optional = true }` + `cuda = ["dep:math-cuda"]`
    feature added to `crypto/stark`.

  ## Why this design

  - **Feature-gated, not replacement.** The `cuda` flag is opt-in. The CPU
    path is the default and untouched.
  - **Parity tests are the contract.** Every kernel produces bit-identical
    u64 output to the CPU reference. Tests assert on raw u64s, not on
    `FieldElement`s, so non-canonical representations must match exactly.
  - **Single backend, stream pool.** One CUDA context per process; a
    32-stream round-robin pool lets rayon-parallel callers overlap kernel
    launches. Twiddles are computed once on host and uploaded once per
    `log_n`.
  - **Device-resident LDE handles.** `GpuLdeBase` / `GpuLdeExt3` wrap an
    `Arc<CudaSlice<u64>>` so the device buffer survives past the call.
    This lets later passes read the LDE without H2D transferring it again
    (a 240 MB+ saving per round at prover scale).

  ## Verification

  | Check | Result |
  |---|---|
  | `cargo build --workspace --release` | ✓ |
  | `cargo build -p stark --features cuda --release` | ✓ |
  | `cargo build -p lambda-vm-prover --release` (no cuda, CPU consumer) | ✓ — does not pull `math-cuda` |
  | `cargo test -p math-cuda --release` | ✓ **30 / 30 parity tests pass** |

  Run on RTX 5090 (sm_120 / Blackwell), driver 595.58.03, CUDA 13.1
  toolchain. PTX targets `compute_89` (Ada virtual arch); the driver
  JIT-compiles for the actual GPU at module load, so the same PTX runs on
  Ada / Hopper / Blackwell.

  ### Microbench (`bench_quick.rs`, `--ignored`)

  | Workload | CPU rayon | GPU | Ratio |
  |---|---|---|---|
  | Single-col LDE 2^18, blowup=4 | 115 ms | 13.6 ms | 8.48× |
  | Batched 64 cols, log_n=16, blowup=4 | 100 ms | 13.0 ms | 7.73× |
  | Prover-scale 20 cols × 2^20, blowup=4 (median of 10) | 344 ms | 207 ms | **1.66×** |

  Single-shot timings (other than the median-of-10 row) are directional
  only.

  ## Test plan

  - [x] `cargo build --workspace --release` builds clean
  - [x] CPU-only consumer crates build without invoking nvcc
  - [x] `cargo build -p stark --features cuda` compiles cleanly
  - [x] Every `crypto/math-cuda/tests/*.rs` parity test passes (30 tests)
  - [x] `bench_quick` runs end-to-end on the GPU

  ## Notes for reviewers

  - **`cudarc 0.19` features**: `driver, nvrtc, std, cuda-12080, dynamic-loading`.
    `dynamic-loading` means `libcuda` is loaded at runtime, not linked, so
    the binary is portable across driver versions. `cuda-12080` selects the
    CUDA 12.8 ABI bindings; verified working against a CUDA 13.1 toolchain
    + driver 595.58.03.
  - **Workspace builds require nvcc**: because `crypto/math-cuda` is a
    workspace member, `cargo build --workspace` invokes its `build.rs`,
    which calls nvcc. Consumer-crate builds via `cargo build -p <crate>`
    for `lambda-vm-prover`/`cli` do **not** pull in `math-cuda`. CPU-only
    dev-loop: use `-p <crate>` instead of `--workspace`.
  - **Parity contract**: bit-identical u64 output to the CPU reference,
    not field-equal. Catches non-canonical-representation drift early.
  - **`GpuLdeBase` / `GpuLdeExt3` ship without direct test coverage** in
    this PR — they're constructed only via the `_keep` variants and consumed
    by integrations that aren't here yet. The byte-equivalence contract on
    the LDE buffers themselves is enforced by the existing parity tests.


crypto/math-cuda/src/device.rs
```diff
@@ -0,0 +1,248 @@
...
+        for _ in 0..STREAM_POOL_SIZE {
+            streams.push(ctx.new_stream()?);
+        }
+        let pinned_staging = Mutex::new(PinnedStaging::empty());
+        // Separate "utility" stream for twiddle uploads and other bookkeeping;
+        // not part of the pool that callers rotate through.
+        let util_stream = ctx.new_stream()?;
+
+        // Goldilocks TWO_ADICITY is 32, so log_n ≤ 32 covers every LDE size
+        // the prover can produce. Overshoot by one for safety.
```
gabrielbosio: I think the addition by one is not for safety but for correctness:


```suggestion
        // Cache is indexed by log_n. Valid range is [0, TWO_ADICITY] since
        // Goldilocks has roots of unity for orders 2^0..=2^TWO_ADICITY only.
        // Length = TWO_ADICITY + 1 to allow indexing at log_n = TWO_ADICITY.
```

MauroToscano (changes requested):
cudarc cuda-13010 feature is too strict (real bug). Every test panics on CUDA-13.0 drivers because cudarc tries to dlsym cuDevSmResourceSplit, which only
  exists in 13.1+. The driver here only exports cuDevSmResourceSplitByCount. Suggest cuda-13000 or feature-detect.

Default compute_89 PTX JIT-fails on Blackwell. Comment claims one PTX works across Ada/Hopper/Blackwell — empirically false with nvcc 13.0 + driver 580 +
  RTX 5090, which rejects the PTX as "unsupported toolchain". Setting CUDARC_NVCC_ARCH=sm_120 fixes it. Either emit a fat binary or detect arch in build.rs.

The static note about "build.rs empty-PTX stub fallback" is wrong. No fallback exists; cargo check --workspace panics on no-nvcc CI runners. Per-crate non-cuda builds still work (math-cuda is optional = true).

MauroToscano:
> cudarc cuda-13010 feature is too strict (real bug). Every test panics on CUDA-13.0 drivers because cudarc tries to dlsym cuDevSmResourceSplit, which only exists in 13.1+. The driver here only exports cuDevSmResourceSplitByCount. Suggest cuda-13000 or feature-detect.
> 
> Default compute_89 PTX JIT-fails on Blackwell. Comment claims one PTX works across Ada/Hopper/Blackwell — empirically false with nvcc 13.0 + driver 580 + RTX 5090, which rejects the PTX as "unsupported toolchain". Setting CUDARC_NVCC_ARCH=sm_120 fixes it. Either emit a fat binary or detect arch in build.rs.
> 
> The static note about "build.rs empty-PTX stub fallback" is wrong. No fallback exists; cargo check --workspace panics on no-nvcc CI runners. Per-crate non-cuda builds still work (math-cuda is optional = true).

Actually instead of setting it to 13000 check adding this:

  cudarc = { version = "0.19", default-features = false, features = [
      "driver",
      "nvrtc",
      "std",
      "cuda-version-from-build-system",  # auto-pick based on local nvcc
      "fallback-latest",                  # don't panic if nvcc missing
      "dynamic-loading",
  ] }

crypto/math-cuda/src/lde.rs
```diff
@@ -0,0 +1,991 @@
+//! Full coset LDE on device. Mirrors `Polynomial::coset_lde_full_expand` in
+//! `crypto/math/src/fft/polynomial.rs` algebraically:
+//!
+//! Input  : N evaluations (natural order) of a poly on the standard subgroup,
+//!          plus coset weights (size N). The weights include the `1/N` iFFT
+//!          normalisation, matching the `LdeTwiddles::coset_weights` format at
+//!          `crypto/stark/src/prover.rs:248` — i.e. `weights[i] = g^i / N`.
```
gabrielbosio: Nit:

```suggestion
//!          `crypto/stark/src/prover.rs:248`, i.e. `weights[i] = g^i / N`.
```

crypto/math-cuda/src/lde.rs
```diff
@@ -0,0 +1,991 @@
...
+    }
+    let lde_size = n * blowup_factor;
+    let log_n = n.trailing_zeros() as u64;
+    let log_lde = lde_size.trailing_zeros() as u64;
+
+    let be = backend();
+    let stream = be.next_stream();
+    let staging_slot = be.pinned_staging();
+
+    let debug_phases = std::env::var("MATH_CUDA_PHASE_TIMING").is_ok();
```
gabrielbosio: Is this needed on main or was it for debugging purposes?

crypto/math-cuda/src/lde.rs
```diff
@@ -0,0 +1,991 @@
...
+    // SAFETY: staging is locked, the slice alias ends before we unlock.
+    let pinned = unsafe { staging.as_mut_slice(m * lde_size) };
+    if debug_phases {
+        phase("staging lock + grow", &mut last);
+    }
+
+    // Pack columns into first m*n slots of the pinned buffer. Parallel: pinned
+    // writes are DRAM-bandwidth bound, saturates at ~8 cores on modern
+    // hardware, so rayon shaves 20+ ms at prover scale.
+    use rayon::prelude::*;
```
gabrielbosio: Why not moving this at the top?

crypto/math-cuda/src/lde.rs
```diff
@@ -0,0 +1,991 @@
...
+    // call to the driver, saturates PCIe.
+    stream.memcpy_dtoh(&buf, &mut pinned[..m * lde_size])?;
+    stream.synchronize()?;
+    if debug_phases {
+        phase("D2H (one shot into pinned)", &mut last);
+    }
+
+    // Split pinned → per-column Vec<u64>s. The first write to each virgin
+    // Vec page-faults, which can dominate total time (~75 ms for 128 MB).
+    // Parallelise so the fault cost spreads across CPU cores.
```
gabrielbosio:
```suggestion
    // Split pinned → per-column Vec<u64>s. The first write to each virgin
    // Vec page-faults, which can dominate total time. Parallelise so the
    // fault cost spreads across CPU cores.
```

crypto/math-cuda/src/lde.rs
```diff
@@ -0,0 +1,991 @@
...
+    let pinned_ptr = pinned.as_ptr() as usize;
+    let out: Vec<Vec<u64>> = (0..m)
+        .into_par_iter()
+        .map(|c| {
+            // set_len skips the O(N) zero-init that vec![0; n] would do
+            // (saves ~75 ms per 128 MB at prover scale). copy_from_slice
+            // below writes every slot before any reader sees the Vec.
```
gabrielbosio:
```suggestion
            // set_len skips the O(N) zero-init that vec![0; n] would do.
            // copy_from_slice below writes every slot before any reader
            // sees the Vec.
```

crypto/math-cuda/src/lde.rs
```diff
@@ -0,0 +1,991 @@
...
+    }
+    drop(staging);
+    Ok(out)
+}
+
+/// Like `coset_lde_batch_base` but writes directly into caller-provided
+/// output slices instead of allocating fresh `Vec<u64>`s. Each output slice
+/// must already have length `n * blowup_factor`. Saves ~50–100 ms of pageable
+/// allocator work + page faults at prover scale because the caller's Vecs
+/// have been sized once and are reused across calls.
```
gabrielbosio:
```suggestion
/// Like `coset_lde_batch_base` but writes directly into caller-provided
/// output slices instead of allocating fresh `Vec<u64>`s. Each output slice
/// must already have length `n * blowup_factor`. Avoids pageable allocator
/// work and page faults at prover scale because the caller's Vecs have been
/// sized once and are reused across calls.
```

crypto/math-cuda/src/lde.rs
```diff
@@ -0,0 +1,991 @@
...
+        m_u32,
+    )?;
+
+    stream.memcpy_dtoh(&buf, &mut pinned[..m * lde_size])?;
+    stream.synchronize()?;
+
+    // Parallel copy pinned → caller outputs. Caller's Vecs may still fault
+    // on first write; we spread that cost across rayon cores.
+    #[allow(unused_imports)]
+    use rayon::prelude::*;
```
gabrielbosio: Why not moving it to the top?

crypto/math-cuda/src/lde.rs
```diff
@@ -0,0 +1,991 @@
...
+    staging.ensure_capacity(m * lde_size, &be.ctx)?;
+    // SAFETY: staging is locked, the slice alias ends before we unlock.
+    let pinned = unsafe { staging.as_mut_slice(m * lde_size) };
+    if debug_phases {
+        phase("staging lock + grow", &mut last);
+    }
+
+    // Pack columns into first m*n slots of the pinned buffer. Parallel: pinned
+    // writes are DRAM-bandwidth bound, saturates at ~8 cores on modern
+    // hardware, so rayon shaves 20+ ms at prover scale.
```
gabrielbosio:
```suggestion
    // Pack columns into first m*n slots of the pinned buffer. Parallel:
    // pinned writes are DRAM-bandwidth bound, so rayon spreads the cost
    // across CPU cores.
```

crypto/math-cuda/tests/ext3_edge.rs
```diff
@@ -0,0 +1,176 @@
+//! Adversarial edge-case parity tests for `ext3::mul` on the GPU.
+//!
+//! The CUDA `dot3` (kernels/ext3.cuh:62-98) manually tracks overflow when
+//! summing three u128 products in split u64 hi/lo registers. The CPU
```
gabrielbosio:
```suggestion
//! The CUDA `dot3` in kernels/ext3.cuh manually tracks overflow when
//! summing three u128 products in split u64 hi/lo registers. The CPU
```

crypto/math-cuda/src/lde.rs
gabrielbosio: also I would remove the line referenced because it may change in the future:

```suggestion
//!          `crypto/stark/src/prover.rs` — i.e. `weights[i] = g^i / N`.
```

crypto/math-cuda/tests/ext3_sub.rs
```diff
@@ -0,0 +1,109 @@
+//! Parity test for `ext3::sub` (kernels/ext3.cuh:41-45). This device
```
gabrielbosio:
```suggestion
//! Parity test for `ext3::sub` in kernels/ext3.cuh. This device
```

crypto/math-cuda/tests/lde.rs
```diff
@@ -0,0 +1,116 @@
...
+use math::field::goldilocks::GoldilocksField;
+use math::field::traits::{IsField, IsPrimeField};
+use math::polynomial::Polynomial;
+use rand::{Rng, SeedableRng};
+use rand_chacha::ChaCha8Rng;
+
+type Fp = FieldElement<GoldilocksField>;
+
+/// Build the coset weights `[1/N, g/N, g²/N, …, g^{n-1}/N]` — this is the
+/// layout `crypto/stark/src/prover.rs:248` uses, with `1/N` pre-folded into the
```
gabrielbosio:
```suggestion
/// layout `crypto/stark/src/prover.rs` uses, with `1/N` pre-folded into the
```

crypto/math-cuda/tests/lde_batch_into.rs
```diff
@@ -0,0 +1,87 @@
+//! Direct parity test for `coset_lde_batch_base_into` (lde.rs:331), the
```
gabrielbosio:
```suggestion
//! Direct parity test for `coset_lde_batch_base_into` (lde.rs), the
```

crypto/math-cuda/tests/ntt.rs
```diff
@@ -0,0 +1,138 @@
...
+#[test]
+fn ntt_sizes_medium() {
+    for log_n in 11..=16 {
+        assert_ntt_match(log_n, 200 + log_n);
+    }
+}
+
+#[test]
+fn ntt_size_2_to_20() {
+    // The hot LDE size. One seed is enough; any mismatch screams loudly.
```
gabrielbosio: I would drop this comment

crypto/math-cuda/tests/ntt.rs
```diff
@@ -0,0 +1,138 @@
...
+    assert_ntt_match(20, 0xDEAD);
+}
+
+#[test]
+fn ntt_trivial_sizes() {
+    // Power-of-two below the interesting range — should still pass.
```
gabrielbosio: I would drop this comment

crypto/math-cuda/tests/ntt_known.rs
```diff
@@ -0,0 +1,125 @@
...
+
+type Fp = FieldElement<GoldilocksField>;
+
+fn canon(x: u64) -> u64 {
+    GoldilocksField::canonical(&x)
+}
+
+/// p(x) = 1 + x. NTT at size N is the vector [p(omega^0), p(omega^1), ...,
+/// p(omega^(N-1))] for `omega` a primitive N-th root of unity. We compute
+/// the reference by direct exponentiation of `omega` — no FFT involved —
```
gabrielbosio:
```suggestion
/// the reference by direct exponentiation of `omega` (no FFT involved)
```

crypto/math-cuda/tests/ntt_known.rs
```diff
@@ -0,0 +1,125 @@
...
+    // GoldilocksField uses bit-reversed coefficient layout for forward NTT:
+    // input coeffs at index `i` become `p(omega^bitrev(i))` — confirm by
+    // matching against the existing forward() API which random-input tests
+    // already validate against `Polynomial::evaluate_fft`. The known-poly
+    // value lets us catch systematic errors in that pipeline that random
+    // inputs miss.
+
+    // Input coefficients: [1, 1, 0, 0, ..., 0]. (Natural order, lowest
+    // degree first — `Polynomial::new` and `math_cuda::ntt::forward` both
+    // expect this convention.)
```
gabrielbosio, line 36:
```suggestion
    // GoldilocksField uses bit-reversed coefficient layout for forward NTT:
    // input coeffs at index `i` become `p(omega^bitrev(i))`. Confirm by
    // matching against the existing forward() API which random-input tests
    // already validate against `Polynomial::evaluate_fft`. The known-poly
    // value lets us catch systematic errors in that pipeline that random
    // inputs miss.

    // Input coefficients: [1, 1, 0, 0, ..., 0]. (Natural order, lowest
    // degree first. `Polynomial::new` and `math_cuda::ntt::forward` both
    // expect this convention.)
```

crypto/math-cuda/build.rs
```diff
@@ -0,0 +1,78 @@
...
+    println!("cargo:rerun-if-changed=kernels/{src}");
+    println!("cargo:rerun-if-env-changed=CUDA_HOME");
+    println!("cargo:rerun-if-env-changed=CUDA_PATH");
+    println!("cargo:rerun-if-env-changed=CUDARC_NVCC_ARCH");
+
+    // No nvcc on PATH → emit an empty PTX stub so the crate still compiles.
+    // include_str! in src/device.rs needs the file to exist at build time.
+    // Any runtime kernel call will then panic from cudarc when loading the
+    // empty module — which is the right failure mode (we can't run GPU code
+    // without nvcc on the build host anyway).
```
gabrielbosio:
```suggestion
    // When nvcc is missing from PATH, emit an empty PTX stub so the crate
    // still compiles. include_str! in src/device.rs needs the file to exist
    // at build time. Any runtime kernel call panics in cudarc when loading
    // the empty module. We can't run GPU code without nvcc on the build
    // host anyway.
```

crypto/math-cuda/build.rs
```diff
@@ -0,0 +1,78 @@
...
+    if !have_nvcc {
+        fs::write(&out_path, "").expect("failed to write empty PTX stub");
+        return;
+    }
+
+    // Emit PTX for a virtual architecture; the CUDA driver JIT-compiles it for the
+    // actual GPU at load time, so one PTX works across Ada/Hopper/Blackwell. Override
+    // with CUDARC_NVCC_ARCH to pin a specific compute capability.
```
gabrielbosio:
```suggestion
    // Emit PTX for a virtual architecture, which the CUDA driver JIT-compiles
    // for the actual GPU at load time, so one PTX works across
    // Ada/Hopper/Blackwell. Override with CUDARC_NVCC_ARCH to pin a specific
    // compute capability.
```

crypto/math-cuda/build.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+fn main() {
+    // Headers are not compiled; emit rerun-if-changed so edits trigger rebuilds.
+    println!("cargo:rerun-if-changed=kernels/goldilocks.cuh");
+    println!("cargo:rerun-if-changed=kernels/ext3.cuh");
+
+    // Probe for nvcc once. Workspace consumers (clippy, fmt, CPU-only test
+    // runners) build math-cuda incidentally without using its kernels; allow
+    // those to succeed by stubbing out PTX when nvcc is unavailable.
```
gabrielbosio:
```suggestion
    // Headers aren't compiled, so emit rerun-if-changed to rebuild on
    // header edits.
    println!("cargo:rerun-if-changed=kernels/goldilocks.cuh");
    println!("cargo:rerun-if-changed=kernels/ext3.cuh");

    // Probe for nvcc once. Workspace consumers (clippy, fmt, CPU-only test
    // runners) build math-cuda incidentally without using its kernels. Stub
    // out PTX when nvcc is unavailable so those builds succeed.
```

gabrielbosio (commented):
It would be nice if this PR adds some make commands to run tests from `math-cuda`. #575 already has those.

crypto/math-cuda/kernels/ntt.cu
```diff
@@ -0,0 +1,287 @@
...
+
+/// Up to 8 DIT butterfly levels fused in one kernel using shared memory.
+///
+/// Ported from Zisk's `br_ntt_8_steps` (`pil2-stark/src/goldilocks/src/ntt_goldilocks.cu`),
+/// simplified to single-column. Each block of 256 threads processes 256
+/// elements in on-chip shared memory, running up to 8 butterfly levels
+/// without writing to global memory between them — cuts DRAM traffic by up
+/// to 8× vs the per-level kernel.
+///
+/// `base_step` selects which 8-level window this launch handles (0, 8, 16…).
```
gabrielbosio: Nit:

```suggestion
/// `base_step` selects which 8-level window this launch handles (0, 8, 16, ...).
```

crypto/math-cuda/src/ntt.rs
```diff
@@ -0,0 +1,205 @@
...
+
+use cudarc::driver::{LaunchConfig, PushKernelArg};
+use math::field::element::FieldElement;
+use math::field::goldilocks::GoldilocksField;
+use math::field::traits::{IsFFTField, IsField};
+
+use crate::Result;
+use crate::device::backend;
+
+/// Host-side twiddle table: `[ω^0, ω^1, …, ω^{n/2-1}]` where ω is the
```
gabrielbosio:
```suggestion
/// Host-side twiddle table: `[ω^0, ω^1, ..., ω^{n/2-1}]` where ω is the
```

crypto/math-cuda/tests/lde.rs
```diff
@@ -0,0 +1,116 @@
...
+use math::field::element::FieldElement;
+use math::field::goldilocks::GoldilocksField;
+use math::field::traits::{IsField, IsPrimeField};
+use math::polynomial::Polynomial;
+use rand::{Rng, SeedableRng};
+use rand_chacha::ChaCha8Rng;
+
+type Fp = FieldElement<GoldilocksField>;
+
+/// Build the coset weights `[1/N, g/N, g²/N, …, g^{n-1}/N]` — this is the
```
gabrielbosio:
```suggestion
/// Build the coset weights `[1/N, g/N, g²/N, ..., g^{n-1}/N]` — this is the
```

ColoCarletti: Resolevd in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/src/ntt.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/kernels/ntt.cu
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/build.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/build.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/build.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/tests/ntt_known.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/tests/ntt_known.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/tests/ntt.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/tests/ntt.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/tests/lde_batch_into.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/src/lde.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/src/lde.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/src/lde.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/src/lde.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/src/lde.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/src/lde.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/tests/ext3_sub.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/tests/lde.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/tests/ext3_edge.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/src/lde.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

crypto/math-cuda/src/lde.rs
ColoCarletti: Resolved in https://github.com/yetanotherco/lambda_vm/pull/576/changes/8c529718ee6b8f0ae4287c3cab39b3dd552783dc

MauroToscano (approved):

gabrielbosio (approved):

PR #585, opened 2026-05-13
docs: refresh READMEs

ColoCarletti:


README.md
gabrielbosio: Nit:

```suggestion
##### Compile them directly
```

prover/README.md
```diff
@@ -0,0 +1,54 @@
+# Lambda VM Prover
+
+STARK prover for the Lambda VM. Proves correct execution of RISC-V ELF binaries by generating a multi-table STARK proof (CPU, decoder, bitwise, LT, memory, load, MUL, DVRM, shift, branch, Keccak, …) and provides the matching native verifier.
```
gabrielbosio: Table names are out of sync with `spec/README.md`

spec/README.md
```diff
@@ -1,11 +1,19 @@
...
-This repository contains specification for [`LambdaVM`](https://github.com/yetanotherco/lambda_vm).
-The specification is written in [`Typst`](https://typst.app/) and can be rendered by [`shiroa`](https://myriad-dreamin.github.io/shiroa/) as either a file (pdf) or a wiki (html).
+# Lambda VM Specification
 
-## Installation & Development setup
-1. [Install `Typst`](https://github.com/typst/typst?tab=readme-ov-file#installation).
-2. [Install `shiroa`](https://myriad-dreamin.github.io/shiroa/guide/installation.html).
-3. Clone this repository.
-4. Open the repository in a terminal and execute `shiroa serve`.
+Formal specification of the Lambda VM. Covers the per-chip AIR constraints (CPU, decode, ALU, bitwise, branch, memory, MEMW, LOAD, MUL, DVRM, shift, LT, HALT, COMMIT, …), the memory argument, and the LogUp lookup framework that links the tables.
```
gabrielbosio: Table names are out of sync with `prover/README.md`

executor/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+let elf_bytes = std::fs::read("program.elf")?;
+let program = Elf::load(&elf_bytes)?;
+let executor = Executor::new(&program, /* private input */ vec![])?;
+let result = executor.run()?;
+
+println!("Executed {} instructions", result.logs.len());
+```
+
+For chunked execution (useful when you don't want to hold all logs in memory), drive the executor via `executor.resume()` in a loop until it returns `None`, then call `executor.finish()`. See [`bin/cli/src/main.rs`](../bin/cli/src/main.rs) for an example.
```
gabrielbosio: `executor.resume()` returns a `Result<Option>` not an `Option`

```suggestion
For chunked execution (useful when you don't want to hold all logs in memory), drive the executor via `executor.resume()` in a loop until it yields `None`, then call `executor.finish()`. See [`bin/cli/src/main.rs`](../bin/cli/src/main.rs) for an example.
```

README.md
gabrielbosio: Remove toolchain install command execution since `make compile-programs-rust` already does that. Also, match toolchain version used in Makefile:

```suggestion
rustup component add rust-src --toolchain nightly-2026-02-01
```

gabrielbosio: Right now, `make deps` is outdated. It would be nice if it downloads pre-installed libraries or it installs and compiles them depending on the OS.

gabrielbosio, line 171: I would add an entry for `make test-math-cuda`

README.md
```diff
@@ -98,7 +142,18 @@ Following [ethrex](https://github.com/lambdaclass/ethrex):
 
 ## Documentation
 
-Full documentation can be found in [docs](./docs/). It is currently a work in progress, we expect that as more features and components become ready, they will be included in the docs.
+High-level documentation lives in [`docs/`](./docs/):
+
+- [Overview of VM flow](./docs/general_flow.md) — the pipeline from source code to proof
```
gabrielbosio, line 147: We can change `general_flow.md` to be more specific:

- No mention of the actual ISA (RV64IM) — only the README states this.
- No mention of the proof system family (STARK + FRI).
- No pointers to the corresponding crates/dirs, so a reader can't bridge from the diagram to the code.
- Uses "arithmetisation" / "constraint system" generically without naming AIR + LogUp lookups, which are the actual choices.
- Doesn't mention precompiles/syscalls (Keccak landed in #519), even though they now affect both the executor and the witness.

README.md
```diff
@@ -98,7 +142,18 @@ Following [ethrex](https://github.com/lambdaclass/ethrex):
...
+- [Proof system overview](./docs/cryptography/proof_system.md) — design goals and primitives
```
gabrielbosio, line 148: Sub-sections of `proof_system.md` are empty. We can address this later in another PR but at least we should mention that is a WIP.

README.md
```diff
@@ -98,7 +142,18 @@ Following [ethrex](https://github.com/lambdaclass/ethrex):
...
+- [Lookup arguments](./docs/cryptography/lookup.md) — how tables are linked
```
gabrielbosio: `lookup.md` is outdated:

1. `TableInteraction` struct doesn't exist. Replaced by BusInteraction at crypto/stark/src/lookup.rs:1312, with a completely different
shape:

    ```rust
    pub struct BusInteraction {
        pub bus_id: u64,
        pub multiplicity: Multiplicity,
        pub values: Vec<BusValue>,
        pub is_sender: bool,
    }
    ```
    
2. The doc's flag_columns: Vec<usize> / value_columns: Vec<usize> is from an earlier API.
3. `LOGUP_CHALLENGE_Z` is gone. Only `LOGUP_CHALLENGE_ALPHA = 1` and `LOGUP_NUM_CHALLENGES = 2` exist (crypto/stark/src/lookup.rs:100). Code
reads `z` as `challenges[0]` directly.
4. No mention of `bus_id`. The current system supports multiple named buses (an enum like `BusId { Add, Mul, Sub }`) with
senders/receivers matched per bus, central to the design now, absent from the doc.
5. Two-stage value combination missing. The doc shows one linear combination. The code does:
- `BusValue::Packed { packing, .. }`: combine limbs within a value using powers-of-2 or custom coefficients
- then bus fingerprint with powers of alpha over those bus elements
6. Multiplicity is now an enum `(One, Column, Sum, Negated, Diff, Sum3, Linear)`, not a positive/negative number. Sign is carried by
`is_sender: bool` on the interaction.
7. BusValue types (`Packed { packing, .. }`, `Linear(Vec<LinearTerm>)`) not documented.
8. Bus balance is written as `sum(sender_values) − sum(receiver_values) = 0` (line 1324) rather than the doc's `sum(sends) + sum(receives) = 0`.

README.md
gabrielbosio, line 218: The make command creates a flamegraph of the execution of the prover given what program?

gabrielbosio: We could also add the `make bench-math-cuda`

gabrielbosio (commented):
`executor/README.md` and `syscalls/README.md` should be accessible through root README

syscalls/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+```rust
+use lambda_vm_syscalls as syscalls;
+
+pub fn main() {
+    let input: Vec<u8> = syscalls::syscalls::get_private_input();
+    syscalls::syscalls::print_string(&format!("Received {} bytes\n", input.len()));
+    syscalls::syscalls::commit(&input);
+}
+```
```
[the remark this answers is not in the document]
ColoCarletti: Fixed in: https://github.com/yetanotherco/lambda_vm/pull/585/commits/2bc6fe23df9f4eead5b6e5811610b44bac35422e

README.md
ColoCarletti: Fixed in: https://github.com/yetanotherco/lambda_vm/pull/585/commits/2bc6fe23df9f4eead5b6e5811610b44bac35422e

prover/README.md
ColoCarletti: Fixed in: https://github.com/yetanotherco/lambda_vm/pull/585/commits/2bc6fe23df9f4eead5b6e5811610b44bac35422e

spec/README.md
ColoCarletti: Fixed in: https://github.com/yetanotherco/lambda_vm/pull/585/commits/2bc6fe23df9f4eead5b6e5811610b44bac35422e

executor/README.md
ColoCarletti: Fixed in: https://github.com/yetanotherco/lambda_vm/pull/585/commits/2bc6fe23df9f4eead5b6e5811610b44bac35422e

README.md
ColoCarletti: Fixed.

README.md
ColoCarletti: Fixed.

README.md
ColoCarletti: Fixed.

README.md
ColoCarletti: Fixed

ColoCarletti: I've removed the make deps references from the README; making it cross-platform will be tracked as a separate follow-up

ColoCarletti: Fixed.

README.md
ColoCarletti: Tracking as a follow-up

README.md
gabrielbosio: tracked in #586

gabrielbosio (approved):

PR #589, opened 2026-05-18
Test/ef io edge cases

MauroToscano:


MauroToscano:
Closing as stale — opened in May, now CONFLICTING, and its core (the `ef_io` interface + `ef_io_demo`) already landed via #579.

On the edge-case coverage: two of the three programs here are already covered on `main` — `ef_io_no_input` overlaps the existing "`prove_with_inputs` with empty input must match `prove`" backward-compat test, and `ef_io_multi_write` overlaps `ef_io_demo`'s multi-call write-concatenation tests (`test_prove_ef_io_demo_concatenates` / `test_ef_io_demo_concatenates_writes`). The one case that is **not** covered is `ef_io_zero_write` (a zero-length `write_output` as a no-op) — the demo test feeds a 12-byte input so it never exercises a zero-length write. Worth re-adding that one as a small standalone test against current `main` if we want the corner covered.

PR #603, opened 2026-05-21
cleanup(merkle-tree): reject non-power-of-two leaves, dedupe helpers

diegokingston:
Behavior-preserving code-quality pass on `crypto/crypto/src/merkle_tree` (Poseidon backends, FriMerkleTree/Keccak256Backend, and the BatchProof / arity-4 machinery were explicitly kept).

Padding removed (the one real footgun)
- `build_from_hashed_leaves` padded a non-power-of-two leaf set by repeating the last leaf. That makes the root of `[.., x]` collide with the root of `[.., x, x]` — the root stopped binding the leaf count. Every prover commitment is already over a power-of-two domain, so the padding never fired in production. `build`/`build_from_hashed_leaves` now return `None` for a non-power-of-two count; `complete_until_power_of_two` is deleted.

Helper dedup / std reuse
- `utils.rs` had two index-helper pairs computing the same thing: `sibling_index`/`parent_index` (underflow at the root) and the root-safe `get_sibling_pos`/`get_parent_pos`. Consolidated onto the root-safe pair; `build_merkle_path` updated.
- Deleted the custom `is_power_of_two` (reinvented `usize::is_power_of_two`, and underflowed at 0).
- `utils` is now `pub(crate)` — internal plumbing, not public API.

Dead code
- Removed `Proof`'s hand-rolled `Serializable`/`Deserializable`: the live path is the `#[derive(serde::*)]` on `Proof` + bincode, these had zero callers, and the `Deserializable` impl's hardcoded `chunks(8)` was wrong for the actual 32-byte node type.
- Inlined `create_proof` (a wrapper that always returned `Some`).

Misc
- Pair/vector backends finalize digests with `.finalize().into()`, like `FieldElementBackend` already did (was `[0u8; N]` + `copy_from_slice`).
- Comment fixes (banner comments -> doc comments, typos).
- Tests updated to power-of-two inputs; the `odd_set_of_leaves` test now asserts the `None` rejection.

44 crypto + 124 stark lib tests pass; lint clean; disk-spill builds.

### DIFF
--- a/crypto/crypto/src/merkle_tree/backends/field_element.rs
+++ b/crypto/crypto/src/merkle_tree/backends/field_element.rs
@@ -84,7 +84,7 @@ mod tests {
 
     #[test]
     fn hash_data_field_element_backend_works_with_keccak_256() {
-        let values: Vec<FE> = (1..6).map(FE::from).collect();
+        let values: Vec<FE> = (1..9).map(FE::from).collect();
         let merkle_tree =
             MerkleTree::<FieldElementBackend<F, Keccak256, 32>>::build(&values).unwrap();
         let proof = merkle_tree.get_proof_by_pos(0).unwrap();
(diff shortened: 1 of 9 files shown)


MauroToscano (changes requested):
This one we should be careful, the padding is fine for the merkle tree, and we may need the serde serialization. The double serialization has the purpose of being able to serialize in binary modes or text modes

PR #610, opened 2026-05-22
refactor(prover): bind the proven statement into the Fiat-Shamir transcript

diegokingston:
## Summary

The VM prover and verifier built their Fiat-Shamir transcript with
`DefaultTranscript::new(&[])` — **empty**. The statement being proven (the
program, its public output, the table layout) was never absorbed, so the
verifier's challenges did not depend on it. Soundness was preserved only by
*out-of-band* reconstruction checks in the verifier (recomputed preprocessed
commitments, recomputed `expected_bus_balance`, `table_counts` cross-check).

This binds the statement directly into the transcript, so soundness no longer
depends on the completeness of that set of checks.

## What it does

- New `prover/src/statement.rs`: a canonical, **versioned**, **length-prefixed**,
  domain-tagged encoder that digests the full statement — ELF hash, public
  output, all `table_counts` fields, and page layout — into a 32-byte
  `statement_seed`.
- Seeds the transcript with it at **all three** construction sites that must
  agree: the prove path, the verify path, and the bus-balance replay
  (`replay_transcript_phase_a`). For an honest proof every site derives
  identical challenges; any tampered statement field makes the verifier's seed
  diverge and verification rejects.
- `tiny-keccak` is promoted from a dev-dependency to a dependency (already in
  the lockfile) for the statement digest.

## Severity

This is **hardening / defense-in-depth**, not a patch for a live forgery — the
statement is currently bound out-of-band. The value is making the soundness
argument local and self-evident, and it is a **prerequisite for recursion** (a
recursive verifier derives challenges purely from the transcript). Peer systems
(Plonky3, SP1, ZisK) all bind the statement into the transcript.

Breaking proof-format change: challenges differ, so old proofs will not verify.

## Test plan

- [x] `prover/src/tests/statement_tests.rs` — 8 encoder unit tests (determinism,
      per-field distinctness, domain tag, length-prefix boundary collision). All pass.
- [x] No regressions: full `prover` lib suite is 281 passed / 78 failed vs a
      baseline of 273 / 77 — i.e. **+8 passing** (the new encoder tests) and
      **+1 failing**, which is the new e2e test below.
- [ ] **`proof_binds_the_program_it_was_generated_for`** (e2e: honest roundtrip
      + wrong-program rejection; this is what actually exercises seed
      consistency across the three sites). It needs the compiled program
      artifacts (`make compile-programs-rust`). **It could not be run in the
      authoring environment** — no RISC-V toolchain, and the checked-in asm
      fixtures are stale (`sub.elf` uses a removed syscall, which already breaks
      several pre-existing e2e tests). **Please confirm this test passes in CI.**

### DIFF
--- a/prover/Cargo.toml
+++ b/prover/Cargo.toml
@@ -19,6 +19,7 @@ serde = { version = "1.0", features = ["derive"] }
 rayon = { version = "1.8.0", optional = true }
 sysinfo = { version = "0.31", default-features = false, features = ["system"] }
 log = "0.4"
+tiny-keccak = { version = "2.0", features = ["keccak"] }
 
 [dev-dependencies]
 env_logger = "*"
@@ -26,7 +27,6 @@ criterion = { version = "0.5", default-features = false }
 bincode = "1"
 tikv-jemallocator = "0.6"
 tikv-jemalloc-ctl = { version = "0.6", features = ["stats"] }
-tiny-keccak = { version = "2.0", features = ["keccak"] }
 # Enable stark's test-utils so cross-crate tests can reach
 # `compute_precomputed_commitment_for_testing`. Only active under cargo test/bench.
 stark = { path = "../crypto/stark", features = ["test-utils"] }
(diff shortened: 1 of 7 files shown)


MauroToscano:
I streamlined it here: https://github.com/yetanotherco/lambda_vm/pull/61

MauroToscano:
https://github.com/yetanotherco/lambda_vm/pull/617

PR #624, opened 2026-05-26
spec: rework CPU for smaller footprint

RobinJadoul:


spec/decode.typ
```diff
@@ -34,33 +34,50 @@ Empty rows with the following content can be added to achieve this:
 
 #render_chip_padding_table(chip, config)
 
-Note that this row sets the `EBREAK` flag.
-Given that `CPU` asserts that `EBREAK = 0` (see @cpu:c:ebreak_traps), using this "padding-instruction" would immediately make the CPU table unprovable.
-Note moreover that the `pc` is set to $7$.
-This value is the _smallest odd number_ (i.e., not reachable during regular execution) that is more than _$4$_ (i.e., the max `pc`-increment) greater than _$1$_ (i.e., the `pc`-value used in the #link(<cpu-padding-decode-row>)[additional instruction] referred to by `CPU`-padding lines).
+This is simultaneously the row that is used for padding rows in the CPU,
+if the multiplicity if nonzero,
```
erik-3milabs:
```suggestion
if the multiplicity is nonzero,
```

spec/src/decode_uncompressed.toml
```diff
@@ -35,131 +35,130 @@ name = "write_register"
 type = "Bit"
 desc = "whether the result should be written to `rd` ($=0$ for memory write and when $#`rd` = #`x0`$."
```
erik-3milabs:
```suggestion
desc = "whether the result should be written to `rd` ($=0$ for memory write and when $#`rd` = #`x0`)$."
```

spec/src/decode_uncompressed.toml
```diff
@@ -35,131 +35,130 @@ name = "write_register"
...
+name = "instruction_length"
+type = "Byte"
+desc = "How many bytes is this instruction in the program"
 
-[[variables.output]]
-name = "SHIFT"
+[[variables.auxiliary]]
+name = "alu_op"
+type = "Byte"
+desc = "Operation selector value for the ALU"
```
erik-3milabs, line 86: This being a `Byte` suggests weird things might happen in `alu_flags`. Should it be assigned a smaller type? Something like `B4`?

spec/cpu.typ
```suggestion
It bases its decisions on the entry of the `DECODE` table (@decode) corresponding to the current program counter (PC).
```

spec/src/cpu.toml
```diff
@@ -235,65 +146,41 @@ pad = 0
...
-name = "is_equal"
-type = "Bit"
-desc = "Whether `rv1` and `arg2` are equal"
-pad = 0
-
 [[variables.auxiliary]]
 name = "branch_cond"
 type = "Bit"
-desc = "Whether a branch is taken, i.e., the branch condition"
+desc = "Whether a branch is taken, the branch condition evaluates to true, or we are doing an unconditional jump"
```
```suggestion
desc = "Whether a branch is taken: the branch condition evaluates to true, or we are doing an unconditional jump"
```

spec/src/cpu.toml
```diff
@@ -235,65 +146,41 @@ pad = 0
...
 pad = 0
 
 # Virtual
+[[variables.virtual]]
+name = "JALR"
+type = "Bit"
+desc = "Read whether our BRANCH corresponds to a JAL(R) instruction from `mem_flags`, as `MEMORY` and `BRANCH` are mutually exclusive"
+def = "mem_flags"
```
erik-3milabs, line 182: `mem_flags: Byte`, which suggests `JALR`s definition is incorrect.

spec/cpu.typ
```diff
@@ -31,35 +31,49 @@ The #cpu chip is comprised of #nr_variables variables that are expressed using #
 #render_chip_assumptions(chip, config)
 
 = Constraints
+
 First, we perform a decoding lookup for the current PC.
+Instructions having the `word_instr` flag set are not decoded here, as they are delegated to the `CPU32` chip.
+In that case, we ensure that the current row of the CPU cannot have any other observable effects.
 
 #render_constraint_table(chip, config, groups: "decode")
```
erik-3milabs, line 44: Perhaps this should be put under its own subsubsection?

spec/cpu.typ
```diff
@@ -31,35 +31,49 @@ The #cpu chip is comprised of #nr_variables variables that are expressed using #
...
-The flags and register indices looked up from the decoding need to be checked,
-as they are communicated through the interaction in a packed form.
+All values in `packed_decode` need to be checked to ensure
+the packing is correct for the interaction.
 In contrast, we know ahead of time that decoding will ensure proper range checks for `pc` and `imm`.
 Similarly, since `next_pc` will propagate through the memory argument and be looked up
-in the instruction decoding on the next cycle, it is forced to be in the correct range.#rj[is this true, do we need this elsewhere for chip assumptions?]
-For the auxiliary columns, we need to check the limbs of `arg1`, `arg2`, and `res`.
+in the instruction decoding on the next cycle, it is forced to be in the correct range.
+The final value for `next_pc` is similarly fixed by the memory finalization.
```
```suggestion
Similarly, since `next_pc` will propagate through the memory argument and be looked up
in the instruction decoding on the next cycle, it is forced to be in the correct range;
the final value for `next_pc` is fixed by the memory finalization.
```

spec/src/cpu.toml
```diff
@@ -538,114 +336,55 @@ tag = "IS_BYTE"
...
 prefix = "A"
 
 [[constraints.alu]]
-kind = "template"
-tag = "ADD"
-cond = ["+", "ADD", "LOAD"]
-input = [["cast", "arg1", "DWordWL"], ["cast", "arg2", "DWordWL"]]
-output = ["cast", "res", "DWordWL"]
+kind = "arith"
+constraint = "$#`arg2` = #`MEMORY` * #`imm` + #`BRANCH` * (1 - #`JALR`) * #`rv2` * #`BRANCH` * #`JALR` * #`instruction_length` + (1 - #`MEMORY`) * (1 - #`BRANCH`) * (#`rv2` + #`imm`)$"
```
```suggestion
constraint = "$#`arg2` = #`MEMORY` dot #`imm` + #`BRANCH` dot (1 - #`JALR`) dot #`rv2` + #`BRANCH` dot #`JALR` dot #`instruction_length` + (1 - #`MEMORY`) dot (1 - #`BRANCH`) dot (#`rv2` + #`imm`)$"
```
Note here the important change from * to + !

spec/src/cpu.toml
```diff
@@ -538,114 +336,55 @@ tag = "IS_BYTE"
...
+poly = ["-", ["idx", "arg2", "i"],
+    ["*", "MEMORY", ["idx", "imm", "i"]],
+    ["*", "BRANCH", ["not", "JALR"], ["idx", "rv2", "i"]],
+    ["*", "BRANCH", "JALR", ["idx", ["arr", "instruction_length", 0], "i"]],
+    ["*", ["not", "MEMORY"], ["not", "BRANCH"], ["idx", ["+", "rv2", "imm"], "i"]]
```
erik-3milabs: should this not be
```suggestion
    ["*", ["not", "MEMORY"], ["not", "BRANCH"], ["+", ["idx", "rv2", "i"], ["idx", "imm", "i"]]
```
?
I don't recall the element-wise addition of arrays being defined before.

spec/cpu.typ
```diff
@@ -70,32 +84,36 @@ Constraints regarding whether `pc_double_read` corresponds to an `AUIPC` instruc
...
-- `double_pc_read` could be integrated into decoding, so that `AUIPC` could set `read_register1 = 0` and no extra MEMW access for `rv1` is needed at this point.
+The memory interaction itself is handled by the `MEMORY` signature,
+which will read the `mem_flags` argument to perform either a `LOAD` or a `STORE`.
+We refer to the previous section's description of `arg2` for how
+the address is computed.
 
-== System
+The value to (potentially) be written back to `rd` is stored in `rvd`,
+which can either come from either the ALU, in case of an ALU operation or a JALR branch,
+or from the MEMORY interaction.
```
erik-3milabs, line 96:
```suggestion
The value to (potentially) be written back to `rd` is stored in `rvd`,
which can either come from the ALU --- in case of an ALU operation or a JALR branch ---
or from the MEMORY interaction.
```

spec/src/cpu.toml
```diff
@@ -687,16 +426,16 @@ multiplicity = "write_register"
...
-multiplicity = "LOAD"
+multiplicity = "MEMORY"
 
 [[constraints.mem]]
-kind = "interaction"
-tag = "MEMW"
-input = [0, ["cast", "res", "DWordWL"], ["cast", "arg2", ["Byte", 8]], ["+", "timestamp", ["cast", 1, "DWordWL"]], "memory_2bytes", "memory_4bytes", "memory_8bytes"]
-multiplicity = "STORE"
+kind = "arith"
+constraint = "$!#`MEMORY` => #`rvd` = #`res`$"
```
erik-3milabs: consistency
```suggestion
constraint = "$#`!MEMORY` => #`rvd` = #`res`$"
```

spec/src/cpu.toml
```diff
@@ -712,124 +451,50 @@ input = ["prev_pc_timestamp_borrow"]
 kind = "interaction"
 tag = "memory"
 input = [1, ["arr", ["+", ["*", 2, 255], "i"], 0], ["arr", ["+", ["-", ["idx", "timestamp", 0], ["*", 3, ["not", "pc_double_read"]]], ["*", ["^", 2, 32], "prev_pc_timestamp_borrow"]], ["-", ["idx", "timestamp", 1], "prev_pc_timestamp_borrow"]], ["idx", "pc", "i"]]
-multiplicity = ["not", "pad"]
 iter = ["i", 0, 1]
+multiplicity = 1
 
 [[constraints.mem]]
 kind = "interaction"
 tag = "memory"
```
erik-3milabs, line 497: the existence of signatures with both 'MEMORY' and 'memory' as tag is confusing. I propose we rename one of them.

spec/src/cpu.toml
```diff
@@ -712,124 +451,50 @@ input = ["prev_pc_timestamp_borrow"]
...
 multiplicity = "branch_cond"
 
-[[constraints.misc]]
+[[constraints.branch]]
 kind = "template"
 tag = "ADD"
-input = ["pc", ["*", ["+", ["*", 2, "c_type_instruction"], ["*", 4, ["not", "c_type_instruction"]]], ["cast", 1, "DWordWL"]]]
+input = ["pc", ["arr", "instruction_length", 0]]
 output = "next_pc"
 desc = "Increment `pc` to `next_pc` if we're not branching"
```
erik-3milabs, line 528: Are we missing a condition on `1-branch_cond` here?

spec/src/cpu.toml
```diff
@@ -353,7 +220,22 @@ name = "decode"
...
+
+[[constraints.decode]]
+kind = "arith"
+constraint = "$#`word_instr` => #`BRANCH = 0`$"
+poly = ["*", "word_instr", "BRANCH"]
+
+[[constraints.decode]]
+kind = "arith"
+constraint = "$#`word_instr` => #`ECALL = 0`$"
+poly = ["*", "word_instr", "ECALL"]
```
erik-3milabs, line 261: do we additionally want to constrain that `word_instr => read_registerx = 0` (and also `write_register = 0`)? Technically, these constraints would be superfluous (in case of `word_instr`, the variables `read_register1`, `2` and `write_register` are free to be chosen, and setting them to 1 will lead to competing memory accesses in CPU and CPU32 => unprovable), but adding them might increase the "safety margin" a bit?

spec/src/cpu32.toml
```diff
@@ -0,0 +1,402 @@
...
+name = "SUB"
+type = "Bit"
+desc = "Whether the full ALU is active"
+pad = 0
+
+[[variables.auxiliary]]
+name = "instruction_length"
+type = "Byte"
+desc = "The length of this instruction"
+pad = 4
```
erik-3milabs: Perhaps this should be an `input` rather than `aux`? It is passed in through the interaction as such.

spec/cpu32.typ
```diff
@@ -0,0 +1,56 @@
...
+#let chip = load_chip("src/cpu32.toml", config)
+
+#show: book-page(chip.name)
+#let cpu32 = raw(chip.name)
+
+The #cpu32 chip is used to delegate the 32-bit instructions of the RV64I instruction set
+from the main CPU table (@cpu).
+All 32-bit instructions are ALU-only instructions, so the BRANCH, MEMORY and ECALL paths need no elaboration.
+The timestamp and PC have already been read by the CPU table at this point, and need no further checking,
+the PC for the next instruction will also already be handled by CPU.
```
erik-3milabs:
```suggestion
The timestamp and PC have already been read by the CPU table at this point, and need no further checking;
the PC for the next instruction will also already be handled by CPU.
```

spec/src/cpu32.toml
```diff
@@ -0,0 +1,402 @@
...
+
+[[constraints.ext]]
+kind = "arith"
+constraint = "$#`arg1[0]` = #`rv1[:2]`$"
+poly = ["-", ["idx", "arg1", 0], ["idx", ["cast", "rv1", "DWordWL"], 0]]
+
+[[constraints.ext]]
+kind = "arith"
+constraint = "$#`arg1[1]` = (2^(32) - 1) #`signed` dot #`rv1_sign`$"
+poly = ["-", ["idx", "arg1", 1], ["*", ["-", ["^", 2, 32], 1], "signed", "rv1_sign"]]
```
erik-3milabs:
```suggestion
[[constraints.ext]]
kind = "arith"
constraint = "$#`arg1[1]` = (2^(32) - 1) dot #`signed` dot #`rv1_sign`$"
poly = ["-", ["idx", "arg1", 1], ["*", ["-", ["^", 2, 32], 1], "signed", "rv1_sign"]]
```

spec/src/cpu32.toml
```diff
@@ -0,0 +1,402 @@
...
+tag = "BYTE_ALU"
+input = [["opsel", "AND"], 32, "alu_flags"]
+output = ["*", 32, "signed"]
+multiplicity = "μ"
+
+[[constraints.ext]]
+kind = "template"
+tag = "SIGN"
+input = [["idx", "rv1", 1], 1]
+output = "rv1_sign"
```
erik-3milabs, line 374: Why not make the `SIGN` conditional on `signed` instead of `1`, and then remove `signed` from the `arg1[1]` definition? That would be more in line with how we typically use the `SIGN`-template.

Same applies to `rv2_sign` and `arg2[1]`.

spec/src/cpu32.toml
```diff
@@ -0,0 +1,402 @@
...
+    ["*", ["^", 2, 26], "rd"],
+    ["*", ["^", 2, 34], "instruction_length"],
+    ["*", ["^", 2, 42], "alu_flags"],
+]
+
+[[variables.multiplicity]]
+name = "μ"
+type = "Bit"
+desc = ""
+pad = 0
```
erik-3milabs, line 171: This multiplicity is not range checked. I wouldn't be surprised if this is _technically_ not needed, but I do think that adding it would be a good thing to do.

spec/dvrm.typ
```diff
@@ -27,10 +27,14 @@ The #dvrm chip provides division and remainder functionality, both signed and un
 The #dvrm chip is comprised of #nr_variables variables that are expressed using #nr_columns columns and leverages #nr_interactions interaction(s):
 #render_chip_variable_table(chip, config)
 
-= Assumptions
-#render_chip_assumptions(chip, config)
+// = Assumptions
+// #render_chip_assumptions(chip, config)
```
erik-3milabs:
```suggestion
```

spec/dvrm.typ
```diff
@@ -106,7 +110,8 @@ Rewriting R1, we find the constraint $not#`overflow` => #`n` - #`r` = #`qd`$.
 #footnote([Recall that @dvrm:c:sign_q allows to assert this equality even when `overflow`.])
 Since `n`, `d`, `q` and `r` are all 64-bit integers, we must assert this equality $mod 2^128$, rather than $mod 2^64$.
 To this end, we introduce `extended_n_sub_r` and leverage the `MUL` chip to verify that it is equal to $#`qd` mod 2^128$ using constraints @dvrm:c:mul_lower and @dvrm:c:mul_upper;
-@dvrm:c:q_range is included to uphold assumption @mul:a:rhs.
+@dvrm:c:q_range is included to uphold assumption @mul:c:rhs.
+// TODO: Update because this is now checked by MUL
```
erik-3milabs: ping

spec/decode.typ
```diff
@@ -34,33 +34,50 @@ Empty rows with the following content can be added to achieve this:
...
 For the purposes of explaining decoding, we decompress #decode's `packed_decode` variable into its constituent variables.
 Note that the below table is _not_ used in practice: it is solely used for the purposes of this explanation.
+The construction of the `alu_flags` and `mem_flags` columns is given here through virtual columns.
 
 #let config = load_config()
 #let uncompressed_chip = load_chip("src/decode_uncompressed.toml", config)
 
 #render_chip_variable_table(uncompressed_chip, config)
 
+First, we provide a mapping from an an ALU operation "descriptor" to the numerical value as used for the `alu_op` column:
```
erik-3milabs: We should clarify that this table also specifies the `opsel`-mapping used throughout (haven't found such a clarification yet).

spec/src/eq.toml
erik-3milabs, line 1: missing assumptions on `a` and `b` being range checked.

spec/src/eq.toml
```diff
@@ -0,0 +1,93 @@
...
+
+[[constraints.all]]
+kind = "arith"
+constraint = "$#`res` = #`eq` xor #`invert`$"
+poly = ["-", ["+", "res", ["*", 2, "eq", "invert"]], "eq", "invert"]
+
+[[constraints.all]]
+kind = "interaction"
+tag = "ALU"
+input = [["cast", "a", "DWordWL"], ["cast", "b", "DWordWL"], ["+", ["opsel", "EQ"], ["*", 64, "invert"]]]
```
erik-3milabs: unnecessary casting; `a` and `b` are already `WL`s by definition

spec/src/eq.toml
erik-3milabs, line 1: you could go for an all-zero padding here, since `EQ-C4` has multiplicity = mu.

spec/expr.typ
```diff
@@ -99,6 +99,10 @@
 // Typeset an expression as code
 #let expr_to_code = make_expr_formatter(
   (
+    "opsel": (pp, rec, e) => {
+      assert(type(e.at(1)) == type(""), message: "opsel expects a string")
+      repr(e.at(1))
```
erik-3milabs: can't say that I'm a fan of the "..." notation this introduces. 

What do you think of
```suggestion
      $chevron.l #raw(e.at(1)) chevron.r$
```
?

If you agree, please apply here and for `expr_to_math`

spec/src/lt.toml
```diff
@@ -85,15 +97,37 @@ ref = "lt:a:range_lhs"
...
+kind = "template"
+tag = "IS_BIT"
+input = ["signed"]
+ref = "lt:c:range_signed"
+
+[[constraints.defs]]
+kind = "template"
+tag = "IS_BIT"
+input = ["invert"]
+ref = "lt:c:range_invert"
```
erik-3milabs: I don't think these should necessarily be placed in the constraint group "defs". Either move them to a different group, or update the group name to more accurately reflect the content.

spec/mul.typ
```diff
@@ -32,9 +32,9 @@ The #mul chip is comprised of #nr_variables variables that are expressed using #
  $mat(delim: #none, top; bottom)$
 }
 
-= Assumptions
-The following range checks are assumed to be performed/enforced outside of this chip:
-#render_chip_assumptions(chip, config)
+// = Assumptions
+// The following range checks are assumed to be performed/enforced outside of this chip:
+// #render_chip_assumptions(chip, config)
```
erik-3milabs:
```suggestion
```

spec/src/shift.toml
```diff
@@ -138,31 +138,54 @@ desc = ""
 pad = 0
 
 
-
 # Assumptions
 
 [[assumptions]]
-desc = "`IS_HALF[in[i]]`"
+desc = "`IS_WORD[shift[3]]`"
```
erik-3milabs: do we even need this assumption? we never use shift[3]

spec/src/shift.toml
```diff
@@ -138,31 +138,54 @@ desc = ""
...
 
-[[assumptions]]
-desc = "`IS_BIT<direction>`"
-ref = "shift:a:direction"
+[[constraints.input]]
+kind = "template"
+tag = "IS_BYTE"
+input = [["idx", "shift", "i"]]
+iter = ["i", 0, 1]
+ref = "shift:c:range_shift"
```
erik-3milabs, line 164: do we even need the checks for shift[0] and shift[1]? we never use them.

spec/src/signatures.toml
```diff
@@ -49,11 +49,24 @@ tag = "DECODE"
...
 input = ["DWordWL", "DWordWL", "BaseField"]
 
-# SHIFT[out; in, shift, direction, signed, word_instr]
+# CPU32[timestamp, pc, instruction_length]
 [[signatures]]
-tag = "SHIFT"
+tag = "CPU32"
 kind = "interaction"
-input = ["DWordHL", "Byte", "Bit", "Bit", "Bit"]
+input = ["DWordWL", "DWordWL", "Byte"]
```
erik-3milabs: Since the CPU needs the `instruction_length` as a "return value" from CPU32, it might be nicer to turn it into the `output` for this interaction.

spec/src/store.toml
```diff
@@ -0,0 +1,116 @@
...
+desc = "Whether to write exactly 1 byte"
+def = ["-", "μ", "write2", "write4", "write8"]
+
+# Multiplicity
+
+[[variables.multiplicity]]
+name = "μ"
+type = "Bit"
+desc = ""
+pad = 0
```
erik-3milabs, line 55: not range checked; technically not required, but might be nice to add.

spec/src/store.toml
```diff
@@ -0,0 +1,116 @@
...
+
+[[constraints.all]]
+kind = "template"
+tag = "IS_BIT"
+input = [["+", "write2", "write4", "write8"]]
+
+[[constraints.all]]
+kind = "arith"
+constraint = "$#`write2` + #`write4` + #`write8` => #`μ`$"
+poly = ["*", ["+", "write2", "write4", "write8"], ["not", "μ"]]
```
erik-3milabs, line 98:
```suggestion
[[constraints.all]]
kind = "arith"
constraint = "$#`write2` + #`write4` + #`write8` => #`μ` = 1$"
poly = ["*", ["+", "write2", "write4", "write8"], ["not", "μ"]]
```

spec/src/store.toml
```diff
@@ -0,0 +1,116 @@
...
+
+[[constraints.all]]
+kind = "template"
+tag = "IS_BYTE"
+input = [["idx", "value", "i"]]
+iter = ["i", 0, 7]
```
erik-3milabs, line 105: missing multiplicity

spec/src/bytewise.toml
erik-3milabs, line 1: missing assumption that `op` is range checked elsewhere?

spec/src/load.toml
```diff
@@ -110,6 +120,7 @@ input = [0, "base_address", ["cast", "res", ["BaseField", 8]], "timestamp", "rea
 output = "res"
 multiplicity = "μ"
 
+# TODO? MSB8 batching?
```
erik-3milabs: ping

spec/src/store.toml
```diff
@@ -0,0 +1,116 @@
...
+
+[[constraint_groups]]
+name = "output"
+
+[[constraints.output]]
+kind = "interaction"
+tag = "MEMORY"
+input = ["timestamp", "base_address", ["cast", "value", "DWordWL"], ["+", ["*", 4, "write2"], ["*", 8, "write4"], ["*", 16, "write8"]]]
+output = ["cast", 0, "DWordWL"]
+multiplicity = ["-", "μ"]
```
erik-3milabs, line 122: Looks like the STORE bit is erroneously omitted here:

```suggestion
[[constraints.output]]
kind = "interaction"
tag = "MEMORY"
input = ["timestamp", "base_address", ["cast", "value", "DWordWL"], ["+", 1, ["*", 4, "write2"], ["*", 8, "write4"], ["*", 16, "write8"]]]
output = ["cast", 0, "DWordWL"]
multiplicity = ["-", "μ"]
```

spec/src/signatures.toml
```diff
@@ -49,11 +49,24 @@ tag = "DECODE"
...
+kind = "interaction"
+input = ["DWordWL", "DWordWL", "Byte"]
+output = "DWordWL"
+
+# MEMORY[out; timestamp, address, value, flags]
+[[signatures]]
+tag = "MEMORY"
+kind = "interaction"
+input = ["DWordWL", "DWordWL", "DWordWL", "Byte"]
 output = "DWordWL"
```
erik-3milabs, line 71: I propose we merge `out` and `value` on the bus:
1. when MEMORY=1, these values seem to be mutually exclusive in the CPU
2. LOAD and STORE have the `mem_op` flag inside of `mem_flags` to figure out which of the two picks it up.

spec/src/cpu.toml
erik-3milabs: For the same reason, it might be nice to add `word_instr => ALU = 0` here.

spec/src/cpu.toml
```diff
@@ -712,124 +451,50 @@ input = ["prev_pc_timestamp_borrow"]
...
-constraint = "$#`!LOAD` => #`rvd[0]` = #`res[:4]`$"
-poly = ["*", ["not", "LOAD"], ["-", ["idx", "rvd", 0], ["idx", ["cast", "res", "DWordWL"], 0]]]
-
-[[constraints.ext]]
-kind = "arith"
-constraint = "$#`!LOAD` => #`rvd[1]` = (1 - #`word_instr`) dot #`res[4:]` + #`res_ext_bit` dot (2^(32) - 1)$"
-desc = "_Sign_ extend the output if it wasn't a `LOAD`. Only `LOAD` has both `write_register = 1` and `rvd ≠ res`. `LOAD` and `word_instr` are disjoint"
-poly = ["*", ["not", "LOAD"], ["-", ["idx", "rvd", 1], ["*", ["not", "word_instr"], ["idx", ["cast", "res", "DWordWL"], 1]], ["*", "res_ext_bit", ["-", ["^", 2, 32], 1]]]]
-
+multiplicity = 1
```
erik-3milabs: I don't see how this multiplicity=1 plays nicely with padding rows.

spec/src/signatures.toml
erik-3milabs, line 85: is there a reason the MEMW interaction has not merged the three width-selector bits on the bus?

spec/src/cpu.toml
erik-3milabs: Also, these are both +1, where I'd expect them to be +1/-1

spec/src/mul.toml
erik-3milabs, line 1: missing range checks on `lhs_signed` and `rhs_signed`. 
Right now, one could fill `lhs_signed = 3` and `rhs_signed=0` and prove the wrong thing.

erik-3milabs (changes requested):
Note to self for follow-up review: have a closer look at the interplay between CPU and DECODE padding.

spec/src/cpu.toml
RobinJadoul: Maybe better to make it an explicit assumption: `MEMORY = 0 => IS_BIT<mem_flags>`

spec/src/cpu.toml
RobinJadoul: I think we've always had that, for exactly purposes like the choice between `rv2` and `imm` when one is guaranteed to be 0

spec/src/cpu.toml
RobinJadoul: We always read `PC`, decode it as a NOP row and write back the next PC as `PC + instruction_length = PC + 0`
Since there's no backing chip for this, at most it's a bit extra time to compute the logup sum here, _assuming_ there's an optimization to make zeros faster; we're not increasing the area with this.
Similar to the IS_BYTE checks that still happen for the decode consistency, they need to be summed here, and otherwise only increase a counter on the BITWISE table

RobinJadoul: Not sure I understand what the second comment is referring to

spec/src/decode_uncompressed.toml
RobinJadoul: Hmm, true, it should probably be B5 then, how the alu_flags is laid out
That'll cost two byte checks, I suppose, but we can't really not do them

spec/src/load.toml
RobinJadoul: Possible future optimization that's too uncertain to mention in the spec itself atm
Likely not really worth it, since afaik this is the only place we're actually still using MSB8

spec/src/lt.toml
RobinJadoul: It felt like the input range could count as a definition, but I can move them, sure

spec/src/shift.toml
RobinJadoul: We definitely use them, if only for reconstruction to a DWordWL on the bus
The shift[0] check could probably disappear because of the BYTE_ALU interaction, but it feels safer to keep them

spec/src/signatures.toml
RobinJadoul: That makes for the CPU to need to do a multiplex on the argument, which would complicate it, and potentially grow it, I think

spec/src/signatures.toml
RobinJadoul: It could optimize STORE a bit, but only a little, because you're trading off 3 cols + 3 bit checks for 1 col + 1 byte check
LOAD needs the individual bits anyway, and it makes the signature for other calls to MEMW uglier, so the gain didn't entirely seem worth it

spec/dvrm.typ
RobinJadoul: Leaving this for an upcoming check on all chips what optimizations/changes have become possible due to the changes the CPU update induces

spec/expr.typ
RobinJadoul: I don't hate it

spec/cpu.typ
RobinJadoul: I'll try moving the CPU32 interaction here instead, making it hopefully more coherent

spec/src/cpu.toml
RobinJadoul: Trying `MEMOP` instead, open to naming suggestions

spec/src/cpu.toml
RobinJadoul: `write_register` may actually not be safe to leave unconstrained, the rest doesn't matter as much, since those shouldn't have side effects

spec/src/cpu32.toml
RobinJadoul: Making `output` instead, in combination with the later remark

spec/src/eq.toml
RobinJadoul: Ah, I forgot to render those :)

spec/src/bytewise.toml
RobinJadoul: Not really, it's implicitly validated by `BYTE_ALU`

RobinJadoul:
Maybe 9b30b9e doesn't belong in this PR exactly, but it seemed simple enough for me to not want to deal with having a stacked PR for it :)

spec/src/decode_uncompressed.toml
erik-3milabs: > That'll cost two byte checks, I suppose, [...]

1. Where are you proposing to add two checks?
2. How do you arrive at _two_?

spec/src/cpu.toml
erik-3milabs: This essentially means that `mem_flags` is becoming a "multi-purpose column":  the interpretation of the value (`Bit` or `Byte`) is context dependent. 

Another place where we have this, is inside of `MEMW`: there `value` is interpreted differently based on `is_register` (`Byte` or `Word`). Back there, we settled on having `value: BaseField[8]` to avoid ambiguity/misinterpretations. 

What do you think about setting `mem_flags: BaseField` here as well?

spec/src/cpu.toml
```diff
@@ -210,9 +210,33 @@ desc = "`MEMORY` and `BRANCH` are mutually exclusive"
...
+[[constraints.assumptions]]
+kind = "arith"
+constraint = "$not (#`MEMORY` and #`BRANCH`)$"
+poly = ["*", "MEMORY", "BRANCH"]
+
+[[constraints.assumptions]]
+kind = "arith"
+constraint = "$(1 - #`MEMORY` - #`BRANCH`) => (#`read_register2` = 0 or #`imm[i]` = 0)$"
+poly = ["*", ["-", 1, "MEMORY", "BRANCH"], "read_register2", ["idx", "imm", "i"]]
+iter = ["i", 0, 1]
```
erik-3milabs: Could this be compressed as

```suggestion
[[constraints.assumptions]]
kind = "arith"
constraint = "$(1 - #`MEMORY` - #`BRANCH`) => (#`read_register2` = 0 or #`imm` = 0)$"
poly = ["*", ["-", 1, "MEMORY", "BRANCH"], "read_register2", ["+", ["idx", "imm", "0"], ["idx", "imm", "1"]]
```

spec/src/decode_uncompressed.toml
RobinJadoul: That was me being confused about where the change would be at first. You'd need two bute checks to check for a B4, if used e.g. to decompose the field

spec/src/cpu.toml
RobinJadoul: Bit is a subtype of Byte, so I'd still go with the more restrictive version here. We couldn't pack the mem_flags field if it was bigger than a byte.

spec/src/cpu.toml
```diff
@@ -210,9 +210,33 @@ desc = "`MEMORY` and `BRANCH` are mutually exclusive"
...
-desc = "When `MEMORY + BRANCH = 0`, either `rs2 = 0` or `imm = 0` should be enforced by the decoding. This is needed for `arg2`."
+desc = "When `MEMORY + BRANCH = 0`, either `read_register2 = 0` or `imm = 0` should be enforced by the decoding. This is needed for `arg2`."
 ref = "cpu:a:arg2-multiplex"
 
+[[assumptions]]
+desc = "$#`!MEMORY` => #`IS_BIT<mem_flags>`$"
+
+[[constraint_groups]]
+name = "assumptions"
+
```
erik-3milabs, line 221: This is an interesting evolution: we're now writing assumptions as constraints.

How about we instead let assumptions share the same "backbone" as constraints? In addition to improving readability/conciseness, it'd also be easier to machine-read the assumptions (which might be interesting for checking things later on)

spec/src/cpu.toml
erik-3milabs: The only side effect would be that the timestamp gets updated. Under the current design, I don't see that turning into a problem. 

However, we've been talking about inlining the CPU's memory accesses. This would also require inlining the `old_timestamp` variables. I can imagine these old timestamps enabling some funky time-travel shenanigans if not constrained properly on padding rows. Adding these constraints here now might help with a safer translation alter on.

spec/src/cpu32.toml
```diff
@@ -181,6 +181,15 @@ iter = ["i", 0, 1]
...
 desc = "`read_register2 = 0` or `imm = 0`, enforced by decoding."
 
+[[constraint_groups]]
+name = "assumptions"
+
+[[constraints.assumptions]]
+kind = "arith"
+constraint = "$#`read_register2` = 0 or #`imm[i] = 0`$"
+poly = ["*", "read_register2", ["idx", "imm", "i"]]
+iter = ["i", 0, 1]
```
erik-3milabs: This could be merged as
```suggestion
kind = "arith"
constraint = "$#`read_register2` = 0 or #`imm = 0`$"
poly = ["*", "read_register2", ["+", ["idx", "imm", 0],  ["idx", "imm", 1]]
```

spec/src/cpu.toml
erik-3milabs: > Not sure I understand what the second comment is referring to

The thing you fixed in 88fc8cf24deb9976c41a27e11a69f91e3e988e09 ;-)

spec/src/decode_uncompressed.toml
erik-3milabs: > You'd need two bute checks to check for a B4, if used e.g. to decompose the field

You mean you'd have to `IS_BYTE<X>` and `IS_BYTE<16X>` ?

> Hmm, true, it should probably be B5 then, how the alu_flags is laid out

You did change it into a B4 here. I kinda agreed that it should be B5, because there would otherwise be an "undeclared" bit inside of `alu_flags` which felt... tricky. What was the reason for switching to B4 after all?

spec/src/load.toml
erik-3milabs: Then I'd propose to spin it off as a separate feature and remove here; keeps things nice and clean.

spec/src/shift.toml
erik-3milabs: Wow, my remark did not make sense. Apologies.

spec/src/signatures.toml
erik-3milabs: I'm don't know whether that is actually the case, but I'm not going to investigate it; it is too much effort for too little reward.

spec/dvrm.typ
erik-3milabs: 1. The point I was trying to make was that the text still reads "assumption", while referring to a constraint.
2. Consider attaching this TODO to an issue; that way you don't need to hold it in mental memory & others can more easily find it if they were to pick that issue up.

spec/cpu.typ
```diff
@@ -30,36 +30,55 @@ The #cpu chip is comprised of #nr_variables variables that are expressed using #
...
+For the pure ALU path, `arg2` is computed as `rv2 + imm`, which relies on @cpu:a:arg2-multiplex to
+be either `rv2` or `imm`, depending on the instruction.
+The other contributions for `arg2` are specific to the (mutually exclusive, @cpu:a:mem-branch-mutex)
+`MEMORY` and `BRANCH` flags:
+- For the `MEMORY` path, we want the output of the ALU to be $#`rv1` + #`imm`$, as that is the
+  address at which the memory access occurs.
+- For the `BRANCH` path, when we do not have a JALR instruction, we want the ALU output to reflect the branch condition.
+- For the `BRANCH` path, when it is a JALR instruction, we want the ALU output to contain the next instructions PC value, to store into `rd`.
+
+Instructions having the `word_instr` flag set are delegated to the `CPU32` chip, which will do its own decoding and execution of the instruction.
```
erik-3milabs: I believe this sentence is out-of-date now.
```suggestion
```

spec/src/decode_uncompressed.toml
RobinJadoul: Currently, and presumably for the foreseeable future, it fits in 4 bits anyway; this avoids having to add B5 to the config
The constraints I was thinking of were `IS_BYTE<x>` and `IS_BYTE<x + (256 - 16)>`

spec/src/cpu.toml
RobinJadoul: Presumably, the inlining would follow the IS_HALF[delta] approach, which should be fine, being enforced with multiplicity read_registerX; but I don't disagree that having a few extra arith constraints is probably not worth leaving it out

spec/dvrm.typ
RobinJadoul: Filed under #638

RobinJadoul: (I do semi-regularly do a grep for TODO, so it's not purely in mental memory ;) )

spec/src/cpu.toml
RobinJadoul: Given that we have range-checks, I presume yes

spec/src/cpu.toml
RobinJadoul: Under previous assumptions, this has been tough. Taking inventory at the moment, we have:
- arithmetic constraints, as added in this PR/9b30b9ed
- IS_WORD assumptions
So it may be achievable
The commit itself came from a similar kind of inventarisation, and taking a bit of a safety-first approach to assumptions on the decoding specifically.
Logging #639 for this

spec/cpu.typ
RobinJadoul: How so? We still have the CPU32 interaction for this

spec/src/load.toml
RobinJadoul: Included this in #638

spec/src/cpu.toml
```diff
@@ -59,10 +59,10 @@ pad = 0
 
 # We encode C-type instructions with a length of 2, as this generality allows fusing common instruction combos
 [[variables.input]]
-name = "instruction_length"
+name = "half_instruction_length"
 type = "Byte"
 desc = "The number of bytes consumed by this instruction, commonly used to indicate whether the instruction is of C type, i.e., whether it is 2 bytes long instead of 4"
```
erik-3milabs: this description should be updated.

spec/src/cpu.toml
```diff
@@ -463,8 +471,8 @@ multiplicity = "MEMORY"
 
 [[constraints.mem]]
 kind = "arith"
-constraint = "$#`!MEMORY` => #`rvd` = #`res`$"
-poly = ["*", ["not", "MEMORY"], ["-", ["idx", "rvd", "i"], ["idx", ["cast", "res", "DWordWL"], "i"]]]
+constraint = "$#`!MEMORY` and #`!BRANCH` => #`rvd` = #`res`$"
+poly = ["*", ["-", 1, ["+", "MEMORY", "BRANCH"]], ["-", ["idx", "rvd", "i"], ["idx", ["cast", "res", "DWordWL"], "i"]]]
```
erik-3milabs: This might be easier to read
```suggestion
poly = ["*", ["-", 1, "MEMORY", "BRANCH"], ["-", ["idx", "rvd", "i"], ["idx", ["cast", "res", "DWordWL"], "i"]]]
```

spec/src/cpu.toml
```diff
@@ -385,11 +394,10 @@ prefix = "A"
...
 [[constraints.alu]]
 kind = "arith"
-constraint = "$#`arg2` = #`MEMORY` dot #`imm` + #`BRANCH` dot (1 - #`JALR`) dot #`rv2` + #`BRANCH` dot #`JALR` dot #`instruction_length` + (1 - #`MEMORY`) dot (1 - #`BRANCH`) dot (#`rv2` + #`imm`)$"
+constraint = "$#`arg2` = #`MEMORY` dot #`imm` + #`BRANCH` dot #`rv2` + (1 - #`MEMORY`) dot (1 - #`BRANCH`) dot (#`rv2` + #`imm`)$"
 poly = ["-", ["idx", "arg2", "i"],
     ["*", "MEMORY", ["idx", "imm", "i"]],
-    ["*", "BRANCH", ["not", "JALR"], ["idx", "rv2", "i"]],
-    ["*", "BRANCH", "JALR", ["idx", ["arr", "instruction_length", 0], "i"]],
+    ["*", "BRANCH", ["idx", "rv2", "i"]],
     ["*", ["not", "MEMORY"], ["not", "BRANCH"], ["idx", ["+", "rv2", "imm"], "i"]]
```
erik-3milabs: with MEMORY and BRANCH mutually exclusive, this could be changed to `(1-MEMORY-BRANCH)(rv2+imm)`, lowering the degree in the process.

spec/cpu.typ
erik-3milabs: That is correct. However, this sentence is in the CPU description twice: the new one you provided when you moved the CPU32 interaction up, and this (old) one that does not add any value here.

erik-3milabs (approved):
Green light!

PR #627, opened 2026-05-26
feat(crypto/mmcs): standalone Multi-Matrix Commitment Scheme [PR1 of streaming-MMCS plan]

diegokingston:
**Draft.** PR1 of a 3-PR plan to replace lambda-vm's N-Merkle-trees-per-round commit pipeline with a single Plonky3-style MMCS. This PR is the standalone module + threat model; no prover code touched, no proof format change.

## Why this PR is standalone

Past MMCS attempts in this repo regressed perf because the design pattern was wired into the hot path before the soundness model + memory cost were locked down. This PR isolates the API and the threat model so they can be reviewed in code form, with all 8 injection vectors as passing/failing tests, before any production change.

## Plan summary

| PR | Scope | LOC |
|---|---|--:|
| **PR1 (this)** | Standalone `MmcsBuilder` / `Mmcs` / `MmcsOpening` + tests | ~440 |
| PR2 | Wire into Round 1 main-trace commit, replace per-chip loop | ~500 |
| PR3 | Extend to aux trace (Round 2) + composition (Round 3) | ~300 |

Each PR is independently shippable with its own bench gate. The big proof-size win (~20× fewer Merkle path hashes per query for lambda-vm's chip topology) materializes after PR3 + a follow-up FRI batched-opening change.

## What's in this PR

- Module `crypto/crypto/src/merkle_tree/mmcs.rs` (437 LOC including tests).
- Plonky3-style layer injection: matrices sorted by `padded_height` desc (ties by `tag` asc); layer 0 = largest matrix's leaves; smaller matrices inject via `compress(node, leaf)` at the layer whose length matches their `padded_height`.
- `Open(global_index) → (per-matrix leaves at shifted indices, sibling chain)`.
- `Verify(opening, root, expected_specs)` reconstructs the root bottom-up.

## MVP restrictions (documented in module head)

- Distinct `padded_height` per matrix. Same-height is Phase 2.
- No SIMD / parallel hashing.
- Caller materializes full leaf digest arrays (no streaming chunks yet; that wraps on top later).
- Single root (no caps).

## Threat model coverage

Each of the 8 vectors from the design doc has a dedicated test:

| Vector | Test | Mitigation |
|---|---|---|
| 1. Cross-matrix row swap | `v1_cross_matrix_row_swap_is_rejected` | Matrix-leaf order check in `verify` |
| 2. Padding-byte freedom | `v2_unpadded_matrix_is_rejected_at_build` | `NotPowerOfTwo` at `add_matrix` |
| 3. Layer-injection order | `v3_same_height_matrices_rejected_in_mvp` | `DuplicateHeight` at `finalize` |
| 4. Leaf re-labelling | `v4_auth_path_forgery_via_relabeling_is_rejected` | Tag bound into opening |
| 5. Wrong leaf data | `v5_wrong_leaf_data_is_rejected` | Root mismatch |
| 6. Index tampering | `v6_index_tampering_rejected` | Path divergence |
| 7. Truncated path | `v7_truncated_path_is_rejected` | `siblings.len()` vs depth check |
| 8. Verifier-side spec mismatch | `v8_lying_about_spec_is_rejected` | `expected_specs` from config, never proof |

## Test plan

- [x] `cargo build -p crypto` — clean.
- [x] `cargo test -p crypto mmcs` — 13/13 passing.
- [x] `make lint` — clean across all three clippy configs.

### DIFF
--- a/crypto/crypto/src/merkle_tree/mmcs.rs
+++ b/crypto/crypto/src/merkle_tree/mmcs.rs
@@ -1,0 +1,436 @@
+//! Multi-Matrix Commitment Scheme (MMCS): a single Merkle root that
+//! commits to multiple matrices of (different) heights, with one
+//! authentication path per query covering all matrices.
+//!
+//! Plonky3-style layer injection: sort matrices by `padded_height` desc;
+//! layer 0 = largest matrix's leaves; compress pairs upward; at each
+//! layer whose length matches a smaller matrix's `padded_height`, inject
+//! that matrix's leaves via `compress(node_i, matrix.leaves[i])`.
+//!
+//! MVP scope:
+//! - All matrices have distinct `padded_height` (matches lambda-vm topology).
+//! - No SIMD, no streaming, no caps. Standalone module, not wired to prover.
+//!
+//! Security: see `docs/mmcs-streaming-design.md` for the 8-vector threat
+//! model; each vector is tested below.
+
+use alloc::vec::Vec;
+
+use super::traits::IsMerkleTreeBackend;
+
+/// Per-matrix domain separator. Caller-defined; verifier reconstructs
+/// from chip spec.
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct MatrixTag(pub [u8; 8]);
+
+impl MatrixTag {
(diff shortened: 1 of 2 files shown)


diegokingston:
Cerrado: scaffolding, no debía abrirse como PR. La rama queda con el commit para iterar localmente.

PR #641, opened 2026-06-01
refactor(math): remove dead/self-referential code from polynomial.rs

diegokingston:
Remove functions with no production callers (only exercised by their own tests) from crypto/math/src/polynomial.rs:

- compose_fft (was #[cfg(test)], tested only by composition_fft_works)
- pad_with_zero_coefficients + pad_with_zero_coefficients_to_length
- mul_with_ref (no Polynomial Mul operator exists)
- new_monomial

Adjust affected tests: drop self-tests, rewrite the evaluate-coverage tests to build polynomials via Polynomial::new, and inline the single production-test call site (X^8) in prover_tests.


MauroToscano (changes requested):
Why do we even have the Polynomial at this point ? You cannot do much with them, right ?

MauroToscano:
Closing for now — rather than a piecemeal dead-code removal here, we want to re-review the whole polynomial module holistically. We'll fold this cleanup into that pass.

PR #642, opened 2026-06-02
spec: Move to the epoch system

RobinJadoul:
After initial discussion, the choice was made for epoch-based proving, so this PR now formalizes this.
Additionally, it deals with the integration of COMMIT into this new paradigm by use of memory domain separation.
Since epochs are short, we can also reduce the surface area everywhere by only using `Word` variables for timestamps. This leaves open potential optimizations in MEMW where now only single-limb values need to be compared.

### DIFF
--- a/spec/commit.typ
+++ b/spec/commit.typ
@@ -65,6 +65,17 @@ However, since it is practically impossible to commit more than $2^64-2^32$ byte
 
 Next, we read the `value` located at buffer address `address` and commit to it under the given `index`.
 This is only performed when we have not yet reached the `end` of the commit sequence.
+Values are committed by letting the verifier initialize and finalize the global memory argument (see @memory and @streaming),
+with the claimed commitments in its own domain separated part of memory, with domain separator value 3.
+This chip then checks that the same value as the one being committed is then found at the corresponding address.
+In doing this, we enforce that all values being committed match the claimed commitment,
+and the verifier can additionally check that register 254 contains the correct value to ensure
+the correct amount of bytes have been committed.#footnote[
+  We additionally note here that for very large commitments (with index $>= 2^32$),
+  the address can potentially become denormalized, but since no other chips or systems interact with
+  this memory domain, there is no issue.
+  The usual consistency guarantee from the LogUp argument and correct initialization as for general addresses applies.
+]
 #render_constraint_table(chip, config, groups: "commit")
 
 In parallel, we compute $#`address_incr` = #`address` + 1$ (@commit:c:address_incr) as address of the next byte to commit, and $#`count_decr` = #`count` - 1$ (@commit:c:count_decr) as the number of bytes that still has to be committed after committing this byte.
--- a/spec/commit.typ
+++ b/spec/commit.typ
@@ -66,7 +66,7 @@ However, since it is practically impossible to commit more than $2^64-2^32$ byte
 Next, we read the `value` located at buffer address `address` and commit to it under the given `index`.
 This is only performed when we have not yet reached the `end` of the commit sequence.
 Values are committed by letting the verifier initialize and finalize the global memory argument (see @memory and @streaming),
-with the claimed commitments in its own domain separated part of memory, with domain separator value 3.#footnote[
+with the claimed commitments in its own domain separated part of memory, with domain separator value 2.#footnote[
   In order to make sure the verifier can properly finalize the committed values, the last epoch can "bring forward"
   all commitments from earlier epochs, similar to padded values, in the `L2G` table.
   Then the contribution of the commitments only consists of the tuples `(2, address, last_epoch_index, value)`, which is entirely known to the verifier.
--- a/spec/commit.typ
+++ b/spec/commit.typ
@@ -73,10 +73,10 @@ with the claimed commitments in its own domain separated part of memory, with do
 ]
 This chip then checks that the same value as the one being committed is then found at the corresponding address.
 In doing this, we enforce that all values being committed match the claimed commitment,
-and the verifier can additionally check that register 254 contains the correct value to ensure
+and the verifier should additionally check that register 254 contains the correct value to ensure
 the correct amount of bytes have been committed.#footnote[
   We additionally note here that for very large commitments (with index $>= 2^32$),
-  the address can potentially become denormalized, but since no other chips or systems interact with
+  the (commit space) address can potentially become denormalized, but since no other chips or systems interact with
   this memory domain, there is no issue.
   The usual consistency guarantee from the LogUp argument and correct initialization as for general addresses applies.
 ]
--- a/spec/commit.typ
+++ b/spec/commit.typ
@@ -66,7 +66,11 @@ However, since it is practically impossible to commit more than $2^64-2^32$ byte
 Next, we read the `value` located at buffer address `address` and commit to it under the given `index`.
 This is only performed when we have not yet reached the `end` of the commit sequence.
 Values are committed by letting the verifier initialize and finalize the global memory argument (see @memory and @streaming),
-with the claimed commitments in its own domain separated part of memory, with domain separator value 3.
+with the claimed commitments in its own domain separated part of memory, with domain separator value 3.#footnote[
+  In order to make sure the verifier can properly finalize the committed values, the last epoch can "bring forward"
+  all commitments from earlier epochs, similar to padded values, in the `L2G` table.
+  Then the contribution of the commitments only consists of the tuples `(2, address, last_epoch_index, value)`, which is entirely known to the verifier.
+]
 This chip then checks that the same value as the one being committed is then found at the corresponding address.
 In doing this, we enforce that all values being committed match the claimed commitment,
 and the verifier can additionally check that register 254 contains the correct value to ensure
--- a/spec/src/commit.toml
+++ b/spec/src/commit.toml
@@ -160,10 +160,17 @@ ref = "commit:c:read_value"
 
 [[constraints.commit]]
 kind = "interaction"
-tag = "COMMIT"
-input = ["index", "value"]
+tag = "memory"
+input = [3, ["arr", "index", 0], ["cast", 0, "Word"], "value"]
 multiplicity = ["-", "μ", "end"]
-ref = "commit:c:commit_value"
+ref = "commit:c:commit_value_out"
+
+[[constraints.commit]]
+kind = "interaction"
+tag = "memory"
+input = [3, ["arr", "index", 0], ["cast", 0, "Word"], "value"]
+multiplicity = ["-", ["-", "μ", "end"]]
+ref = "commit:c:commit_value_in"
 
 [[constraint_groups]]
 name = "end"
--- a/spec/src/commit.toml
+++ b/spec/src/commit.toml
@@ -161,14 +161,14 @@ ref = "commit:c:read_value"
 [[constraints.commit]]
 kind = "interaction"
 tag = "memory"
-input = [3, ["arr", "index", 0], ["cast", 0, "Word"], "value"]
+input = [2, ["arr", "index", 0], 0, "value"]
 multiplicity = ["-", "μ", "end"]
 ref = "commit:c:commit_value_out"
 
 [[constraints.commit]]
 kind = "interaction"
 tag = "memory"
-input = [3, ["arr", "index", 0], ["cast", 0, "Word"], "value"]
+input = [2, ["arr", "index", 0], 1, "value"]
 multiplicity = ["-", ["-", "μ", "end"]]
 ref = "commit:c:commit_value_in"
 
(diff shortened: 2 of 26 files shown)


spec/streaming.typ
```diff
@@ -0,0 +1,87 @@
...
+
+= Approach 2: Prove-epoch
+
+In the second potential approach, we observe that the majority of constraints on a contiguous section of cycles
+is independent of the other cycles: only the consistency of the memory accesses has to cross all boundaries.
+As such, we proceed in "epochs" ---the size of which is again informed by the memory pressure--- where
+all table-to-table interactions can be proven within a single epoch.
+
+To deal with cross-epoch memory, we introduce a "local-to-global" table per epoch that, in essence,
+is an epoch-local memory initialiation and finalization mechanism.
```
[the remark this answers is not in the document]
RobinJadoul: 2 is clearly enforced by the global memory LogUp described by the approach

spec/streaming.typ
```diff
@@ -0,0 +1,87 @@
...
+
+This approach comes at a clear cost of extra interpolations and Merkle tree evaluations,
+but manages to keep a clean evaluation model that almost directly (apart from the early commitments and batching)
+corresponds to the naive, memory-intensive proving model.
+An optimization is possible to reduce the cost of the *Open* phase, by keeping the internal nodes of the merkle tree
+in memory, obviating the need to recompute it; while still dropping the biggest memory cost (the leaves).
+
+To distribute this workload across worker nodes or multiple GPUs, it is possible to "send off"
+each batch of tables being retired to the next available worker.
+With some care, it is likely possible to decouple the Fiat-Shamir challenges of multiple retirement batches,
```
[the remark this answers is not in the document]
RobinJadoul: The underlying idea is that the randomizer for a batch depends on the batch itself, so any offset elsewhere will be randomized with a different challenge

spec/streaming.typ
```diff
@@ -0,0 +1,87 @@
...
+The overarching idea in both is to let the prover build up tables, and once memory-pressure grows
+too much, perform some prover work to allow evicting tables from memory.
+The difference lies in exactly what and how much proving happens, and how much extra computation is needed
+later on.
+
+= Approach 1: Prove-and-retire
+
+In the first potential approach, we proceed in multiple passes:
+- *Commit*: The prover proceeds through execution, witness generation and interpolation as usual
+  once the memory pressure becomes too large, the prover batch commits to all full (or large enough) tables
```
erik-3milabs:
```suggestion
- *Commit*: The prover proceeds through execution, witness generation and interpolation as usual.
  Once the memory pressure becomes too large, the prover batch commits to all full (or large enough) tables
```

erik-3milabs (commented):
solid outline.

spec/streaming.typ
```diff
@@ -0,0 +1,140 @@
...
+We currently assume that no more than $2^20$ epochs can occur in one execution.
+This can be extended at the cost of extra interactions and potentially extra columns.
+
+#render_constraint_table(l2gchip, config, groups: "consistency")
+
+Then we have local constraints that should be enforced inside of the epoch.
+
+#render_constraint_table(l2gchip, config, groups: "local")
+
+And finally, the interactions that are pare of the global memory consistency proof.
```
cdesaintguilhem:
```suggestion
And finally, the interactions that are part of the global memory consistency proof.
```

spec/commit.typ
```diff
@@ -65,6 +65,21 @@ However, since it is practically impossible to commit more than $2^64-2^32$ byte
 
 Next, we read the `value` located at buffer address `address` and commit to it under the given `index`.
 This is only performed when we have not yet reached the `end` of the commit sequence.
+Values are committed by letting the verifier initialize and finalize the global memory argument (see @memory and @streaming),
+with the claimed commitments in its own domain separated part of memory, with domain separator value 3.#footnote[
```
cdesaintguilhem: This still says the separator value is `3`, but the footnote seems to imply it
should be `2`, and `memory.typ` still says it should be `2`.

cdesaintguilhem (changes requested):
The big changes look good to me. I notice the text still mentions `3` for the domain separator value for commitment values whereas other parts use `2`. Also, I spotted a typo.

spec/commit.typ
RobinJadoul:
```suggestion
with the claimed commitments in its own domain separated part of memory, with domain separator value 2.#footnote[
```

spec/src/page.toml
```diff
@@ -44,19 +44,14 @@ kind = "template"
...
-kind = "template"
-tag = "IS_BYTE"
-input = ["fini"]
-
 [[constraints.all]]
 kind = "interaction"
-tag = "memory"
-input = [0, "address", ["cast", 0, "DWordWL"], "init"]
+tag = "GLOBAL_MEM"
+input = [0, "address", 0, "init"]
```
erik-3milabs, line 50: By setting the init_epoch=0, the global initialization "phase" effectively becomes epoch 0. This means that the first _actual_ epoch is 1 (since `fini_epoch - init_epoch - 1 >= 0` due to L2G-C2), and thus `fini_epoch` in `L2G` should start at 1.
1) I don't see this written down anywhere (and feels like a valuable addition)
2) it feels weird to start with 1...? Then again, having initialization start at -1 might be just as annoying.

spec/src/commit.toml
```diff
@@ -160,10 +160,17 @@ ref = "commit:c:read_value"
...
 multiplicity = ["-", "μ", "end"]
-ref = "commit:c:commit_value"
+ref = "commit:c:commit_value_out"
+
+[[constraints.commit]]
+kind = "interaction"
+tag = "memory"
+input = [2, ["arr", "index", 0], 1, "value"]
+multiplicity = ["-", ["-", "μ", "end"]]
+ref = "commit:c:commit_value_in"
```
erik-3milabs, line 173: Could you elaborate why there are two interactions here? I am unable to gather this from the text.

spec/src/cpu.toml
```diff
@@ -488,14 +488,14 @@ input = ["prev_pc_timestamp_borrow"]
 [[constraints.mem]]
 kind = "interaction"
 tag = "memory"
-input = [1, ["arr", ["+", ["*", 2, 255], "i"], 0], ["arr", ["+", ["-", ["idx", "timestamp", 0], ["*", 3, ["not", "pc_double_read"]]], ["*", ["^", 2, 32], "prev_pc_timestamp_borrow"]], ["-", ["idx", "timestamp", 1], "prev_pc_timestamp_borrow"]], ["idx", "pc", "i"]]
+input = [1, ["arr", ["+", ["*", 2, 255], "i"], 0], ["-", "timestamp", ["*", 3, ["not", "pc_double_read"]]], ["idx", "pc", "i"]]
```
erik-3milabs, line 480: IIUC, `prev_pc_timestamp_borrow` is now unused and can therefore be removed.

spec/src/halt.toml
```diff
@@ -21,39 +20,39 @@ name = "all"
...
+input = [1, ["arr", ["+", ["*", 2, 255], "i"], 0], ["+", "timestamp", 1], ["idx", "pc", "i"]]
 multiplicity = 1
 iter = ["i", 0, 1]
 ref = "halt:c:consume_pc"
 
 [[constraints.all]]
 kind = "interaction"
 tag = "memory"
-input = [1, ["arr", ["+", ["*", 2, 255], "i"], 0], ["arr", ["+", ["idx", "timestamp", 0], 1], ["idx", "timestamp", 1]], ["idx", ["arr", 1, 0], "i"]]
+input = [1, ["arr", ["+", ["*", 2, 255], "i"], 0], ["+", "timestamp", 1], ["idx", ["arr", 1, 0], "i"]]
```
erik-3milabs: This is easier to read, don't you think?
```suggestion
input = [1, ["arr", ["+", ["*", 2, 255], "i"], 0], ["+", "timestamp", 1], ["-", 1, "i"]]
```

spec/src/l2g.toml
```diff
@@ -0,0 +1,89 @@
...
+multiplicity = 1
+
+
+[[constraint_groups]]
+name = "local"
+
+[[constraints.local]]
+kind = "interaction"
+tag = "memory"
+input = ["domain", "address", 0, "init_value"]
```
erik-3milabs, line 68: Won't setting `timestamp=0` here given problems when this value is accessed is written on the first cycle of the epoch? On that write, the `old_timestamp < timestamp` wouldn't hold, right?

spec/commit.typ
```diff
@@ -65,6 +65,21 @@ However, since it is practically impossible to commit more than $2^64-2^32$ byte
 
 Next, we read the `value` located at buffer address `address` and commit to it under the given `index`.
 This is only performed when we have not yet reached the `end` of the commit sequence.
+Values are committed by letting the verifier initialize and finalize the global memory argument (see @memory and @streaming),
+with the claimed commitments in its own domain separated part of memory, with domain separator value 2.#footnote[
+  In order to make sure the verifier can properly finalize the committed values, the last epoch can "bring forward"
+  all commitments from earlier epochs, similar to padded values, in the `L2G` table.
```
erik-3milabs, line 71: what is to prevent the prover from not committing and just bringing the initialization forward?

spec/src/commit.toml
RobinJadoul: The commit memory space is essentially used as read-only memory, so we get a memory read here (using 1 instead of `timestamp` for no particular reason). We do this so that it still fits nicely within the existing L2G tables, so we can't just conjure up a new value or consume the value, without making the L2G unbalanced.

spec/src/l2g.toml
RobinJadoul: This is in line with how the pre-epoch memory system and `PAGE` worked. The CPU preprocessed timestamp column starts at `t = 4 * 1`

spec/commit.typ
RobinJadoul: The verifier checks the final value of the commit index register (x254), so if a wrong amount of bytes was committed, verification fails.

spec/src/page.toml
RobinJadoul: Initialization at -1 is probably weirder, and potentially conflicts with ordering checks (since init_epoch needs to be range-checked).
It's also in line with the timestamp initialization within an epoch.

spec/src/commit.toml
erik-3milabs: Do I understand correctly that whenever value `x` under index `i` is committed in an epoch `e`, that epoch's L2G table will include entry
```
domain = 2
address = i
init_epoch = 0
init_value = x
fini_timestamp = 1
fini_value = x
fini_epoch = e
```
?

spec/commit.typ
```diff
@@ -65,6 +65,21 @@ However, since it is practically impossible to commit more than $2^64-2^32$ byte
...
 This is only performed when we have not yet reached the `end` of the commit sequence.
+Values are committed by letting the verifier initialize and finalize the global memory argument (see @memory and @streaming),
+with the claimed commitments in its own domain separated part of memory, with domain separator value 2.#footnote[
+  In order to make sure the verifier can properly finalize the committed values, the last epoch can "bring forward"
+  all commitments from earlier epochs, similar to padded values, in the `L2G` table.
+  Then the contribution of the commitments only consists of the tuples `(2, address, last_epoch_index, value)`, which is entirely known to the verifier.
+]
+This chip then checks that the same value as the one being committed is then found at the corresponding address.
+In doing this, we enforce that all values being committed match the claimed commitment,
+and the verifier can additionally check that register 254 contains the correct value to ensure
```
erik-3milabs:
```suggestion
and the verifier should additionally check that register 254 contains the correct value to ensure
```

spec/commit.typ
```diff
@@ -65,6 +65,21 @@ However, since it is practically impossible to commit more than $2^64-2^32$ byte
...
+the correct amount of bytes have been committed.#footnote[
+  We additionally note here that for very large commitments (with index $>= 2^32$),
+  the address can potentially become denormalized, but since no other chips or systems interact with
```
erik-3milabs: how could this happen? C9 increments the address using the ADD chip, it should not be able to overflow, right?

spec/src/commit.toml
RobinJadoul: That seems accurate, yes

spec/commit.typ
RobinJadoul: The address from which the data to commit is read will be fine, but we're also using `[index, 0]` as "address" in the commit memory space, which could get denormalized (C7)

erik-3milabs: I see. In that case, please update the text to clarify this "address in commit memory space", because the use of "address" to refer to that was unclear to me.

erik-3milabs (approved):

PR #646, opened 2026-06-03
Add extra constraints to prevent register side effects in CPU32 padding rows

RobinJadoul:


spec/src/cpu32.toml
```diff
@@ -350,6 +355,21 @@ multiplicity = "write_register"
...
+
+[[constraints.logup]]
+kind = "arith"
+constraint = "$#`!μ` => #`read_register2 = 0`$"
+poly = ["*", ["not", "μ"], "read_register2"]
+
+[[constraints.logup]]
+kind = "arith"
+constraint = "$#`!μ` => #`write_register = 0`$"
+poly = ["*", ["not", "μ"], "write_register"]
```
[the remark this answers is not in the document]
RobinJadoul: ALU should be free of side-effects, and is similarly not gated on `word_instr` in the `CPU` chip

spec/src/cpu32.toml
```diff
@@ -284,6 +284,11 @@ input = [["idx", "res", "i"]]
 multiplicity = "μ"
 iter = ["i", 0, 3]
 
+[[constraints.range]]
+kind = "template"
+tag = "IS_BIT"
+input = ["μ"]
```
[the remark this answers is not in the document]
RobinJadoul: Whoops, my grep skills failed me

spec/src/cpu32.toml
erik-3milabs, line 190:
```suggestion
[[constraints.assumptions]]
kind = "arith"
constraint = "$#`read_register2` = 0 or #`imm = 0`$"
poly = ["*", "read_register2", ["+", ["idx", "imm", 0], ["idx", "imm", 1]]]
```

erik-3milabs, line 426: To save some lookup-effort on padding rows, this should be conditioned on `mu`.

```suggestion
[[constraints.ext]]
kind = "template"
tag = "SIGN"
input = [["idx", "res", 1], "μ"]
output = "res_sign"
```

erik-3milabs, line 394: For this to be safe, `signed` needs to be _always_ range checked (not just on active `mu=1` rows, as is happening now through the `AND`-lookup). Reason being that without this constraint, `signed` could be given arbitrary values in padding-rows (e.g., `-1`) and use this `SIGN` template (and the one for `rv2` as well) to cancel out invalid `MSB16` elsewhere in the program.

I checked the other places `SIGN` is used; all their conditions are properly range checked 100% of the time.

erik-3milabs (changes requested):
This patch looks fine. In reviewing, I found a different bug in CPU32, though.

spec/src/cpu32.toml
RobinJadoul: What exactly is the diff here?

erik-3milabs: `imm[i]` = 0 => `imm` = 0

Weird way to render it though, GitHub.

RobinJadoul: Doesn't make a massive difference with the lookup being precomputed, but it doesn't hurt either

RobinJadoul: Resolving through a `signed => μ` constraint

spec/src/cpu32.toml
```diff
@@ -375,6 +375,11 @@ multiplicity = ["-", "μ"]
 [[constraint_groups]]
 name = "ext"
 
+[[constraints.ext]]
+kind = "arith"
+constraint = "$#`signed` => #`μ`$"
```
erik-3milabs: I'd make it a touch more explicit.
```suggestion
constraint = "$#`signed` eq.not 0 => #`μ` = 1$"
```
Other than that, lgtm.

erik-3milabs (approved):

PR #647, opened 2026-06-03
Feat/streaming prover

diegokingston:


MauroToscano:
We are now using: https://github.com/yetanotherco/lambda_vm/pull/685

PR #651, opened 2026-06-08
continuations (first step): Add epoch split and proving

nicole-graus:
## Description 

First step toward continuations (proving arbitrarily long executions by splitting them into epochs. This lands the foundation; cross-epoch soundness is a follow-up.

- `Executor::run_epochs(n)` splits execution into epochs of N cycles, capturing each epoch's boundary state (pc, registers, memory).
- `Traces::from_image_and_logs` builds one epoch's traces from an explicit starting memory + register image (the first epoch from the ELF, later epochs from the previous epoch's snapshot).
- HALT is applied only to the final epoch (`is_final` / `VmAirs.include_halt`); intermediate epochs keep their boundary state.
- `VmAirs::new` takes an optional `register_init` so an epoch's REGISTER commitment can use its boundary register snapshot.
- Each epoch can be proven and verified in isolation (tests for the first and a non-first epoch).

### DIFF
--- a/executor/src/vm/execution.rs
+++ b/executor/src/vm/execution.rs
@@ -30,6 +30,20 @@ pub struct ExecutionResult {
 /// Size of each log chunk - balances memory usage vs callback overhead
 const CHUNK_SIZE: usize = 100_000;
 
+/// Default number of cycles (instructions) per continuation epoch.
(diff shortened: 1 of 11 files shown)


MauroToscano:
We will just use: We are now using: https://github.com/yetanotherco/lambda_vm/pull/685

PR #654, opened 2026-06-08
refactor: compute the zero-init page commitment once (step 2)

MauroToscano:
**Stacked on #653** (single global page size). Review/merge that first; this PR's diff is only the step-2 change.

## What
Compute the zero-init page preprocessed commitment **once per `VmAirs` construction** and share it across all zero-init pages, replacing the process-global `ZERO_PAGE_4K_COMMITMENT` `OnceLock`.

## Why this is safe (non-cryptographic)
Every zero-init page (stack/heap/BSS) has the *identical* commitment — OFFSET is page-relative and INIT is all-zero, so it depends only on the proof options (page size is the global `DEFAULT_PAGE_SIZE` after #653). The verifier still **derives the value itself from public data**, exactly as before — committed bytes are bit-identical, trust model unchanged. This is a pure optimization + cleanup, deliberately landed *before* the cryptographic step.

## Also fixes a latent bug
The removed `OnceLock` cached keyed only by page *content*, not by `(blowup, coset)` — so a process verifying at more than one blowup would reuse the first blowup's root and falsely reject later proofs. (Masked today because `verify()` pins blowup=2.) The per-construction compute uses the actual options.

## Explicitly NOT here (→ step 3, the cryptographic step)
- No compile-time static const (trust anchor).
- No verifier-provided `page_commitments` injection.
ELF data pages still recompute per page.

## Testing
`cargo test --release -p lambda-vm-prover` → **384 passed, 0 failed**. fmt + clippy clean.

### DIFF
--- a/prover/src/lib.rs
+++ b/prover/src/lib.rs
@@ -403,6 +403,14 @@ impl VmAirs {
             register::preprocessed_commitment(proof_options, elf.entry_point),
             register::NUM_PREPROCESSED_COLS,
         );
+        // Every zero-init page shares one preprocessed commitment: OFFSET is
+        // page-relative and INIT is all-zero, so it depends only on the proof
+        // options. Compute it once here rather than once per zero-init page.
+        // Every program has at least one zero-init page (the stack is
+        // zero-initialized), so this commitment is always used.
+        let zero_init_commitment =
+            page::compute_precomputed_commitment(&page::PageConfig::zero_init(0), proof_options);
+
         let pages: Vec<_> = page_configs
             .iter()
             .map(|config| {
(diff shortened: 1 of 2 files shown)


MauroToscano:
> ## Codex Code Review
> **Findings**
> 
> * **Low - unnecessary PAGE commitment build when no zero-init page uses it**: [prover/src/lib.rs](/home/runner/work/lambda_vm/lambda_vm/prover/src/lib.rs:409) now computes the zero-init PAGE preprocessed commitment unconditionally before iterating `page_configs`. That runs the full PAGE FFT/LDE/Merkle path even for calls with `page_configs` empty, only private-input pages, or only ELF data pages. This is a significant avoidable cost for verifier/prover setup. Compute it lazily on the first non-private `init_values.is_none()` page, or guard with `page_configs.iter().any(...)`.
> 
> No security vulnerabilities or correctness bugs found in the diff.

This is wrong. There will always be at least one zero initialized page, since the stack needs a zero page

nicole-graus (approved):

PR #655, opened 2026-06-09
spec: `ECSM`

erik-3milabs:
This PR introduces the `ECSM` (_Elliptic Curve Scalar Multiplication_) accelerator.


spec/src/ecsm.toml
```diff
@@ -0,0 +1,609 @@
...
+[[assumptions]]
+desc = "$(#`addr_xG` mod 2^32) + 24 < 2^32$"
+ref = "ec:a:addr_xG_alignment"
+
+[[assumptions]]
+desc = "$(#`addr_k` mod 2^32) + 31 < 2^32$"
+ref = "ec:a:addr_k_alignment"
+
+[[assumptions]]
+desc = "$(#`addr_xR` mod 2^32) + 24 < 2^32$"
```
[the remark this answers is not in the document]
erik-3milabs: This + 31 is on purpose, to make sure the `EC_SCALAR` chip doesn't trip up.
This comment can be disregarded.

spec/ecsm.typ
```diff
@@ -0,0 +1,442 @@
...
+  #`x2` &equiv x_G^2 mod p,\
+  y_G^2 &equiv x_G dot #`x2` + b  mod p\
+$
+which combine to $y_G^2 equiv x_G^3 + b mod p$.
+Rewriting the two relations, we get
+$
+  q_0 &= (x_G^2 - #`x2`) dot p^(-1),\
+  q_1 &= (y_G^2 - x_G dot #`x2`-b) dot p^(-1) + p.
+$
+Using the fact that $x_G, y_G, #`x2` in [0, p)$, we find that $q_0 in [0, p)$ and $q_0 in [0, 2p)$.
```
erik-3milabs:
```suggestion
Using the fact that $x_G, y_G, #`x2` in [0, p)$, we find that $q_0 in [0, p)$ and $q_1 in [0, 2p)$.
```

spec/ecsm.typ
```diff
@@ -0,0 +1,442 @@
...
+Here, we capture the index of the most significant 1-bit of `k` in `len_k`.
+Note: if the prover decides to capture a lesser significant bit here, the LogUp will not balance, as the skipped bits will never taken off the bus.
+Next, we interact with the `ECDAS` chip, providing `G` both as the accumulator, and increment (@ec:c:start_double_add); we specifically instruct the chip to start with a _double_-operation.
+After completing its double-and-add sequence, the result is captured in `R` (@ec:c:receive_double_add).
+#render_constraint_table(ecsm_chip, config, groups: "delegate")
+
+=== Range check `xR`
+Before storing $x_R$, it is verified that $x_R in [0, p)$.
+To this end, witness $#`xR_sub_p` := #`xR` - p mod 2^256$ is added to `p`; if the addition sums to `xR` and overflows $mod 2^256$, it must hold that $#`xR` < p$.
+The addition is constrained by requiring that `c3` are bits (@ec:c:range_c3); an overflow occurs if and only if $#`c3[7]` = 1$ (@ec:c:xR_addition_overflows).
```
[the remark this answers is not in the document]
erik-3milabs: good point. wrong line though

erik-3milabs:
> ## Codex Code Review
> 
> **Findings**
> 
> **High: inactive `EC_SCALAR` rows emit `SERVE_K` tokens.** In [spec/src/ec_scalar.toml](/home/runner/work/lambda_vm/lambda_vm/spec/src/ec_scalar.toml:107), the recursive `SERVE_K` interaction uses `multiplicity = ["not", "last_limb"]`, independent of `μ`. Since line 99 forces `last_limb = 0` when `μ = 0`, every inactive/padding row emits a free `SERVE_K` token. Those can be consumed by active scalar rows and produce `BIT` tokens not rooted in the ECSM request’s `addr_k`, breaking the binding between `k` and the double-and-add sequence. Gate this by activity, e.g. `μ * (1 - last_limb)` or `μ - last_limb` with the existing implication.

Indeed. Fixed it.

> **High: ECDAS lambda carry recurrence uses `c0[0]` instead of `c0[i]`.** In [spec/src/ecdas.toml](/home/runner/work/lambda_vm/lambda_vm/spec/src/ecdas.toml:276), the `i = 1..63` lambda constraint starts with `2^8 * c0[0]`. The rendered constraint and carry pattern require `2^8 * c0[i]`. As written, most lambda carry limbs are not tied into the recurrence correctly, so the modular relation for `λ` is underconstrained for higher limbs. This can let invalid EC add/double steps satisfy the spec.

Fixed it

erik-3milabs:
> ## Code Review — ECSM Accelerator Spec
>
> ### Minor observations
> 
> * The doubling formula is silently specialized to `a=0`; the general theory section (which still writes `E(a,b,p)`) could note this specialization explicitly.

Technically, this is already addressed: the introduction mentions that it accelerates points in `E(0, b, p)`

> * Several new files are missing a trailing newline (`about_ecalls.typ`, `ec_scalar.toml`, `ecdas.toml`).

Fixed

> * The aside explaining why `y_A != 0` for secp256k1 says "as previously established" but the establishment is not in this document — citing the odd-order argument explicitly would make it self-contained.

It is in the same document, but a section earlier. Disregarding the comment.

spec/about_ecalls.typ
```diff
@@ -31,4 +31,5 @@ Negative numbers (represented as 2s complement 64-bit numbers), are used for our
 / 64: `write` (@commit)
 / 93: `exit` (@halt)
 / -1: `SHA256` (@sha256)
-/ -2: `KECCAK` (@keccak)
\ No newline at end of file
+/ -2: `KECCAK` (@keccak)
+/ -3: `ECSM` (@ecsm)
```
RobinJadoul: Maybe interesting to make this `-11` instead, so that we essentially reserve some extra space for more hashes?

spec/src/ecsm.toml
```diff
@@ -0,0 +1,629 @@
...
+ref = "ec:a:addr_xR_alignment"
```
RobinJadoul: How safe is this? The fact that this comes from an ECALL means that it is not an assumption we can ensure at the spec level, and rather something that is pushed down into the guest program.

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+This chip is triggered by an `ECALL` with the opcode indicating this chip:
+#render_constraint_table(ecsm_chip, config, groups: "ecall")
+
+=== Read `xG`
+Once triggered, it loads register `x11` to see where $x_G$ is stored in memory (@ec:c:read_addr_xG) and subsequently load $x_G$ in (@ec:c:read_xG).
+Assumption @ec:a:addr_xG_alignment ensures no overflows happen when incrementing the address in @ec:c:read_xG.
+Note: `xG` is assumed to be range checked, since they're read from memory.
+#render_constraint_table(ecsm_chip, config, groups: "read_xG")
+
+=== Constrain `Gy`
```
RobinJadoul:
```suggestion
=== Constrain `yG`
```

For consistency

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+When $x_P = x_Q$ and $y_P eq.not - y_Q$, one instead uses $lambda = frac(3x_P^2, 2y_P, style: "horizontal")$.
+The remaing case that $(x_P, y_P) = (x_Q, -y_Q)$ corresponds with $Q = -P$; the addition results in $#inf$.
+
+#strong("Scalar multiplication.")
+An addition operation gives rise to an algorithm for scalar multiplication.
+Given curve point $P$ and scalar $k$, the multiple $k times P$ can trivially be computed as $P + P + ... + P$.
+This accelerator instead leverages the _double-and-add_ #footnote(link("https://en.wikipedia.org/wiki/Elliptic_curve_point_multiplication#Double-and-add")) technique, which utilizes only $O(log(k))$ additions for the full multiplication.
+
+#strong("This accelerator.")
+The purpose of this accelerator is to speed up the scalar multiplication $k times G$ for scalar $k in [1, N)$ and point $G in E(0, b, p) without {#inf}$ with $p in [2^248, 2^256)$.
```
RobinJadoul: Does it stop working for `p < 2^248`, or is it just wasteful?

spec/src/ec_scalar.toml
```diff
@@ -0,0 +1,117 @@
...
+name = "timestamp"
+type = "DWordWL"
+desc = "timestamp at which to serve the constant"
+pad = 0
+
+[[variables.input]]
+name = "ptr"
+type = "DWordWL"
+desc = "pointer to the first byte of the scalar"
+pad = 0
```
RobinJadoul: Why are we doing a bunch of MEMWs in this chip? The calling ECSM has already read all of K into byte limbs, so we can just take in the limb as input and decompose it, right?

I guess the reason is that EC_SCALAR recurses onto itself, reducing the interactions in the ECSM table, but I think for total area it'd be better feed in the bytes directly; at which point, there may be some tradeoff between how many bits EC_SCALAR does at once for width vs depth, but again probably total area tells us to just make it 256 wide?
In the case 256 columns wide wouldn't be the right call, just making multiple calls into EC_SCALAR from ECSM should nevertheless still be the same area cost as the recursion, but simpler because there's no recursion logic to be performed.

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+=== Subroutine
+With point $G$ and scalar $k$ fully constructed, we delegate bit-by-bit serving of the scalar `k` to the `EC_SCALAR` chip.
+Here, we capture the index of the most significant 1-bit of `k` in `len_k`.
+Note: if the prover decides to capture a lesser significant bit here, the LogUp will not balance, as the skipped bits will never taken off the bus.
+Next, we interact with the `ECDAS` chip, providing `G` both as the accumulator, and increment (@ec:c:start_double_add); we specifically instruct the chip to start with a _double_-operation.
+After completing its double-and-add sequence, the result is captured in `R` (@ec:c:receive_double_add).
+#render_constraint_table(ecsm_chip, config, groups: "delegate")
+
+=== Range check `xR`
+Before storing $x_R$, it is verified that $x_R in [0, p)$.
```
RobinJadoul, line 181: If this is not guaranteed on the output of ECDAS, is there any point in the recursion where this could cause issues?
Additionally, should we check `G` to be normalized mod `p` correctly?

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+  )
+$
+
+Recall that the addition of two curve points $A, B$ is treated differently based on three cases:
+#enum(indent: 1em, numbering: n => strong(raw("Case "+str(n)+".")),
+  enum.item[$x_A eq.not x_B$],
+  enum.item[$x_A eq x_B$ and $y_A eq.not -y_B$, or],
+  enum.item[$x_A eq x_B$ and $y_A eq -y_B$]
+)
+where _double_ may encounter the last two cases, while _add_ may encounter all three.
```
RobinJadoul: 1. How does double encounter the third case (except in a binary field)?
2. I think we can be certain that add steps are always "pure" add, if we only have a double-and-add to deal with (since we always have that `log_G(A) < log_G([2^t]G) = 2^t` when (conditionally) adding the doubled G to A at step `t`. (this is going by the usual way I look at double-and-add, this one seems to go from MSB to LSB instead, but a similar argument should then hold)

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+Therefore, the #ecsm and #ecdas chips were designed to avoid this case.
+To see how, note that #ecsm
++ is the sole chip that can "activate" the #ecdas chip by issuing an `ECDAS` lookup,
++ enforces that $G$ and the initial $A$ do not equal $#inf$, and
++ ensures $k in [1, N)$, where $N$ denotes the order of the curve.
+This combined yields that neither doubling $A$ or adding $A + G$ can produce $#inf$:
+
+*Double.*
+For $2A$ to equal $#inf$, the curve must have _even_ order.
+Since the order of the `secp256k1` curve is _odd_, such a point does not exist.
```
RobinJadoul: This is curve-specific, where the rest of the chips try to be curve-agnostic. The argument should hold for any prime-order subgroup in a curve, which _is_ the usual thing, but it still implies some extra limitations on the functionality of the accelerator.

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+#let ecdas = raw(ecdas_chip.name)
+
+The #ecdas chip (_Elliptic Curve Double-and-Add Sequence_) is responsible for accelerating the addition of two curve points, or the doubling of a single curve point. 
+More specifically, given curve points $A$ (accumulator) and $G$ (generator), and selector bit `op`, it performs the mapping
+$
+  (A, G) mapsto cases(
+    (A + A, &G) &text("if") #`op` = 0,
+    (A + G, &G) &text("if") #`op` = 1
+  )
+$
```
RobinJadoul, line 231: This feels similar to some of the "ladder" setups, like [https://eprint.iacr.org/2017/212.pdf, section 5.3]. I'm assuming always doing both double _and_ add may be a bit more costly though?
(Did any of the ladders or complete addition laws that we discussed work out in any reasonable way, btw?)

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+  enum.item[$x_A eq.not x_B$],
+  enum.item[$x_A eq x_B$ and $y_A eq.not -y_B$, or],
+  enum.item[$x_A eq x_B$ and $y_A eq -y_B$]
+)
+where _double_ may encounter the last two cases, while _add_ may encounter all three.
+Cases 2 and 3 may, for specific inputs, evaluate to $#inf$:
+a point that has no native short-Weierstrass representation.
+Therefore, the #ecsm and #ecdas chips were designed to avoid this case.
+To see how, note that #ecsm
++ is the sole chip that can "activate" the #ecdas chip by issuing an `ECDAS` lookup,
```
RobinJadoul: So this should maybe be codified as an assumption, so avoid any potential future consumers of ECDAS messing this up.

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+
+The #ecdas chip is comprised of #nr_variables variables that are expressed using #nr_columns columns and leverages #nr_interactions interaction(s):
+#render_chip_variable_table(ecdas_chip, config)
+
+== Constraints
+First, the chips receives the input for this double/add step:
+#render_constraint_table(ecdas_chip, config, groups: "receive")
+
+=== Operation switching
+The `op`-flag determines whether $R := 2A$ (0) or $R:= A+G$ (1).
```
RobinJadoul, line 271: Would splitting this into two separate dedicated chips with the same observable interactions not technically be better for area?

spec/ecsm.typ
RobinJadoul: Also, while not relevant to the ECDSA verification usecase that is the first priority, with an eye towards ZK, having e.g. a private key or nonce used in a scalar multiplication might leak information through table lengths, so that is something to keep in mind.

spec/src/ecsm.toml
```diff
@@ -0,0 +1,629 @@
...
+[[constraints.delegate]]
+kind = "interaction"
+tag = "BIT"
+input = ["timestamp", "len_k"]
+multiplicity = ["-", "μ"]
+
+[[constraints.delegate]]
+kind = "interaction"
+tag = "ECDAS"
+input = ["timestamp", "xG", "yG", "xG", "yG", ["-", "len_k", 1], 0]
```
RobinJadoul: I think this means you _always_ double, so we fail to correctly compute `[1]G`?
As discussed, this could also be due to `len_k` being weirdly defined

spec/src/ec_scalar.toml
RobinJadoul: There may even be some approach where we avoid the recursion of ECDAS as well, and simply let the EC_SCALAR chip emit the sequence of double/add steps that need to be performed. Not entirely sure if that wouldn't introduce too many more interactions atm, but it's potentially cleaner than having to reason about the recursion stopping.

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+  x_R &equiv lambda^2 - 2x_A &&mod p.\
+$
+Introducing non-negative witnesses $q''_0$ and $q''_1$, we convert these into
+$
+  2lambda y_A - 3x_A^2 + (#`r` - q''_0) p &= 0,\
+  lambda^2 - 2x_A - x_G - x_R + (#`r` - q''_1) p &= 0.\
+$
+#aside("The case of " + $y_A = 0$ + ".")[
+  Special attention should be paid to the first relation: if $y_A = 0$, $lambda$ can again be chosen freely.
+  As previously established, $y_A != 0$ for all points on the `secp256k1` curve.
```
RobinJadoul: For a specific curve, instead of the general situation.
The argument does hold for prime-order (or even odd-order) (sub)groups again too, since (x, 0) is a point of order 2.

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+$
+Introducing the non-negative witnesses $q'_0, q'_1$ and $q_2$, we can convert these relations into
+$
+  lambda (x_G - x_A) - y_G + y_A + (#`r` - q'_0) p &= 0,\
+  lambda^2 - x_A - x_G - x_R + (#`r` - q'_1) p &= 0,\
+  lambda (x_A - x_R) - y_A - y_R + (#`r` - q_2) p &= 0,\
+$
+for some $r in NN$ to be fixed later.
+
+#aside("The case of " + $x_A = x_G$ + ".")[
```
RobinJadoul: This starts to feel a bit repetitive, perhaps we should gather all these arguments (that also refer to the previous ones) in a single place.
For instance, this aside is essentially only arguing that `A = ±G => op = "double"`, in which case λ will be constrained differently

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+  The remaining case $A=G$ is the intial state pushed onto the LogUp by #ecsm (@ec:c:start_double_add), with `op`-flag set to $0$ (_double_), not `add`.
+  Hence, this situation cannot occur.  
+]
+
+We rewrite the relations to find
+$
+  q'_0 &= #`r` + p^(-1) dot (lambda (x_G - x_A) - y_G + y_A),\
+  q'_1 &= #`r` + p^(-1) dot (lambda^2 - x_A - x_G - x_R),\
+  q_2  &= #`r` + p^(-1) dot (lambda (x_A - x_R) - y_A - y_R)\
+$
```
RobinJadoul, line 277: Worth clarifying that by construction, this division must hold. I got confused here for a while by assuming we needed to be working mod some power of 2 for `p^-1` to even exist.
In this case, it's actually division and maybe the `p^-1` being there as a "number" is even confusing.

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+from which we can conclude that $q'_0, q_2 in (#`r`-p, #`r`+p)$ and $q'_1 in (#`r`, #`r` + p)$.
```
RobinJadoul: Don't we get something like `q'_1 >= r - 3(p - 1)/p = r - 3 + 3/p > r - 3`? And some similar off-by-a-few potentially in the other bounds?

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+  curve-constants $a$, $b$, $p$, $r$ and $N$, and look them up when a scalar multiplication selects them.
+  The selection procedure could be done through the `ECALL` number; the #ecsm chip would accept multiple numbers, setting an internal "curve-selector" field accordingly.
+- Transitioning from `U256BL`s to `U256HL`s would roughly halve the number of columns in both the #ecsm and #ecdas chips.
+  This would likely require increasing the sizes of the carries from 16 to 24 bits.
+  Since the carries need to be range checked, one would have to investigate whether
+    - it would be possible to perform a 24-bit range-check lookup,
+    - one could set up a 24-bit range-check table. This could be as narrow as two columns.
+    - have some hybrid version, where there is a native lookup table for x-bits, and a dynamic table for outliers (high carries are not encountered frequently).
+
+= Discussing the carries <ecsm-limb_carry>
```
RobinJadoul: We may also want to formalize the correctness of the "range check a carry" approach by itself a bit more generally (presumably in ADD where it is first used?) so that we can refer to that and not have to think too much about when it is correct or not.
I think the general statement would be something like a range check on the range `[0, R)` for a carry `2^k * c` being correct as long as `R * c < p_BaseField`, since then we can guarantee that the carry calculation holds over the integers

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+To constrain `x2` and $y_G$ in #ecsm, and $lambda$, $x_R$ and $y_R$ in #ecdas, we use (variations of) the same technique:
+- multiplications are performed limb-by-limb, 
+- a set of carry-limbs is used to exchange the underflow/overflow from one limb to another, and
+- the carry limbs are range constrained to ensure only one output value is possible.
+
+We now explore this carry-technique and provide some proofs.
+
+== Lemma 1
```
RobinJadoul: Feels a bit weird to have manual lemma/corollary numbering in the a title. Probably best to switch to some typst package for those, if we're intending to formalize/prove more anyway.

spec/src/ecdas.toml
```diff
@@ -0,0 +1,503 @@
...
+
+[[constraints.yR]]
+kind = "arith"
+constraint = "$#`c2`_63 = 0$"
+poly = ["idx", "c2", 63]
+
+[[constraints.yR]]
+kind = "interaction"
+tag = "IS_HALF"
+input = [["+", ["idx", "c2", "i"], 16320]]
```
RobinJadoul: Can we maybe move these magic constant columns into a constant column of the chip? It's nicer to have them all in the same place to check correspondence with the analysis and make sure everything is up-to-date if we do change the analysis at some point (e.g. because we get Half limbs)

spec/ecsm.typ
```diff
@@ -0,0 +1,452 @@
...
+from which we can conclude that $q'_0, q_2 in (#`r`-p, #`r`+p)$ and $q'_1 in (#`r`, #`r` + p)$.
+When doubling, only the formulae for $lambda$ and $x_R$ are different:
+$
+  lambda &equiv (3x_A^2)/(2y_A) &&mod p,\
+  x_R &equiv lambda^2 - 2x_A &&mod p.\
+$
+Introducing non-negative witnesses $q''_0$ and $q''_1$, we convert these into
+$
+  2lambda y_A - 3x_A^2 + (#`r` - q''_0) p &= 0,\
+  lambda^2 - 2x_A - x_G - x_R + (#`r` - q''_1) p &= 0.\
```
RobinJadoul: Having `x_G` in there looks wrong and propagates a bit further

spec/ecsm.typ
RobinJadoul: After thinking more, it's _probably_ fine, since everything is constrained to be correct over the integers, so having some multiple of p as additive error shouldn't hinder correctness.
That being said, for specific structure of `p` (such as the `secp256k1` one), there may be more efficient limb-wise checks possible

spec/ecsm.typ
RobinJadoul: After reading more, yes, the always double _and_ add is very likely more expensive, since we can multiplex quite a bit in ECDAS (at the cost, maybe, of less tight range analysis)

spec/src/ecdas.toml
```diff
@@ -0,0 +1,509 @@
...
+kind = "template"
+tag = "IS_BIT"
+input = ["next_op"]
+ref = "ecdas:c:range_next_op"
+
+[[constraints.send]]
+kind = "interaction"
+tag = "BIT"
+input = ["timestamp", "round"]
+multiplicity = ["-", "next_op"]
```
RobinJadoul: I'd keep the positive here and the negative in `EC_SCALAR`, to keep with the usual interpretation of μ = -1 meaning that you've "shown" the relation to hold

spec/src/ecdas.toml
```diff
@@ -0,0 +1,509 @@
...
+
+
+[[constraint_groups]]
+name = "send"
+
+[[constraints.send]]
+kind = "template"
+tag = "IS_BIT"
+input = ["next_op"]
+ref = "ecdas:c:range_next_op"
```
RobinJadoul, line 532: The LogUp should take care of it, but probably good to still bit-check `op` itself too

RobinJadoul (commented):
- I do wonder how much potential optimization we're leaving on the table by treating `p` as a generic prime, rather than using the structure it has
- A more descriptive name for the `BIT` and `SERVE_K` interactions may be in order
- The complex arith constraints for equalities mod p tend to have a different order between their "constraint" and "poly", which makes checking correspondence a bit tedious and error-prone
- I have some thoughts about the presentation of the "theory"/background, but trying to focus on the more technical things first

spec/src/ecsm.toml
erik-3milabs: > How safe is this?

I seem to recall that one of our strategies is to range check (address, value, timestamp) for all writes. As a result, all reads can now be assumed safe.
When we adhere to this rule, I believe the tagged assumption to be safe.

_Reasoning_:
Suppose `addr_xG` (or `addr_k` or `addr_R` for that matter) violates the tagged assumption.
That would mean that at least one of the interactions in `ec:c:read_xG` would read a value from an invalid address.
To balance the LogUp, one would have had to (previously) perform a write-only interaction for this same, illegal address.
Under the assumption that all writes range check their address, this should be impossible.

Note: performing a read&write-interaction would not cut it, as this would just move the problem.

> that is pushed down into the guest program

You're right that this is indeed pushed down to the guest program.

Note that this is not the first time this is done; the SHA256-C3 and SHA256-C7 essentially leverage the same trick: if the addresses for `h_addr` and `m_addr` don't align, C2 and C6 will fail.

spec/src/ecsm.toml
erik-3milabs: Yeah, this should be due to `len_k` being weirdly defined. (But please double check!)

I renamed it as `idx_k`. I'm open to other suggestions.

spec/ecsm.typ
erik-3milabs: In a previous version, `p < 2^248` was problematic. Now it is just wasteful. I've modified it to be `p < 2^256`.

spec/ecsm.typ
erik-3milabs: > If this is not guaranteed on the output of `ECDAS`, is there any point in the recursion where this could cause issues?

I don't believe so. In designing, I've worked with the assumption that $x_A, y_A, x_G, y_G in [0, 2^256)$.

> That being said, for specific structure of p (such as the secp256k1 one), there may be more efficient limb-wise checks possible

That may very well be. In designing this chip, I've tried to keep things as generic as possible, including (trying to) make no assumptions on the structure of `p`.

spec/ecsm.typ
erik-3milabs: > This feels similar to some of the "ladder" setups, like [https://eprint.iacr.org/2017/212.pdf, section 5.3]. I'm assuming always doing both double and add may be a bit more costly though?

> After reading more, yes, the always double and add is very likely more expensive, since we can multiplex quite a bit in ECDAS 

Let's run the numbers (for posterity). One would have to 
1. introduce $\geq 384$ base field columns to store $λ'$ and its corresponding $c'_0, q'_0$ as well as the $x'_R$ and $y'_R$ and their corresponding $c1, q'_1$ and $c2, q'_2$ from the second operation,
2. introduce $384/4 = 96$ (note: 2*IS_BYTE => ARE_BYTES + degree-2 batching) extension field columns to store the additional LogUp contributions for the range checks.

i.e., we would be introducing equivalent to $384 + 3*48 = 672$ base field columns, which is a 40% increase compared to the $521+3×388 = 1685$ the chip currently has. The upside, of course, would be that the number of rows reduces from $256 + Hw(k)$ to $256$, which (for uniformly random $k$) results in a ~33% size decrease on average. Note: this benefit decreases even further when $k$ has many leading zeroes, as `ECDAS` skips all of those.

So, yes, always double-and-adding would (on average) be more expensive than the current design.

> Did any of the ladders or complete addition laws that we discussed work out in any reasonable way, btw?

I will admit that I didn't look long & hard at it. A quick analysis suggested that none of the systems had multiplication-counts that were remotely close to this "basic" one. Given that I was able to make this one work, I decided to shelve the complete laws for now. Still, those may come in useful if at any point we want to have a constant-area design, or want to support arbitrary point addition.

> (at the cost, maybe, of less tight range analysis)

wdym by this?

spec/ecsm.typ
erik-3milabs: > 1. How does double encounter the third case (except in a binary field)?

It would not. I removed the sentence. 

> 2. I think we can be certain that add steps are always "pure" add, if we only have a double-and-add to deal with [...] (this is going by the usual way I look at double-and-add, this one seems to go from MSB to LSB instead, but a similar argument should then hold)

I believe I provide this 'similar argument' below. Please let me know if the rewritten paragraph sufficiently clarifies the point I'm trying to make.

> Also, while not relevant to the ECDSA verification usecase that is the first priority, with an eye towards ZK, having e.g. a private key or nonce used in a scalar multiplication might leak information through table lengths, so that is something to keep in mind.

I'm well aware. This is why I made sure to ask Mauro about a couple weeks ago. This design is trying to be as compact as possible, and is "variable-area" as a result.

spec/ecsm.typ
erik-3milabs: I think this depends on one's definition of _technically_:

I believe that your proposal would result in two chips that are nearly identical to this one, but just without the `op` column. Given the width of these chips (>2^10 column-equivalent), selecting a (reasonable) table length of size $2^{10+k}$ for some $k \geq 1$, would effectively cancel this benefit as one would now have an additional chip that needs to be padded to length $2^{10+k}$; just $2^k$ padding rows would suffice to achieve a larger area.

spec/src/ec_scalar.toml
erik-3milabs: >Why are we doing a bunch of MEMWs in this chip? The calling ECSM has already read all of K into byte limbs, so we can just take in the limb as input and decompose it, right?
> I guess the reason is that EC_SCALAR recurses onto itself, reducing the interactions in the ECSM table, but I think for total area it'd be better feed in the bytes directly; 

I ran the numbers, and you're right: switching would save 32 interactions and 160 cells per `ECSM`-call.

> at which point, there may be some tradeoff between how many bits EC_SCALAR does at once for width vs depth, but again probably total area tells us to just make it 256 wide?
> In the case 256 columns wide wouldn't be the right call, just making multiple calls into EC_SCALAR from ECSM should nevertheless still be the same area cost as the recursion, but simpler because there's no recursion logic to be performed.

You're right: Making EC-SCALAR work on all 256 bits at once would reduce the cell-footprint by another ~15% (from 320 to 257).

In fact, we can go even one step further, and merge `EC-SCALAR` into `ECSM`. This would save 2 interactions (the transfer of `k` from ECSM to EC-SCALAR, on both chips) and 33 cells (= multiplicity on `EC-SCALAR` + the byte representation of `k` on `ECSM` can be virtualized from the individual bits).

The thing I keep returning to, is whether _padding_ isn't screwing with minimizing total area as a heuristic. I'll spend some time this afternoon to investigate this for a bit.

spec/ecsm.typ
erik-3milabs: I spun this off as a separate PR: #669

spec/ecsm.typ
erik-3milabs: I've rewritten the section in a way that these statements are now no longer included. Let me know whether you agree the new version is better, or whether you believe the derivation for the ranges of `q` should be reincluded

spec/ecsm.typ
erik-3milabs: Let me know what you think of the (more concise) rewrite.

spec/src/ecsm.toml
RobinJadoul: I think the SHA256 version actually works regardless of the alignment?
SHA256-C3 and SHA256-C7 both use a full `ADD` so there's no requirement on being small enough, since `ADD` can deal with overlow correctly. And if it's not dword-aligned, then you get the fallback to slow version of MEMW, but besides performance, there's no safety or correctness assumption on the guest that I see there.
Please do correct me if I'm reading that wrong.

I think it is just a completeness thing and indeed not soundness as you argue, as long as we keep the initialization of addresses correctly range-checked, which is already a hard requirement due to MEMW.
We _can_ make this assumption, but then I would put it more visibly, as this becomes critical knowledge for consumers, and not just for implementing the spec.

spec/ecsm.typ
RobinJadoul: > I've tried to keep things as generic as possible, including (trying to) make no assumptions on the structure of p.

Maybe just something to make note of as a potential optimization area.

RobinJadoul: I'm not entirely sure what to do with the range check on `G`.
I don't like that it would require another 16 columns, but it may still be the right thing to do out of precaution.

spec/ecsm.typ
RobinJadoul: > (at the cost, maybe, of less tight range analysis)

A loose thought that because of the multiplexing, we need to consider the worst case scenario for the size of the carry, whereas if there's a dedicated operation, you end up with just the analysis of the operation itself

spec/ecsm.typ
RobinJadoul: The comment was made before I had fully looked at everything; and before seeing how much manages to get stuffed in shared columns, I was expecting slightly more non-overlapping pieces than only the op column

spec/src/ec_scalar.toml
erik-3milabs: Since we concluded that padding isn't an issue, I inlined EC_SCALAR into ESCM

spec/src/ecsm.toml
erik-3milabs: 1. You're right: SHA is alignment-agnostic
2. As discussed, we're sticking with having no soundness gaps for now. I updated the spec accordingly.

spec/ecsm.typ
erik-3milabs: > I'm not entirely sure what to do with the range check on G.
I don't like that it would require another 16 columns, but it may still be the right thing to do out of precaution.

I don't think we need it, but honestly, at this point 16 columns is a drop in the bucket (+2.3%). I'll just add it.

spec/ecsm.typ
erik-3milabs: The range for the carries in the DOUBLE case is a superset of those in the ADD range; I don't think demuxing is going to help here.

spec/ecsm.typ
erik-3milabs: As decided by Mauro, I again removed it.

spec/ecsm.typ
```diff
@@ -0,0 +1,367 @@
...
+#attention("Variable space.")[
+    This accelerator is _variable-space_ in the value of $k$; different values of $k$ may result in different table sizes.
+    As such, *this accelerator should only be used for input sets with public $k$.*
+]
+
+The accelerator comprises two chips:
+- *`ECSM` (Elliptic Curve Scalar Multiply)*.
+    This chip is responsible for 
+    - loading $k$ from memory and verifying that it is contained in $[1, N)$,
+    - loading inputs $x_G$ and reconstructing $y_G$,
```
RobinJadoul: and verifying x_G < p

spec/ecsm.typ
```diff
@@ -0,0 +1,367 @@
...
+- Transitioning from `U256BL`s to `U256HL`s would roughly halve the number of columns in both the #ecsm and #ecdas chips.
+  This would likely require increasing the sizes of the carries from 16 to 24 bits.
+  Since the carries need to be range checked, one would have to investigate whether
+    - it would be possible to perform a 24-bit range-check lookup,
+    - one could set up a 24-bit range-check table. This could be as narrow as two columns.
+    - have some hybrid version, where there is a native lookup table for x-bits, and a dynamic table for outliers (high carries are not encountered frequently).
+- `addr_xG[0]`, `addr_k[0]` and `addr_xR[0]` could be `DWordWL`s rather than `HL`s.
+  We use `HL`s as conventient notation.
+  This modification saves 6 columns.
+- the design of these chip is generic, and makes no assumptions on the parameters $b$, $p$ and $N$.
```
RobinJadoul:
```suggestion
- the design of these chip is generic, and makes no assumptions on the parameters $a$, $b$, $p$ and $N$.
```

spec/src/ecsm.toml
```diff
@@ -0,0 +1,753 @@
...
+type = ["Byte", 64]
+desc = "zero-extension of `p`"
+def = {polys=[
+    {iter=["i", 0, 31], poly=["idx", "p", "i"]},
+    {iter=["i", 32, 63], poly=0},
+]}
+
+[[variables.constant]]
+name = "id"
+type = "Bit"
```
RobinJadoul, line 220: Technically not a bit if we want it to be extensible

spec/ecsm.typ
```diff
@@ -0,0 +1,367 @@
...
+  This means that enforcing the above relation does not fully constrain the prover: it can choose which of the two to provide as hint.
+  However, in this setting, this is not a problem: the `ECSM`-chip only outputs the $x$-coordinate of $k times G$, which is the same, irrespective of the chosen root.
+]
+
+Below, we enforce the first of the two sub-relations.
+We emphasize here that @ec:c:c0_63_is_zero is required to ensure the sum evaluates to $0$, rather than just $0 mod 2^256$.
+#render_constraint_table(ecsm_chip, config, groups: "xG2")
+
+Next, we restrict the witness pair $(y_G, #`q1`)$.
+Note there that @ec:c:c1_0 and @ec:c:c1_i multiply `B` and `P` by `μ` to simplify the padding; there are no other side-effects to this since $#`μ` = 1$ on non-padding rows (@ec:c:mu_isbit).
```
RobinJadoul, line 161: I think we're trading off padding simplicity for degree 3 constraints where otherwise we only need degree 2. Minor, but potentially worth a mention in the notes

spec/ecsm.typ
```diff
@@ -0,0 +1,367 @@
...
+After reading `addr_k` from `x12` (@ec:c:read_addr_k), we read `k` from this address (@ec:c:load_k).
+To prevent the point at infinity from showing up during the scalar multiplication, we require that $#`k` < #`N`$.
+This is achieved by requiring that the addition $#`N` + (#`k` - #`N`)$ overflows $mod 2^256$ (@ec:c:k_lt_N).
+Additionally, @ec:c:k_gt_0 ensures that $#`k` > 0$, preventing a case where $#`k` times #`G` = #inf$.
+#render_constraint_table(ecsm_chip, config, groups: "verify_k")
+
+=== Subroutine
+With point $G$ and scalar $k$ fully constructed, we serve scalar `k` bit-by-bit to the `ECDAS` chip.
+On this chip, we do capture the index of the most significant 1-bit of `k` in `idx_k`, to instruct the `ECDAS` chip where to start.
+Note: if the prover decides to capture a lesser significant bit here, the LogUp will not balance, as the skipped bits will never taken off the bus.
```
RobinJadoul, line 175: But what about a _greater_ `idx_k`? We'd probably need to check that `k[idx_k] = 1`, which seems annoying.

spec/ecsm.typ
```diff
@@ -0,0 +1,367 @@
...
+$
+in terms of the positive and negative components to find
+$
+  #`xG`^2 - (#`x2` + q_0 dot p) &= 0, text("and")\
+  (y_G^2 + 2p^2) - (x_G dot #`x2` + a dot x_G + b + q_1 dot p) &= 0.
+$
+Applying @limbs:cor:carry-upper-bound with $(L, n) = (2^8, 32)$, we find that
+$
+  #`c0`_i &in [-8160, 8159],\
+  #`c1`_i &in [-24477, 24478].\
```
RobinJadoul: I think you're off by a few from what I tried to compute quickly, but it's wider than what I got, so should be safe anyway

spec/ecsm.typ
```diff
@@ -0,0 +1,367 @@
...
+which we rewrite in terms of the positive and negative components to find
+$
+  (#`op` dot (lambda x_G + y_A) + (1-#`op`) dot 2lambda y_A + #`r` dot p) - (#`op` dot (lambda x_A + y_G) + (1-#`op`) (3x_A^2 + a) + q_0 dot p) &= 0,\
+  (lambda^2 + #`r` dot p) - (x_A + x_G + x_R + (1-#`op`) (x_A - x_G) + q_1 dot p) &= 0,\
+  (lambda x_A + #`r` dot p) - (lambda x_R + y_A + y_R + q_2 dot p) &= 0.
+$
+Leveraging @limbs:cor:carry-upper-bound with $(L, n) = (2^8, 33)$ and maximizing for the value of $#`op` in {0, 1}$, we gather
+$
+  #`c0`_i &in [-33657, 25242],\
+  #`c1`_i &in [-8416, 16828], text("and")\
```
RobinJadoul, line 349: I got -8418 (μ=1, α = 5, δ=1)

spec/ecsm.typ
```diff
@@ -0,0 +1,367 @@
...
+  #`c2`_i &in [-16830, 16828].\
+$
+Selecting $#`carry_offsets` = (33657,8416,16830)$, we arrive at
+$
+  #`c0`_i + #`carry_offsets[0]` &in [0, 58899] subset.eq [2^16],\
+  #`c1`_i + #`carry_offsets[1]` &in [0, 25244] subset.eq [2^16],text("and")\
+  #`c2`_i + #`carry_offsets[2]` &in [0, 33688] subset.eq [2^16].
```
RobinJadoul: I think 33658

RobinJadoul (commented):
Mostly concerned about the idx_k thing, the rest is either a quick edit or double checking calculations

erik-3milabs:
@claude

spec/ecsm.typ
RobinJadoul: nvm, we're forcing that with `BIT[ts, idx_k]/μ`

spec/ecsm.typ
```diff
@@ -0,0 +1,369 @@
...
+
+#render_constraint_table(ecsm_chip, config, groups: "range_xR")
+
+=== Write `xR`
+We read `addr_xR` from register `x10` (@ec:c:load_addr_xR), and subsequently write `xR` to this address (@ec:c:write_xR).
+Note that the `timestamp` on both memory accesses is offset to allow `addr_xR` to equal `addr_xG` and thus for $x_R$ to overwrite $x_G$ in memory.
+#render_constraint_table(ecsm_chip, config, groups: "write_xR")
+
+== Carry offsets
+We close by deriving the values of `carry_offsets`.
```
RobinJadoul: naming mismatch: `offsets` or `carry_offsets`

spec/ecsm.typ
RobinJadoul:
```suggestion
  #`c1`_i &in [-24478, 24477].\
```

spec/ecsm.typ
```diff
@@ -0,0 +1,369 @@
...
+$
+  #`xG`^2 - (#`x2` + q_0 dot p) &= 0, text("and")\
+  (y_G^2 + 2p^2) - (x_G dot #`x2` + a dot x_G + b + q_1 dot p) &= 0.
+$
+Applying @limbs:cor:carry-upper-bound with $(L, n) = (2^8, 32)$, we find that
+$
+  #`c0`_i &in [-8160, 8159],\
+  #`c1`_i &in [-24477, 24478].\
+$
+When we selectc $#`carry_offsets` = (8160, 24477)$, we arrive at
```
RobinJadoul:
```suggestion
When we selectc $#`carry_offsets` = (8160, 24478)$, we arrive at
```

RobinJadoul (commented):
Decided to quickly script the carry calculations to double check everything.
```py
def M(n, L, mu, alpha):
    assert L >= 4
    delta = int(mu < alpha)
    return mu * n * (L - 1) + alpha - mu - delta

def bounds(n, L, mu_pos, alpha_pos, mu_neg, alpha_neg):
    min = -M(n, L, mu_neg, alpha_neg)
    max = M(n, L, mu_pos, alpha_pos)
    print(f"[{min}, {max}] => [0, {max - min}]")


print("ECSM")
bounds(32, 256, 1, 0, 1, 1)
bounds(32, 256, 3, 0, 3, 1)

print("\nECDA")
bounds(33, 256, 3, 0, 4, 1)
bounds(33, 256, 2, 0, 1, 3)
bounds(33, 256, 2, 0, 2, 2)
```

RobinJadoul (approved):

PR #658, opened 2026-06-10
feat(cuda): GPU batch inverse

ColoCarletti:
## Summary

  Replaces the two remaining CPU `inplace_batch_inverse` sites in the prover with a fused on-device compute+invert pipeline. R3 OOD's per-eval-point `barycentric_inv_denoms` and R4 DEEP's `(x_i − z_k)` denominators now both flow through a single multi-block Hillis-Steele scan kernel and return device-resident `CudaSlice<u64>` handles that downstream dispatchers (`try_barycentric_*_on_handle`, `try_deep_composition_gpu`) read directly.

  Wall-clock parity on `fib_iterative_4M` on a 46-core host (savings overlap with existing GPU work). The win is in PCIe traffic (~6 GB of redundant H2D removed per  prove), peak host RSS at R4 (~288 MB removed), and the architectural primitives (`Arc<CudaStream>` threading, `DenomSign` enum, public `batch_inverse_ext3_dev`  API) that future device-resident extensions can reuse. Pays off proportionally on smaller hosts, larger traces, more eval points, and more tables.

  ## Changes

  - **`crypto/math-cuda/kernels/inverse.cu`** — 6 kernels: sign-flagged `compute_denoms_ext3` (R3 `z − x` vs R4 `x − z`), forward and reverse multi-block scans,
  apply-offsets passes, `batch_inverse_combine_ext3`.
  - **`crypto/math-cuda/src/inverse.rs`** — host driver: `batch_inverse_ext3` (parity-test path), `batch_inverse_ext3_dev` (device→device), fused
  `compute_and_invert_denoms_ext3_dev` with the `DenomSign` enum, recursive scan driver, one ext3 Fermat inverse on host per batch.
  - **`crypto/math-cuda/src/{barycentric,deep}.rs`** — `_with_dev_inv_denoms` variants that take a buffer + offset + caller's stream and slice internally (no
  per-call H2D, no cross-stream race).
  - **`crypto/stark/src/gpu_lde.rs`** — `try_compute_and_invert_inv_denoms_dev` + `try_inv_denoms_dev_with_stream` (acquires backend + stream). Threads
  `Option<(&CudaSlice<u64>, usize, &Arc<CudaStream>)>` through the barycentric and DEEP dispatchers. New `gpu_batch_invert_calls` counter.
  - **`crypto/stark/src/{trace,prover}.rs`** — R3 OOD and R4 DEEP fast paths. CPU `barycentric_inv_denoms` is now lazy on the all-GPU happy path; the CPU `denoms`
  Vec in `compute_deep_composition_poly_evaluations` is only built when the dev-inv-denoms path returns `None`.
  - **Tests** — `batch_inverse.rs` (parity, n in {1, 2..256, 257..1024, 4096..2^16, 2^18, 2^20, 2^22}); `compute_and_invert_denoms.rs` (parity parametrised over both
   signs, R3 and R4 shapes); `invert_ext3_host` parity against `Degree3GoldilocksExtensionField::inv`; `cuda_path_integration` asserts the new counter fires;
  `cuda_fallback_tests::gpu_batch_invert_fault_falls_back_to_cpu` validates the CPU fallback under injected cudarc errors.

  ## Fallback

  Every dispatch returns `None` on TypeId mismatch (non-Goldilocks / non-ext3), below threshold, or any cudarc error. The caller falls through to the existing CPU `inplace_batch_inverse` path with no state change. Runtime-validated by the new fault-injection test.


crypto/math-cuda/kernels/inverse.cu
```diff
@@ -0,0 +1,306 @@
...
+    // Write per-element scan result.
+    if (gid < n) {
+        scan_out[gid * 3 + 0] = shmem[tid].a;
+        scan_out[gid * 3 + 1] = shmem[tid].b;
+        scan_out[gid * 3 + 2] = shmem[tid].c;
+    }
+
+    // Block total = scan value at the last VALID thread of this block.
+    // Two cases: full block (tid == 255) or partial last block (gid == n-1).
+    bool is_last_in_block = (tid == BLOCK_SIZE - 1) || (gid == n - 1);
```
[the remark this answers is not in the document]
ColoCarletti: fix in https://github.com/yetanotherco/lambda_vm/pull/658/changes/d73219ef46e580cff0d8453a75ed98df568c7be5

crypto/stark/src/gpu_lde.rs
```diff
@@ -1164,69 +1234,102 @@ where
...
+            for (p, col) in parts_host.iter().enumerate() {
+                let slice = unsafe { ext3_slice_to_u64::<E>(col) };
+                for (r, chunk) in slice.chunks_exact(3).enumerate() {
+                    packed[(p * 3) * lde_size + r] = chunk[0];
+                    packed[(p * 3 + 1) * lde_size + r] = chunk[1];
+                    packed[(p * 3 + 2) * lde_size + r] = chunk[2];
+                }
             }
+            parts_host_packed = packed;
+            // Host inv_denoms required when going through this path; we
```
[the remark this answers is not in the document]
ColoCarletti: fix in https://github.com/yetanotherco/lambda_vm/pull/658/changes/d73219ef46e580cff0d8453a75ed98df568c7be5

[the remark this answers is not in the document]
ColoCarletti: fixed in d73219ef

crypto/math-cuda/kernels/inverse.cu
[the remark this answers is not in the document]
ColoCarletti: fixed in d73219ef

diegokingston (approved):

crypto/stark/src/gpu_lde.rs
```diff
@@ -1238,6 +1347,166 @@ where
     Some(u64_to_ext3_vec::<E>(&deep_raw))
 }
 
+/// Build `inv_denoms[k*n + i] = 1 / (lift(coset_base[i]) - z_scalars[k])`
+/// entirely on device. Used by both R3 OOD (n = trace_size, k_scalars =
+/// num_eval_points) and R4 DEEP (n = lde_size, k_scalars = 1 +
+/// num_eval_points). Returns a device handle the caller can slice and
+/// thread into downstream dispatchers without ever D2H'ing the inverted
+/// values; on type / threshold / cudarc failure returns `None` so the
```
Oppen, line 1355: Shouldn't type failures be caught by the type system?

Oppen:
> Wall-clock parity on `fib_iterative_4M` on a 46-core host (savings overlap with existing GPU work). The win is in PCIe traffic (~6 GB of redundant H2D removed per prove), peak host RSS at R4 (~288 MB removed), and the architectural primitives (`Arc<CudaStream>` threading, `DenomSign` enum, public `batch_inverse_ext3_dev` API) that future device-resident extensions can reuse. Pays off proportionally on smaller hosts, larger traces, more eval points, and more tables.

I'm confused. If the optimization works and takes away 6GB of H2D transfers, shouldn't wall-clock be better than just parity?

crypto/math-cuda/kernels/inverse.cu
```diff
@@ -0,0 +1,313 @@
...
+        // x - z: lift x to (x, 0, 0), subtract z.
+        d.a = goldilocks::sub(x_i, z.a);
+        d.b = goldilocks::neg(z.b);
+        d.c = goldilocks::neg(z.c);
+    }
+
+    denoms_out[flat * 3 + 0] = d.a;
+    denoms_out[flat * 3 + 1] = d.b;
+    denoms_out[flat * 3 + 2] = d.c;
+}
```
Oppen, line 79: Many of the index multiplications could be hoisted outside of the function to become mere additions.
The choice of `subract_x = 0` meaning "actually DO subtract x" is counter-intuitive.

ColoCarletti: fixed in: https://github.com/yetanotherco/lambda_vm/pull/658/changes/9a8f355bfd8e1ac697d09e0a0e3a37fe786a91c3

crypto/stark/src/gpu_lde.rs
ColoCarletti: Generic F, E so the function fits the rest of the prover. The kernels only handle Goldilocks + ext3 though, and I don’t know if Rust have a clean way to express "only when  F == Goldilocks" without a sealed marker trait threaded through all call sites. Every other dispatcher in this file (R1/R2/R3/R4/FRI) uses the same TypeId pattern happy to switch the whole file, but I think its out of scope here.

ColoCarletti:
> > Wall-clock parity on `fib_iterative_4M` on a 46-core host (savings overlap with existing GPU work). The win is in PCIe traffic (~6 GB of redundant H2D removed per prove), peak host RSS at R4 (~288 MB removed), and the architectural primitives (`Arc<CudaStream>` threading, `DenomSign` enum, public `batch_inverse_ext3_dev` API) that future device-resident extensions can reuse. Pays off proportionally on smaller hosts, larger traces, more eval points, and more tables.
> 
> I'm confused. If the optimization works and takes away 6GB of H2D transfers, shouldn't wall-clock be better than just parity?

Batch invert isn't where the prover spends its time. The dominant CPU costs are constraint evaluation, trace expansion and logup, none of which this PR touches. The savings here sat inside slack the rest of the pipeline already had on a 46-core host.

The point of this PR is the primitives it ships: `batch_inverse_ext3_dev` is exactly what GPU logup needs (logup = `1/(α - bus)` per row), and the `Arc<CudaStream>` threading lets a future GPU constraint evaluator feed R2 without H2D. That's where the considerable wins should show up.

crypto/stark/src/gpu_lde.rs
Oppen: I believe you can do that with a const assertion if the field has access to the prime, e.g.:
```rust
const { assert!(F::P == Golidlocks::P, "GPU only implements Goldilocks field") };
```

Oppen: You probably can do it in a more type-theoretic-friendly way, but it will probably be a mess. This is a one-liner that catches some issues at compile-time.

Oppen (approved):
LGTM.
Re: typing: you could just implement for `FieldElement<GoldilocksField>` and call `<F as Any>::downcast()` or similar on the topmost caller. It's too big a change now that the generic accumulated and this works now, so I believe it's better to merge as is and revisit later.

PR #666, opened 2026-06-12
ethrex integration: Crypto-trait guest + synthetic block fixture generator

ColoCarletti:
### Description

⚠️ **Temporary.** This pins our ethrex guest to the *unmerged* LambdaVM-backend ethrex PR branch and adds an in-repo fixture generator as a stand-in. Once that PR merges to ethrex `main`, we re-pin to the merge commit and switch fixture generation to `ethrex-replay`. The in-repo generator is then removed.

### What this does
- **Migrates the ethrex guest to the new Crypto-trait API** (`execution_program(input, LambdaVmCrypto)`); keccak runs through our precompile. Removes the old `[patch] ethrex-crypto` and c-kzg.
- **Adds `tooling/ethrex-fixtures`** — a small offline tool that builds ethrex blocks with an arbitrary number of ETH transfers and writes the serialized `ProgramInput`. Pinned to the same ethrex rev as the guest so fixtures stay compatible.
- **Regenerates the committed fixtures** (empty / 1-tx / 10-tx) at the new rev and wires empty + 1-tx into the ethrex benchmark.
- Updates the executor native test reference + docs.

### Current limitations
- Blocks/contracts that use the KZG point-evaluation precompile (`0x0a`) are unsupported under the LambdaVM guest today. The `lambdavm` feature does not wire a `kzg-rs`/`c-kzg` backend, so those calls fail closed/revert. The committed temporary fixtures are ETH-transfer-only and do not exercise this; wire KZG before switching to `ethrex-replay` or arbitrary real blocks.


executor/tests/ethrex_10_transfers.bin
Oppen, line 1: Is there any way to get this in a text representation and build the binary from it?

.github/workflows/pr_main.yaml
```diff
@@ -96,7 +96,6 @@ jobs:
 
       - name: Run ignored executor tests
         run: |
-          make prepare-test-data
```
Oppen, line 99: Why is this removed?

MauroToscano: `prepare-test-data` only existed to download the old `executor/tests/ethrex_hoodi.bin` fixture. This PR stops using that fixture: `test_ethrex` now reads the committed synthetic `ethrex_10_transfers.bin`, the empty/1-tx fixtures are also committed and reproducible via `tooling/ethrex-fixtures`, and `test_ckzg` does not depend on Hoodi.

The target is removed from the `Makefile` in the same PR, so leaving this call here would make the ignored-test step fail with `No rule to make target prepare-test-data`. No data-prep step is needed for the ignored executor tests anymore.

MauroToscano: The data is in the repo now, and the target doesn't exist anymore

Oppen (approved):

MauroToscano (approved):

PR #669, opened 2026-06-15
Spec: limb decomposition discussion

erik-3milabs:
This PR introduces a section that discusses (and proves) some properties of performing operations on limb decompositions.

### DIFF
--- a/spec/add.typ
+++ b/spec/add.typ
@@ -28,10 +28,10 @@ This template introduces #nr_interactions interaction(s).
 This template introduces the following constraints
 #render_constraint_table(chip, config)
 
-Note that the correctness of these constraints follows from @lm:limb-decomposition-constraint-correctness, when applied to $(S, L, C, alpha, mu) = (2^64, 2^32, 2, 2, 0)$:
-- the definition of `carry` matches that of @eq:def_ci and @eq:c_-1_is_zero, 
-- @eq:range_ci is enforced by @add:c:carry, and
-- @eq:range_wi follows from @add:a:sum.
+Note that the correctness of these constraints follows from @limbs:lm:limb-decomposition-constraint-correctness, when applied to $(S, L, C, alpha, mu) = (2^64, 2^32, 2, 2, 0)$:
+- the definition of `carry` matches that of @limbs:eq:def_ci and @limbs:eq:c_-1_is_zero, 
+- @limbs:eq:range_ci is enforced by @add:c:carry, and
+- @limbs:eq:range_wi follows from @add:a:sum.
 
 = #sub
 
--- a/spec/add.typ
+++ b/spec/add.typ
@@ -31,3 +31,8 @@ This template introduces #nr_interactions interaction(s).
 = Constraints
 This template introduces the following constraints
 #render_constraint_table(chip, config)
+
+Note that the correctness of these constraints follows from @lm:limb-decomposition-constraint-correctness, when applied to $(S, L, C, alpha, mu) = (2^64, 2^32, 2, 2, 0)$:
+- the definition of `carry` matches that of @eq:def_ci and @eq:c_-1_is_zero, 
+- @eq:range_ci is enforced by @add:c:carry, and
+- @eq:range_wi follows from @add:a:sum.
--- a/spec/limbs_and_carries.typ
+++ b/spec/limbs_and_carries.typ
@@ -2,9 +2,10 @@
 #import "@preview/ctheorems:1.1.3": *
 #import "@preview/equate:0.3.2": equate
 
+
 // Theorem/lemma formatting
 #show: thmrules.with(qed-symbol: $square$)
-#let lemma = thmbox("lemma", "Lemma", fill: rgb("#eee"))
+#let lemma = thmbox("lemma", "Lemma", fill: rgb("#eee"),base_level: 0)
 #let proof = thmproof("proof", "Proof")
 
 // Equation formatting
@@ -26,19 +27,16 @@ Let $S := L^n in NN$ be an upper bound on the integers we want to represent,
 where $L in NN without {0, 1}$ is the number of values a limb can represent 
 and $n in N$ denotes the number of limbs.
 
-Observe that for all $x in [S]$, there exists a unique series of integers 
-$(x_0, x_1, x_2, ...) in [L]^*$ such that
+Observe that for all $x in [S]$, there exists a unique tuple of integers 
+$(x_0, x_1, x_2, x_(n-1)) in [L]^(n)$ such that
 $
   x 
-  = sum_(i=0)^oo x_i dot L^i.
+  = sum_(i=0)^(n-1) x_i dot L^i.
 $ #<eq:decomposition>
+To simplify future notation, we let $x_i = 0$ for all $i >= n$.
 
-Note that $x_i = 0$ for all $i >= n$, since $L^i >= L^n = S > x$.
-As a result, the sum's range can be lowered from $oo$ to $n-1$.
-In practice, we often include $x_i$ with $i >= n$ in equations to simplify notation.
-
-Let us define the multi-linear function $f_(alpha, mu) (x, y, z) := mu dot x dot y + alpha dot z$, 
-over variables $x, y, z in [S]$ and constants $alpha, mu in [L/2-1]$.
+Let us define the family of multi-linear functions $f_(alpha, mu) (x, y, z) := mu dot x dot y + alpha dot z$, 
+over variables $x, y, z in [S]$ and parameters $alpha, mu in NN$.
 Working towards the limb-decomposition of $f_(alpha, mu)$, we rewrite
 $
   f_(alpha, mu) (x, y, z) 
@@ -66,31 +64,33 @@ Note that these $c_i$ effectively move the "overflow" from one limb to the next 
 they're commonly referred to as the _carry_ values.
 
 #lemma[
-    Given $alpha, mu in [L/2-1]$ and $x, y, z in [S]$, $(w_0, w_1, ..., w_(2n-1), 0, 0, ...) in [L]^*$ 
-    for previously defined series $w_i$ is the unique limb decomposition of $f_(alpha, mu)(x, y, z)$
-    if and only if $c_(2n-1) = 0$.
+    For all $g >= 2n$, 
+    $(w_0, w_1, ..., w_(g-1)) in [L]^(g)$ 
+    is the unique $g$-limb decomposition of $f_(alpha, mu)$
+    if and only if $c_(g-1) = 0$.
 ]<lm:wi_decomp_f>
 #proof[
-    Rewriting the definition of $c_i$, we find the equality $w_i = overline(w)_i + c_(i-1) - c_i dot L$.
+    Reordering the definition of $c_i$, we find the equality $w_i = overline(w)_i + c_(i-1) - c_i dot L$.
     Leveraging this, we see that
     $  
-    sum_(r=0)^(2n-1) w_r dot L^r
-    &= sum_(r=0)^(2n-1) (overline(w)_r + c_(r-1) - c_r dot L) dot L^r\
-    &= (sum_(r=0)^(2n-1) overline(w)_r dot L^r) + (sum_(s=0)^(2n-1) c_(s-1) dot L^s) - (sum_(t=0)^(2n-1) c_t dot L^(t+1))\
-    &= (sum_(r=0)^(2n-1) overline(w)_r dot L^r) + (sum_(s=-1)^(2n-2) c_(s) dot L^(s+1)) - (sum_(t=0)^(2n-1) c_t dot L^(t+1))\
-    &= (sum_(r=0)^(2n-1) overline(w)_r dot L^r) + c_(-1) - c_(2n-1) dot L^(2n-1+1)\
-    &= (sum_(r=0)^(2n-1) overline(w)_r dot L^r) - c_(2n-1) dot L^(2n),\
-    &= (sum_(r=0)^(2(n-1)) overline(w)_r dot L^r) - c_(2n-1) dot L^(2n),\
-    &= f_(alpha, mu)(x, y, z),\
+    sum_(r=0)^(g-1) w_r dot L^r
+    &= sum_(r=0)^(g-1) (overline(w)_r + c_(r-1) - c_r dot L) dot L^r\
+    &= (sum_(r=0)^(g-1) overline(w)_r dot L^r) + (sum_(s=0)^(g-1) c_(s-1) dot L^s) - (sum_(t=0)^(g-1) c_t dot L^(t+1))\
+    &= (sum_(r=0)^(g-1) overline(w)_r dot L^r) + (sum_(s=-1)^(g-2) c_(s) dot L^(s+1)) - (sum_(t=0)^(g-1) c_t dot L^(t+1))\
+    &= (sum_(r=0)^(g-1) overline(w)_r dot L^r) + c_(-1) - c_(g-1) dot L^(g-1+1)\
+    &= (sum_(r=0)^(g-1) overline(w)_r dot L^r) - c_(g-1) dot L^(g),\
+    &= (sum_(r=0)^(2(n-1)) overline(w)_r dot L^r) - c_(g-1) dot L^(g),\
+    &= f_(alpha, mu)(x, y, z) + c_(g-1) dot L^(g),\
     $
-    where the second-to-last step follows from the observation that $overline(w)_(2n-1) = 0$.
+    where the second-to-last step follows from the observation that $overline(w)_j = 0$ for $j >= 2n-1$.
+    This limb decomposition is proper if and only if $c_(g-1) = 0$.
 ]
 
 = Upper bounding the carry
-To find the conditions under which $c_(2n-1) = 0$, we prove an upperbound for $c_i$.
+To find the conditions under which $c_(g-1) = 0$, we prove an upperbound for $c_i$.
 
 #lemma("Carry upper bound [part 1]")[
-  Given the definition above, it holds that
+  For $alpha, mu in [L]$, it holds that
   $
   c_i <= mu (i+1) (L-1) + alpha - mu - delta_(mu < alpha)
   $
@@ -109,6 +109,7 @@ To find the conditions under which $c_(2n-1) = 0$, we prove an upperbound for $c
   &overline(w)_0
   &=& alpha dot z_0 + mu dot sum_(j=0)^0 x_(0-j) dot y_j\
   &&<=& alpha (L-1) + mu (L - 1)^2 \
+  &&=& (alpha - delta) L + delta L - alpha + mu (L^2 - 2L) + mu\
   &&=& (alpha - delta) L + mu (L^2 - 2L) + delta L + mu - alpha,\
   text("and hence") &c_0 
   &<=& floor.l ((alpha - delta) L + mu (L - 2)L + delta L + mu - alpha) dot L^(-1) floor.r\
@@ -134,7 +135,7 @@ Since $x_i = y_i = z_i = 0$ for $i >= n$, the upper bound for $overline(w)_i$ is
 We therefore introduce a second lemma:
 
 #lemma("Carry upper bound [part 2]")[
-  Given $c_i$'s earlier definition, it holds that
+  For $alpha in [L], mu in [L/2]$, it holds that
   $
     c_(n+k) <= mu (n - k - 1)(L-2) + mu (n-k) - delta_(alpha < 2mu + delta)
   $
@@ -179,13 +180,15 @@ Again, we note that this upper bound is tight;
 $c_(n+k)$ achieves the bound for all $k in [0, n)$ when $x = y = z = S-1$.
 This includes $k = n-1$, which yields $c_(n+(n-1)) = c_(2n-1) <= mu - delta'$.
 Moreover, $c_(2n) <= floor.l (mu - delta') dot L^(-1) floor.r = 0$ since $mu in [L]$ and thus $c_(2n + i) = 0$ for all $i >= 0$.
-Applying this to @lm:wi_decomp_f, we conclude that $(w_0, w_1, ..., w_(2n-1), 0, 0, ...) in [L]^*$ is the unique limb decomposition
+Applying this to @lm:wi_decomp_f, we conclude that $(w_0, w_1, ..., w_(2n-1)) in [L]^(2n)$ is the unique limb decomposition
 of $f_(alpha, mu) (x, y, z)$ when $mu = 0$ and $alpha < L$, or $mu = 1$ and $alpha <= 2$.
 
-Now, observe that 
+Combining both bounds, we find that
 $
-  &max_(i in [2n]) min(mu (i+1) (L-1) + alpha - mu - delta, mu (2n-i-1)(L-2) + mu (2n-i) - delta')\
-  =& mu n (L-1) + alpha - mu - delta
+&max(&max_(i in [n]) #h(1em) mu (i+1) (L-1) + alpha - mu - delta,\ 
+  &max_(k in [n]) #h(1em) mu (2n-k-1)(L-2) + mu (2n-k) - delta')\
+&= max(& mu n (L-1) + alpha - mu - delta, mu (2n-1)(L-2) + 2 mu n - delta')\
+&= mu n (L-1) + alpha - mu - delta
 $
 acts as an upper bound on all carry elements.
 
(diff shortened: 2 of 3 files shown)


spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,232 @@
...
+Let us define the multi-linear function $f_(alpha, mu) (x, y, z) := mu dot x dot y + alpha dot z$, 
+over variables $x, y, z in [S]$ and constants $alpha, mu in [L/2-1]$.
+Working towards the limb-decomposition of $f_(alpha, mu)$, we rewrite
+$
+  f_(alpha, mu) (x, y, z) 
+  &= mu dot x dot y + alpha dot z\
+  &= mu dot (sum_(i=0)^(n-1) x_i dot L^i) dot (sum_(j=0)^(n-1) y_j dot L^j) + alpha dot sum_(k=0)^(n-1) z_k dot L^k\
+  &= (sum_(i=0)^(n-1) sum_(j=0)^(n-1) mu dot x_i dot y_j dot L^(i + j)) + sum_(k=0)^(n-1) alpha dot z_k dot L^k\
+  &= sum_(i=0)^(n-1) alpha dot z_i dot L^i + sum_(j=0)^(n-1) mu dot x_i dot y_j dot L^(i + j)\
+//   &= sum_(i=0)^(2(n-1)) alpha dot z_i dot L^i + sum_(j=0)^(n-1) mu dot x_i dot y_j dot L^(i + j)\
```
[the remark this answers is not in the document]
erik-3milabs:
```suggestion
  &= sum_(i=0)^(n-1) (alpha dot z_i dot L^i + sum_(j=0)^(n-1) mu dot x_i dot y_j dot L^(i + j))\
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,232 @@
...
+        &<= (mu n (L-1)^2 + alpha (L-1)) + (mu (n-1) (L-1) + alpha - mu - delta)\
+        &<= mu n L^2 + alpha L\
+        &<= C L,\
+    $
+    where the utilized upper bound 
+    $overline(w)_i <= mu n(L-1)^2 + alpha (L-1)$ 
+    can be extracted from the proofs of @lm:carry-upperbound-pt1 and @lm:carry-upperbound-pt2.
+    Moreover, observe that $|X_i inter [L]| <= 1$ since $C < frac(p,L, style:"horizontal")$.
+    Constraint @eq:range_wi therefore enforces that $w_i = overline(w)_i + c_(i-1) mod L$ if $w_(i-1)$ is correct, 
+    and as a result, $c_i = floor.l frac(overline(w)_0, L, style: "horizontal") floor.r$ is correct.
```
erik-3milabs:
```suggestion
    Constraint @eq:range_wi therefore enforces that $w_i = overline(w)_i + c_(i-1) mod L$ if $c_(i-1)$ (and therefore $w_(i-1)$) is correct, 
    and as a result, $c_i = floor.l frac(overline(w)_i + c_(i-1), L, style: "horizontal") floor.r$ is correct.
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,232 @@
...
+$
+  w_i &:= overline(w)_i + c_(i-1) mod L &text("for") i >= 0,\
+  c_i &:= (overline(w)_i + c_(i-1) - w_i ) dot L^(-1) &text("for") i >= 0,\
+$
+with $c_i in NN$ and $c_(-1) := 0$.
+Note that these $c_i$ effectively move the "overflow" from one limb to the next limb up; 
+they're commonly referred to as the _carry_ values.
+
+#lemma[
+    Given $alpha, mu in [L/2-1]$ and $x, y, z in [S]$, $(w_0, w_1, ..., w_(2n-2), 0, 0, ...) in [L]^*$
```
erik-3milabs: From Codex:
```suggestion
    Given $alpha, mu in [L/2-1]$ and $x, y, z in [S]$, $(w_0, w_1, ..., w_(2n-1), 0, 0, ...) in [L]^*$ 
```

erik-3milabs:
The three points from Codex's review have been addressed in the previous commit (7111791)

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,231 @@
...
+Observe that for all $x in [S]$, there exists a unique series of integers 
+$(x_0, x_1, x_2, ...) in [L]^*$ such that
+$
+  x 
+  = sum_(i=0)^oo x_i dot L^i.
+$ #<eq:decomposition>
+
+Note that $x_i = 0$ for all $i >= n$, since $L^i >= L^n = S > x$.
+As a result, the sum's range can be lowered from $oo$ to $n-1$.
+In practice, we often include $x_i$ with $i >= n$ in equations to simplify notation.
```
RobinJadoul: I would suggest writing the sum and first definition of the `x_i` with the narrow range, and if we want to simplify notation, we can suggest here that we let `x_i = 0` for any other value of `i >= n`

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,231 @@
...
+We can introduce helper sequence $c_i$ to transform $overline(w)_i$ into a proper decomposition $w_i$ as:
+$
+  w_i &:= overline(w)_i + c_(i-1) mod L &text("for") i >= 0,\
+  c_i &:= (overline(w)_i + c_(i-1) - w_i ) dot L^(-1) &text("for") i >= 0,\
+$
+with $c_i in NN$ and $c_(-1) := 0$.
+Note that these $c_i$ effectively move the "overflow" from one limb to the next limb up; 
+they're commonly referred to as the _carry_ values.
+
+#lemma[
```
RobinJadoul, line 74: There's something weird with the lemma numbering, it has a leading `0.`

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,231 @@
...
+  where kronecker delta $delta_x$ equals $1$ if $x$ holds, and $0$ otherwise.
+]<lm:carry-upperbound-pt1>
+
+#proof[
+  Since $w_i in [L]$,
+  $c_i 
+  :=(overline(w)_i + c_(i-1) - w_i ) dot L^(-1) 
+  = floor.l (overline(w)_i + c_(i-1)) dot L^(-1) floor.r.$
+  Hence, $c_i$ is maximized when both $overline(w)_i$ and $c_(i-1)$ are, and thus, 
+  by induction, when $overline(w)_j$ is maximized for all $j in [0, i]$.
```
RobinJadoul: Technically the `[0, i]` should be `[0, i)`, I suppose. Maybe just "for all i < j"

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,231 @@
...
+  Given that $x_i, y_i, z_i <= L-1$, it follows that
+  $
+  &overline(w)_0
+  &=& alpha dot z_0 + mu dot sum_(j=0)^0 x_(0-j) dot y_j\
+  &&<=& alpha (L-1) + mu (L - 1)^2 \
+  &&=& (alpha - delta) L + mu (L^2 - 2L) + delta L + mu - alpha,\
```
RobinJadoul:
```suggestion
  &&=& (alpha - delta) L + delta L - alpha + mu (L^2 - 2L) + mu,\
```
makes the transformation a bit more explicit

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,231 @@
...
+  &= mu (n-k-1) (L-2) + mu (n-k)-delta'.
+  $
+  The claimed upper bound now follows for all $k in [0, n)$ by induction. 
+]
+Again, we note that this upper bound is tight; 
+$c_(n+k)$ achieves the bound for all $k in [0, n)$ when $x = y = z = S-1$.
+This includes $k = n-1$, which yields $c_(n+(n-1)) = c_(2n-1) <= mu - delta'$.
+Moreover, $c_(2n) <= floor.l (mu - delta') dot L^(-1) floor.r = 0$ since $mu in [L]$ and thus $c_(2n + i) = 0$ for all $i >= 0$.
+Applying this to @lm:wi_decomp_f, we conclude that $(w_0, w_1, ..., w_(2n-1), 0, 0, ...) in [L]^*$ is the unique limb decomposition
+of $f_(alpha, mu) (x, y, z)$ when $mu = 0$ and $alpha < L$, or $mu = 1$ and $alpha <= 2$.
```
RobinJadoul: There bounds don't match the `μ, α  ∈ [L/2 - 1]` from @lm:wi_decomp_f

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,231 @@
...
+
+Now, observe that 
+$
+  &max_(i in [2n]) min(mu (i+1) (L-1) + alpha - mu - delta, mu (2n-i-1)(L-2) + mu (2n-i) - delta')\
```
RobinJadoul: I don't fully understand where this comes from, in particular the `min`.
I can see the `2n` coming from `n - (i - n)`, but somehow this assumes that the correct bound for one half is the minimum out of the two possible values, which is not a priori obvious.

e.g. for `i = n - 1`, we get `min(μ n (L - 1) + α - μ - δ, μ n (L - 1) + μ - δ')`, which would depend on the relation between μ and α, and for e.g. the case `α = L/2 - 2, μ = 1` this would (wrongly) choose the second as upper bound.
Or in the degenerate case where `μ = 0`, we'd get `min(α - 1, 0) = 0` everywhere as upper bound.

Probably the correct rephrasing is to take `max(max_{i ∈ [n}(...), max_{k ∈ n}(...))` instead

spec/limbs_and_carries.typ
erik-3milabs: Nope: `[0, i]`:  `c_i` is defined in terms of `overline(w)_i` and thus, it is maximized when all `overline(w)_j` are for `j <= i`.

Also, `i<j` would be swapping the terms.

RobinJadoul: Inclusion: yeah, fair
The second part was me swapping the operands accidentally, but I suppose the question then becomes whether `j ≤ i` is nicer to read that `j ∈ [0, i]`

spec/limbs_and_carries.typ
RobinJadoul: Definitely fixed. Open question from me whether we prefer a global namespace for lemmas/theorems of if we want to keep them grouped by chapter (e.g. `lemma 29.1`)

erik-3milabs: I briefly investigated section-grouped indexing, but found that `ctheorems` doesn't know how to play nice with our chapter-setup: the leading 0 is the chapter index (which is never incremented, since our chapters are not numbered). Let's stick with a global namespace for now.

spec/add.typ
```diff
@@ -28,6 +28,11 @@ This template introduces #nr_interactions interaction(s).
 This template introduces the following constraints
 #render_constraint_table(chip, config)
 
+Note that the correctness of these constraints follows from @lm:limb-decomposition-constraint-correctness, when applied to $(S, L, C, alpha, mu) = (2^64, 2^32, 2, 2, 0)$:
+- the definition of `carry` matches that of @eq:def_ci and @eq:c_-1_is_zero, 
+- @eq:range_ci is enforced by @add:c:carry, and
+- @eq:range_wi follows from @add:a:sum.
+
```
erik-3milabs: I'm noticing that the labels of these references are broken; local referencing inside the limb-decomposition section is fine. Are you OK with using the `equate` package despite these broken labels?

RobinJadoul: What exactly are we getting from equate? And does not using it fix the labels?

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
++ the multiplication and addition of limb-decomposed integers (involving carries), 
++ prove an upper bound on the size of the carries in terms of the number of
+  multiplications and additions, and
++ prove correctness of a set of constraints that can be used to constrain quadratic relations.
+
+= Limb decomposition
+Let $[X]$ denote the set ${0, ..., X-1} subset.eq NN$.
+Let $S := L^n in NN$ be an upper bound on the integers we want to represent, 
+where $L in NN without {0, 1}$ is the number of values a limb can represent 
+and $n in N$ denotes the number of limbs.
```
RobinJadoul:
```suggestion
and $n in NN$ denotes the number of limbs.
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+
+Observe that for all $x in [S]$, there exists a unique tuple of integers
```
RobinJadoul:
```suggestion
Observe that for all $x in [S]$, there exists a unique "limb decomposition"
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+and $n in N$ denotes the number of limbs.
+
+Observe that for all $x in [S]$, there exists a unique tuple of integers 
+$(x_0, x_1, x_2, x_(n-1)) in [L]^(n)$ such that
+$
+  x 
+  = sum_(i=0)^(n-1) x_i dot L^i.
+$ #<eq:decomposition>
+To simplify future notation, we define $x_i := 0$ for all $i >= n$.
+Next, we define the family of multi-linear functions
```
RobinJadoul: paragraph break?

RobinJadoul: Do we care that the functions are multi-linear?

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+Observe that for all $x in [S]$, there exists a unique tuple of integers 
+$(x_0, x_1, x_2, x_(n-1)) in [L]^(n)$ such that
+$
+  x 
+  = sum_(i=0)^(n-1) x_i dot L^i.
+$ #<eq:decomposition>
+To simplify future notation, we define $x_i := 0$ for all $i >= n$.
+Next, we define the family of multi-linear functions 
+$
+f_(mu, alpha) (x, y, z) := sum_(m in [mu]) (x^((m)) dot y^((m))) + sum_(a in [alpha]) z^((a))
```
RobinJadoul: Maybe make the variables boldface, as they're vectorial

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+$
+over variables $x, y in [S]^mu$ and $z in [S]^alpha$, with parameters $mu, alpha in NN$.
+Working towards the limb-decomposition of $f_(mu, alpha)$, we rewrite
```
RobinJadoul:
```suggestion
Working towards the limb decomposition of $w  = f_(mu, alpha)(bold(x), bold(y), bold(z))$, we rewrite
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+  &= sum_(r=0)^(2(n-1)) (sum_(j=0)^(r) (sum_(m in [mu]) x^((m))_(r-j) dot y^((m))_j) + sum_(a in [alpha]) z^((a))_r) dot L^r\
+  &= sum_(r=0)^(2(n-1)) overline(w)_r dot L^r, #<eq:f-semi-decomposition>
+$
+where
+$
+  overline(w)_i := sum_(m in [mu]) (sum_(j=0)^(i) x^((m))_(i-j) dot y^((m))_j) + sum_(a in [alpha]) z^((a))_i.
+$
+
+While @eq:f-semi-decomposition closely resembles that of a limb-decomposition (@eq:decomposition), 
+there is the problem that $overline(w)_i$ will exceed $L$ for sufficiently large $alpha, mu, x, y, z$, and $i$.
```
RobinJadoul:
```suggestion
there is the problem that $overline(w)_i$ will generally not be bounded by $L$.
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+We therefore introduce helper sequence $c_i$ to transform $overline(w)_i$ into a proper decomposition $w_i$ as:
```
RobinJadoul:
```suggestion
We therefore introduce a helper sequence, $c_i$, to transform $overline(w)_i$ into a proper decomposition $w_i$ as:
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+  c_i &:= (overline(w)_i + c_(i-1) - w_i )/L &text("for") i >= 0,\
+$
+with $c_i in NN$ and $c_(-1) := 0$.
+Note that these $c_i$ effectively move the "overflow" from one limb to the next limb up; 
+they're commonly referred to as the _carry_ values.
+
+#lemma[
+    For all $g >= 2n-1$, 
+    $(w_0, w_1, ..., w_(g-1)) in [L]^(g)$ 
+    is the unique $g$-limb decomposition of $f_(mu, alpha)$
```
RobinJadoul:
```suggestion
    is the unique $g$-limb decomposition of $w = f_(mu, alpha)((bold(x), bold(y), bold(z))$
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+    &= (sum_(r=0)^(g-1) overline(w)_r dot L^r) - c_(g-1) dot L^(g),\
+    &= (sum_(r=0)^(2(n-1)) overline(w)_r dot L^r) - c_(g-1) dot L^(g),\
+    &= f_(mu, alpha)(x, y, z) - c_(g-1) dot L^(g),\
+    $
+    where the second-to-last step follows from the observation that $overline(w)_j = 0$ for $j > 2(n-1)$.
+    We conclude that $w_i$ is a proper $g$-limb decomposition of $f_(mu,alpha)$ if and only if $c_(g-1) = 0$.
+]
+
+= Upper bounding the carry
+To find some of the conditions under which $c_(g-1) = 0$, we prove two upperbounds for $c_i$.
```
RobinJadoul:
```suggestion
To bound for which $g$ we can guarantee that $c_(g-1) = 0$, we prove two upper bounds for $c_i$.
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
+#import "/book.typ": book-page
+#import "@preview/ctheorems:1.1.3": *
+#import "@preview/equate:0.3.2": equate
+
+
+// Theorem/lemma formatting
+#show: thmrules.with(qed-symbol: $square$)
+#let lemma = thmbox("lemma", "Lemma", fill: rgb("#eee"),base_level: 0)
+#let corollary = thmbox("lemma", "Corrollary", fill: rgb("#eee"), base_level: 0)
```
RobinJadoul:
```suggestion
#let corollary = thmbox("lemma", "Corollary", fill: rgb("#eee"), base_level: 0)
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+  The claimed upper bound now follows for all $k in [0, n)$ by induction. 
+]
+
+Combining both upper bounds, we now find that
+
+#corollary[
+Given $alpha in [L]$ and $mu in [L/2]$, then for all $i <= 2n$:
+$
+c_i <= &max(&max_(i in [n]) #h(1em) mu (i+1) (L-1) + alpha - mu - delta,\ 
+  &max_(k in [n]) #h(1em) mu (2n-k-1)(L-2) + mu (2n-k) - delta')\
```
RobinJadoul: I don't think those 2s should be here

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+&= max(& mu n (L-1) + alpha - mu - delta, mu (2n-1)(L-2) + 2 mu n - delta')\
+&= mu n (L-1) + alpha - mu - delta
```
RobinJadoul: Unless the 2s disappear, this doesn't hold, afaict; when they do disappear, it works (with L ≥ 2), except maybe some weirdness with the deltas I didn't work through entirely

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+$
+]
+
+Note that this upper bound is tight:
```
RobinJadoul: How is this tight, if we're using the upper bound from the first half of the coefficients? Does this refer to the lemma 3 bound instead of the corollary?

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+  &max_(k in [n]) #h(1em) mu (2n-k-1)(L-2) + mu (2n-k) - delta')\
+&= max(& mu n (L-1) + alpha - mu - delta, mu (2n-1)(L-2) + 2 mu n - delta')\
+&= mu n (L-1) + alpha - mu - delta
+$
+]
+
+Note that this upper bound is tight: 
+$c_(n+k)$ achieves the bound for all $k in [n]$ when $x^((m)) = y^((m)) = z^((a)) = S-1$ for all $m in [mu]$ and $a in [alpha]$.
+For $k = n-1$, this yields $c_(n+(n-1)) = c_(2n-1) <= mu - delta'$, which evaluates to zero when $mu = 0$ and $alpha < L$, or $mu = 1$ and $alpha <= 2$.
+We can therefore conclude that $(w_0, w_1, ..., w_(2n-1)) in [L]^(2n)$ is a valid $2n$-limb decomposition of $f_(mu, alpha) (x, y, z)$ in these cases.
```
RobinJadoul:
```suggestion
We can therefore conclude that $(w_0, w_1, ..., w_(2n-1)) in [L]^(2n)$ is a valid $2n$-limb decomposition of $f_(mu, alpha) (bold(x), bold(y), bold(z))$ in these cases.
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+    $
+        overline(w)_i + c_(i-1)
+        &<= (mu n (L-1)^2 + alpha (L-1)) + (mu (n-1) (L-1) + alpha - mu - delta)\
+        &<= mu n L^2 + alpha L\
+        &<= C L,\
+    $
+    where the utilized upper bound 
+    $overline(w)_i <= mu n(L-1)^2 + alpha (L-1)$ 
+    can be extracted from the proofs of @lm:carry-upperbound-pt1 and @lm:carry-upperbound-pt2.
+    Moreover, observe that $|X_i inter [L]| <= 1$ since $C < frac(p,L, style:"horizontal")$.
```
RobinJadoul: Maybe more clear as 
```suggestion
    Moreover, observe that $|X_i inter [L]| <= 1$ since $0 <= C L < p$.
```

spec/limbs_and_carries.typ
```diff
@@ -0,0 +1,240 @@
...
+    observe that $w_i := overline(w)_i + c_(i-1) mod L in X_i$, since
+    $
+        overline(w)_i + c_(i-1)
+        &<= (mu n (L-1)^2 + alpha (L-1)) + (mu (n-1) (L-1) + alpha - mu - delta)\
+        &<= mu n L^2 + alpha L\
+        &<= C L,\
+    $
+    where the utilized upper bound 
+    $overline(w)_i <= mu n(L-1)^2 + alpha (L-1)$ 
+    can be extracted from the proofs of @lm:carry-upperbound-pt1 and @lm:carry-upperbound-pt2.
```
RobinJadoul: Presumably we want to ref the corollary here, otherwise it's not doing much.

RobinJadoul (changes requested):
A couple nits, some typography, and some weirdness left surrounding the corollary. But I think next pass should be there

spec/add.typ
erik-3milabs: > What exactly are we getting from equate?

Two things:
1. sub-equation numbering, and
2. sub-equation referencing

> And does not using it fix the labels?

No, it would mean that we have to reference them more "manually" (e.g., "as in the third sub-equation of @eq:carry_proof")

spec/limbs_and_carries.typ
erik-3milabs: Indeed, the 2s were wrong.
Turns out you need L >= 4. I updated the choice at the top to reflect this.

spec/limbs_and_carries.typ
erik-3milabs: Ah, yes, it refers to Lemma 3. I moved the corollary down to reflect this.

spec/limbs_and_carries.typ
erik-3milabs: I've added the ref. Without, that corollary is still valuable, as we will refer to it in the ECSM doc.

spec/add.typ
RobinJadoul: Fixed now

spec/limbs_and_carries.typ
RobinJadoul: I agree that seems to make it work now; but could we move the L >= 4 assumption into the corollary instead of making that the general case?
It seems like it would be nice to keep the lemmata that don't strictly need it more general if we ever find ourselves wanting to use L = 2 for the binary case somewhere.

RobinJadoul (approved):

PR #670, opened 2026-06-16
spec: Typecheck padding

RobinJadoul:
This manages to catch both the observed issue with the CPUs padding for `half_instruction_length` and an previously unknown issue in the padding for DVRM.

### DIFF
--- a/spec/src/dvrm.toml
+++ b/spec/src/dvrm.toml
@@ -27,7 +27,7 @@ pad = 0
 name = "q"
 type = "DWordHL"
 desc = "The quotient; $#`n` / #`d`$ rounded towards zero."
-pad = 0
+pad = ["arr", 65535, 65535, 65535, 65535]
 
 [[variables.output]]
 name = "r"
--- a/spec/src/dvrm.toml
+++ b/spec/src/dvrm.toml
@@ -27,7 +27,7 @@ pad = 0
 name = "q"
 type = "DWordHL"
 desc = "The quotient; $#`n` / #`d`$ rounded towards zero."
-pad = ["arr", 65535, 65535, 65535, 65535]
+pad = 0
 
 [[variables.output]]
 name = "r"
@@ -41,7 +41,7 @@ pad = 0
 name = "div_by_zero"
 type = "Bit"
 desc = "Whether $#`d`=0$."
-pad = 1
+pad = 0
 
 [[variables.auxiliary]]
 name = "overflow"
--- a/spec/tooling/chip.py
+++ b/spec/tooling/chip.py
@@ -1042,7 +1042,7 @@ class Chip:
 
     def check_assignment(
         self,
-        check_template: dict[str, Callable[[Optional[Type], list[Type], Type], None]],
+        template_checkers: dict[str, "TemplateChecker"],
         values: Optional[dict[str, Type]] = None,
     ):
         env = Environment(self.config, {}, {})
@@ -1067,9 +1067,9 @@ class Chip:
         for c in self.constraints:
             for sig in c.typecheck(env):
                 # Recurse on templates
-                if isinstance(sig, TemplateSignature) and sig.tag in check_template:
+                if isinstance(sig, TemplateSignature) and sig.tag in template_checkers:
                     with reporter.context(repr(c)):
-                        check_template[sig.tag](sig.condition, sig.input, sig.output)
+                        template_checkers[sig.tag](template_checkers, sig.condition, sig.input, sig.output)
 
 
 def build_signature(config: Config, data: dict) -> Signature:
@@ -1110,8 +1110,13 @@ def check_signatures(found: Iterable[Signature], expected: list[Signature]):
         reporter.asserts(any(sig.matches(exp) for exp in expected), f"Unexpected signature: {sig}")
 
 
-def template_checker(check_template: dict[str, Callable[[Optional[Type], list[Type], Type], None]], chip: Chip):
-    def check(cond: Optional[Type], input: list[Type], output: Type):
+# A Function taking a mapping of available templates (for recursive expansion),
+# an optional condition variable, a list of inputs and an output,
+# and checks that it satisfies the template the checker represents.
+type TemplateChecker = Callable[[dict[str, TemplateChecker], Optional[Type], list[Type], Optional[Type]], None]
+
+def build_template_checker(chip: Chip) -> TemplateChecker:
+    def check(template_checkers: dict[str, TemplateChecker], cond: Optional[Type], input: list[Type], output: Optional[Type]) -> None:
         input = input[:]
         values = {}
         for v in chip.variables:
@@ -1119,6 +1124,9 @@ def template_checker(check_template: dict[str, Callable[[Optional[Type], list[Ty
                 case "input":
                     values[v.name] = input.pop(0)
                 case "output":
+                    if output is None:
+                        reporter.error(f"No output available for template output variable {v.name!r}")
+                        return
                     values[v.name] = output
                 case "condition":
                     values[v.name] = cond if cond else Range.const(1)
@@ -1126,7 +1134,7 @@ def template_checker(check_template: dict[str, Callable[[Optional[Type], list[Ty
                     pass
                 case other:
                     reporter.error(f"Cannot check template with variable of category {other!r}")
-        chip.check_assignment(check_template, values)
+        chip.check_assignment(template_checkers, values)
 
     return check
 
@@ -1139,18 +1147,19 @@ if __name__ == "__main__":
 
     reported = False
     chips: list[Chip] = []
-    template_checkers: dict[str, Callable[[Optional[Type], list[Type], Type], None]] = {}
+    template_checkers: dict[str, TemplateChecker] = {}
     for file in sys.argv[3:]:
         if file in sys.argv[1:3]:
             continue
-        chips.append(chip := Chip.from_file(config, file))
-        template_checkers[chip.name] = template_checker(template_checkers, chip)
+        chip = Chip.from_file(config, file)
+        chips.append(chip)
+        template_checkers[chip.name] = build_template_checker(chip)
         reported |= reporter.reported
     if reported:
         sys.exit(1)
 
     if "ADD" in template_checkers:
-        template_checkers["SUB"] = lambda cond, input, output: template_checkers["ADD"](cond, [output, input[1]], input[0])
+        template_checkers["SUB"] = lambda checkers, cond, input, output: checkers["ADD"](checkers, cond, [output, input[1]], input[0])
 
     for chip in chips:
         reporter.update_location(f"Chip {chip.name}")
--- a/spec/tooling/chip.py
+++ b/spec/tooling/chip.py
@@ -4,7 +4,6 @@ import sys
 import tomllib
 from collections.abc import Callable, Iterable
 from dataclasses import dataclass
-from functools import partial
 from pathlib import Path
 from typing import Never, Optional, Self
 
@@ -1030,6 +1029,14 @@ class Chip:
         reporter.update_location("<string>")
         return cls(config, tomllib.loads(s))
 
+    @property
+    def virtual_vars(self) -> Iterable[VirtualVariable]:
+        return (v for v in self.variables if isinstance(v, VirtualVariable))
+
+    @property
+    def concrete_vars(self) -> Iterable[Variable]:
+        return (v for v in self.variables if not isinstance(v, VirtualVariable))
+
     def typecheck(self) -> Iterable[Signature]:
         typemap = {}
         for v in self.variables:
@@ -1040,10 +1047,9 @@ class Chip:
             typemap[v.name] = t
 
         env = Environment(self.config, {}, typemap)
-        for v in self.variables:
-            if isinstance(v, VirtualVariable):
-                with reporter.context(v.name):
-                    v.typecheck(env)
+        for v in self.virtual_vars:
+            with reporter.context(v.name):
+                v.typecheck(env)
         for c in self.constraints:
             with reporter.context(repr(c)):
                 yield from c.typecheck(env)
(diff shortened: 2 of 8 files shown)


spec/src/dvrm.toml
erik-3milabs, line 1: I agree that the padding is invalid.
The proposed solution works.
A cleaner solution might be to have `div_by_zero = 0`.

spec/tooling/chip.py
```diff
@@ -1087,18 +1137,27 @@ def check_signatures(found: Iterable[Signature], expected: list[Signature]):
 
     reported = False
     chips: list[Chip] = []
+    template_checkers: dict[str, Callable[[Optional[Type], list[Type], Type], None]] = {}
     for file in sys.argv[3:]:
         if file in sys.argv[1:3]:
             continue
-        chips.append(Chip.from_file(config, file))
+        chips.append(chip := Chip.from_file(config, file))
```
erik-3milabs: Not a fan of the walrus here.
```suggestion
        chip = Chip.from_file(config, file)
        chips.append(chip)
```

spec/tooling/chip.py
```diff
@@ -1087,18 +1137,27 @@ def check_signatures(found: Iterable[Signature], expected: list[Signature]):
...
+        template_checkers[chip.name] = template_checker(template_checkers, chip)
```
erik-3milabs: this recursion is trippy: to set `template_checkers[some chip]: Callable`, it passes in _all_ of `template_checkers`, which gets bound (?) into a function which is then returned?

This, I believe, requires proper documentation, either directly here, or by wrapping this in a well-named function and documenting that function. I cannot make heads or tails of what is happening here, nor what `template_checkers` even means / is supposed to do.

spec/tooling/chip.py
```diff
@@ -388,6 +402,17 @@ def build_expr(config: Optional["Config"], data: object) -> Expr:
...
 
+def check_padding_fits(var_name: str, config: "Config", type: Type, pad: Expr):
+    def fits(v, t):
+        if isinstance(v, Range):
+            return v.is_const() and constant_fits(v.get_const(), t)
+        else:
+            return isinstance(v, list) and isinstance(t, list) and len(v) == len(t) and all(map(fits, v, t))
+
+    val = pad.typecheck(Environment(config, {}, {}))
+    reporter.asserts(fits(val, type), f"{var_name!r}: Invalid padding {pad!r} for type {type!r}")
```
erik-3milabs: Given that `var_name` is passed in to be solely passed on to `reporter`, consider passing it in through `reporter.context` to give this function a cleaner signature.

spec/tooling/chip.py
erik-3milabs, line 1: Can we do something about the names `typecheck`, `type_match` and `Type`?
To me, `typecheck` suggests a function with boolean output (fail/pass), but instead it returns an object of type `Type` (another stumbling block), which is actually an evaluation of sorts.
I also never know what the difference between `typecheck` and `type_match` are supposed to be.

spec/src/dvrm.toml
RobinJadoul: Encoding 0/1 = 0 then?

spec/tooling/chip.py
RobinJadoul: This is using `type checking` as the act of assigning types to expressions. While technically the correct term for this may actually be `typing`, that feels more confusing to me.
`Type` being a type is kinda inherent to the reification of types that you need in a context like this.

`type_match` is more of a local thing for individual kinds of expression, that may have a different name possible. Where it is used, it's doing essentially some combination of type coercion to make operands compatible and computing the type for the combination of the result. The reason it's separate from the `typecheck` method in those places is because there's a variadic number of operands in those expressions, and we collapse them with a fold. Maybe `typecheck_binary` is a replacement name?

spec/tooling/chip.py
RobinJadoul: `template_checkers` is a collection of functions, indexed by template name/tag, that checks whether a given assignment satisfies the template.

Passing in `template_checkers` could be delayed to the invocations of these checkers as well, we're just tying the knot a bit earlier here.
We need to pass `template_checkers` to an individual checker somewhere to be able to deal with templates inside of a template definition (such as IS_BIT inside of ADD).
The downside of delaying the access to all templates to later is that the typing of `template_checkers` itself needs to become recursive, since now the `Callable` needs an argument of the type `dict[str, Callable[dict[...], Optional[Type], ...]`. Which could in itself work out

spec/src/dvrm.toml
RobinJadoul: Aha, the downside is that it makes the `NEG` workaround no longer work, because then we have `NEG<1, signed>`

erik-3milabs: > Encoding 0/1 = 0 then?

No, still encode 0/0, but just set the `div_by_zero = 0`; it will not trigger any problems since `mu=0`.

spec/tooling/chip.py
```diff
@@ -1030,27 +1034,56 @@ def typecheck(self) -> Iterable[Signature]:
...
         for c in self.constraints:
-            yield from c.typecheck(env)
+            with reporter.context(repr(c)):
+                yield from c.typecheck(env)
+
+    def check_assignment(
+        self,
+        template_checkers: dict[str, "TemplateChecker"],
+        values: Optional[dict[str, Type]] = None,
+    ):
```
erik-3milabs, line 1070: Rather than having `check_assignment` default to checking the padding when no values are provided, I'd instead propose introducing `check_padding` which calls `check_assignment` on the padding.

erik-3milabs: Also: should we additionally verify that `self.variables INSECTION values.keys() = self.variables`? That way we can strain out assignments that propose setting non-existent variables.

spec/tooling/chip.py
```diff
@@ -1030,27 +1034,56 @@ def typecheck(self) -> Iterable[Signature]:
...
+                        return
+                    env.valmap[v.name] = values[v.name]
+        for v in self.variables:
+            if isinstance(v, VirtualVariable):
+                v.populate_env(env)
+
+        for c in self.constraints:
+            for sig in c.typecheck(env):
+                # Recurse on templates
+                if isinstance(sig, TemplateSignature) and sig.tag in template_checkers:
```
erik-3milabs: what if sig.tag not in template_checkers? Wouldn't that indicate a bug?

spec/tooling/chip.py
```diff
@@ -1074,9 +1107,36 @@ def read_signatures(config, filename) -> list[Signature]:
...
+        reporter.asserts(any(sig.matches(exp) for exp in expected), f"Unexpected signature: {sig}")
+
+
+# A Function taking a mapping of available templates (for recursive expansion),
+# an optional condition variable, a list of inputs and an output,
+# and checks that it satisfies the template the checker represents.
+type TemplateChecker = Callable[[dict[str, TemplateChecker], Optional[Type], list[Type], Optional[Type]], None]
+
+def build_template_checker(chip: Chip) -> TemplateChecker:
+    def check(template_checkers: dict[str, TemplateChecker], cond: Optional[Type], input: list[Type], output: Optional[Type]) -> None:
```
erik-3milabs: Consider passing in `sig: TemplateSignature` rather than decomposing it into `cond`, `input` and `output`.

spec/tooling/chip.py
erik-3milabs: Hmmm.

Having reviewed again, I'm not convinced the recursion is the best way to go. Why not introduce a `check_assignment` for templates and have a (global?) `tag_to_chip` map?
Then the last part on `check_assignment` on chips would look something like this:
```suggestion
        for c in self.constraints:
            with reporter.context(repr(c)):
                all_signatures = c.typecheck(env)
                templates = filter(lambda s: isinstance(s, TemplateSignature))
                for sig in templates:
                    tag_to_chip[sig.tag].check_assignment(sig)
```
which I find a lot easier to read.

RobinJadoul: I don't see a big difference, besides now getting a global variable that I don't love having to setup if we ever do end up using this as a "library" for potential other tooling.

spec/tooling/chip.py
RobinJadoul: Not really, we could be checking a chip in isolation and not invoking `chip.py` with all templates present

spec/tooling/chip.py
RobinJadoul: The one other difference is that now check_assignment takes a `Signature`, which I find to not as cleanly represents what an assignment should be. Perhaps there's a way to decompose it into something that maps `sig` into an assignment given only the chip and then chaining a check_assignment — which is currently encoded in the template checker.

RobinJadoul: Aha, and after a quick attempt at implementation, another downside appears: we can't easily have this mapping for `SUB`, which needs to rearrange some things before passing it on to the logic for `ADD`.
So the replacement would end up being a map from tag to (chip, extractor) where extractor needs to give you the assignment for a signature, which is comparable to the current approach, but now implemented as a suggestion.

spec/tooling/chip.py
erik-3milabs: Should we at least have the reporter mention this? Not as an issue, but at least as an heads up.

RobinJadoul: Perhaps, yeah. Added

spec/tooling/chip.py
RobinJadoul: Now simplified after stacking the PR on top of #695

spec/tooling/chip.py
erik-3milabs: `Chip` iterates through both `VirtualVariable`s and non-`VirtualVariable`s a couple times. Consider introducing

```python
@property
def instantiated_variables(self) -> Iterable[Variable]:
    return filter(lambda v: not isinstance(v, VirtualVariable), self.variables)

@property
def virtual_variables(self) -> Iterable[VirtualVariable]:
    return filter(lambda v: isinstance(v, VirtualVariable), self.variables)
```
or something similar to simplify notation & improve readability.

spec/tooling/chip.py
```diff
@@ -1163,33 +1160,24 @@ def sig_to_assignment(sig: Signature, chip: Chip) -> dict[str, Type]:
 
     reported = False
     chips: list[Chip] = []
-    chip_and_assigner_for_tag: dict[str, tuple[Chip, SigAssigner]] = {}
+    chip_for_tag: dict[str, Chip] = {}
```
erik-3milabs: consider absorbing `chips` into `chip_for_tag` (perhaps with a rename)

spec/tooling/chip.py
```diff
@@ -1030,27 +1042,67 @@ def typecheck(self) -> Iterable[Signature]:
...
+    def padding_assignment(self) -> dict[str, Type]:
+        env = Environment(self.config, {}, {})
+        res = {}
+        for v in self.variables:
+            if not isinstance(v, VirtualVariable):
+                t = v.type
+                if isinstance(t, list) and len(t) == 1:
+                    t = t[0]
+                res[v.name] = CastExpr(v.pad, t).typecheck(env)
+        return res
```
erik-3milabs, line 1064: I am noticing quite some overlap between `Chip.typecheck` and `Chip.padding_assignment`, however, I don't see a way to merge them :(

spec/tooling/chip.py
```diff
@@ -1030,27 +1042,67 @@ def typecheck(self) -> Iterable[Signature]:
...
+        for c in self.constraints:
+            with reporter.context(repr(c)):
+                for sig in c.typecheck(env):
+                    # Recurse on templates
+                    if isinstance(sig, TemplateSignature):
+                        if sig.tag not in chip_for_tag:
+                            reporter.warn(f"Template {sig.tag!r} unavailable, skipping check.")
+                            continue
+                        template = chip_for_tag[sig.tag]
+                        template.check_assignment(chip_for_tag, sig_to_assignment(sig, template))
```
erik-3milabs: I'm not a fan of deep nesting (although this one is OK to read). Consider
```suggestion
        for c in self.constraints:
            with reporter.context(repr(c)):
                sigs = c.typecheck(env)
                template_sigs = filter(lambda s: isinstance(s, TemplateSignature), sigs)
                for sig in template_sigs:
                    if sig.tag not in chip_for_tag:
                        reporter.warn(f"Template {sig.tag!r} unavailable, skipping check.")
                        continue
                    template = chip_for_tag[sig.tag]
                    template.check_assignment(chip_for_tag, sig_to_assignment(sig, template))
```
to reduce depth by one.

spec/tooling/chip.py
```diff
@@ -1074,9 +1126,30 @@ def read_signatures(config, filename) -> list[Signature]:
...
-        reporter.asserts(
-            any(sig.matches(exp) for exp in expected), f"Unexpected signature: {sig}"
-        )
+        reporter.asserts(any(sig.matches(exp) for exp in expected), f"Unexpected signature: {sig}")
+
+
+def sig_to_assignment(sig: Signature, chip: Chip) -> dict[str, Type]:
+    input = sig.input[:]
+    output = sig.output
+    cond = sig.condition
```
erik-3milabs, line 1138: I'm curious: what is your take on
```suggestion
    input, output, cond = sig.input[:], sig.output, sig.cond
```
?

spec/tooling/chip.py
```diff
@@ -1,27 +1,42 @@
+import contextlib
 import copy
 import sys
 import tomllib
 from collections.abc import Callable, Iterable
 from dataclasses import dataclass
+from functools import partial
```
erik-3milabs:
```suggestion
```

spec/tooling/chip.py
RobinJadoul: Not worth it, I think. Especially with the `[:]` in there that would get a bit hidden in the line otherwise

spec/tooling/chip.py
RobinJadoul: I don't think the overlap is quite big or meaningful enough to do much with it. There's also the split between values and types in the Environment that makes it  more different.

spec/tooling/chip.py
RobinJadoul: idk if the `filter` entirely is the right way to convey intent here, but reduced it by a level in a different way.

erik-3milabs (approved):

PR #673, opened 2026-06-18
spec: address missing `KECCAK` assumption

erik-3milabs:


erik-3milabs:
This PR introduces a completeness gap. Instead, it was decided to ensure the spec is complete.

PR #674, opened 2026-06-18
spec: fix `keccak`'s `addr` handling

erik-3milabs:


spec/src/keccak.toml
```diff
@@ -8,8 +8,8 @@ pad = 0
 
 [[variables.input]]
 name = "addr"
-type = "DWordBL"
-desc = "memory address storing the first bit of the state"
+type = "DWordHL"
```
RobinJadoul: I think technically we can do a `DWordWL` here?
We're not using `addr` anywhere else other than computing `state_ptr`, so we can get WL range check from the memory and not need the IS_HALF checks.
This does imply a change to the first bullet point in "Notes/potential optimizations" too.

erik-3milabs: I mean, while we're at it, we might as well drop `addr`and just load the first address in `state_ptr[0][0]`? That would still leave the optimization that `state_ptr[0][0]` could be a `DWordWL` (while the others have to be `HL`s), which I think we should forego in favor of a more readable spec. Wdyt?

RobinJadoul: Sure. 
I think the only reason we didn't do that in the first place was to keep the range_ptr ADD iter clean, but we can even keep it as-is, given that ADD<x; x, 0> works just fine

RobinJadoul (approved):

PR #686, opened 2026-06-19
perf/cleanup(stark): tidy LogUp accumulated column + absorbed-1 const…

diegokingston:
…raint

lookup.rs, three byte-identical improvements:

- build_accumulated_column_from_terms: invert N in the BASE field (a base inversion vs a cubic-extension inversion) and multiply base x extension with the base operand on the left. Cleaner; trivial (once per table).
- build_accumulated_column_from_terms: cache each row's column-sum in a single pass and reuse it for the accumulated column, instead of summing every term column twice (drops one full pass over all term columns during trace build).
- LookupAccumulatedConstraint, absorbed.len()==1 arm: conditionally negate the multiplicity in the base field (signed_m + delta*f) instead of an E x E sign multiply that also promotes m base->ext. Removes a per-row extension multiply in the hot constraint-eval / composition-poly path, matching the 2-interaction arm. Identical value: delta*f - m (sender) / delta*f + m (receiver).

stark lib 128/128 (prove+verify; covers absorbed.len()==1 and ==2 arms).


MauroToscano:
Closing this — the `/bench 5` runs come back at -0.1% (well within variance, "no significant change"), so it doesn't give a measurable speedup. The changes are sound and value-identical, but we don't have the review bandwidth to take perf PRs that don't move the benchmark. If the readability parts are worth keeping, feel free to reopen framed as cleanup rather than as an optimization.

PR #688, opened 2026-06-19
refactor(stark): move inlined tests + trace.rs test helpers into src/tests/

diegokingston:
Removes test-only code from production source files in `stark`:

- `trace.rs`: `compute_trace_polys_main` + the Horner reference `get_trace_evaluations` (used only by `prover_tests`, incl. the barycentric-vs-Horner cross-check) → `tests/trace_test_helpers.rs`. `trace.rs` is now pure production.
- Inlined `mod tests` in `grinding.rs`/`bus_debug.rs`/`table.rs` → `tests/` (disk-spill + debug-checks gated as appropriate).

Behavior-preserving: stark lib **128** (133 with `disk-spill`), clippy clean, production builds clean. 5 minimal test-driven `pub(crate)` widenings. (Left `examples/*` inlined tests — those are example fixtures.)


jotabulacios (approved):

MauroToscano (approved):
Verified in an isolated worktree: pure test-only relocation, moved code is verbatim, the test inventory (`cargo test -- --list`) is identical between this branch and main so nothing silently stopped running, and docs/intra-doc links were updated correctly. LGTM.

PR #689, opened 2026-06-19
refactor(prover): move test-only Traces constructors + trim_zero_rows…

diegokingston:
… into src/tests/

Moves three `#[cfg(test)] pub fn` constructors (`from_logs_trimmed`, `from_logs_minimal`, `from_elf_and_logs_minimal`) and `trim_zero_rows` out of production source files (`trace_builder.rs`, `bitwise.rs`) into a new `prover/src/tests/trace_test_helpers.rs` module.

All ~70 call sites are unchanged because the constructors live in an inherent `impl Traces` block in the same crate. No public API changes.


MauroToscano (approved):
Verified pure test-only relocation (identical test inventory vs main, helpers stay #[cfg(test)] with zero production callers). Restored the loud `# WARNING: UNSOUND FOR PRODUCTION` headers on the trimmed-bitwise helpers since they enable a genuinely unsound proving setup if ever reused outside tests. LGTM.

PR #691, opened 2026-06-19
refactor(stark): unify & clean up the commitment layer

diegokingston:
## What

Cleans up and unifies the Merkle-tree / bit-reversed **commitment** layer of the STARK prover, which was duplicated and entangled inside `prover.rs`. The commitment *how* (leaf layout, bit-reverse permutation, Keccak hashing, tree build) now lives in a dedicated module; the prover only orchestrates *when* to commit.

## Commits

1. **`refactor(stark)`: extract Merkle commitment into a `commitment` module** — moves `keccak_leaves_*` + `commit_*` out of `prover.rs` into `crypto/stark/src/commitment.rs`, behind a single grouped core (`keccak_leaves_bit_reversed_grouped` / `commit_bit_reversed`). Keeps both serializations (`as_bytes` / `write_bytes_be`) and `FieldElementPairBackend` as-is.
2. **`perf(stark)`: row-pair the trace commitment** — the trace LDE now uses the same row-pair leaf layout (`ROWS_PER_LEAF = 2`) the composition poly already used: leaf `i` hashes bit-reversed rows `2i`, `2i+1` (a FRI query's value + its symmetric counterpart), so **one Merkle path authenticates both**. ⚠️ This is a **proof-format change** (smaller trace authentication paths), not byte-identical — verifier updated to match (`verify_opening_pair`).
3. **`refactor(stark)`: prover cleanup (A+D)** — `par` helpers at 7 sites, `ROWS_PER_LEAF` const, error propagation, doc/log fixes.
4. **`docs(stark)`: fix commitment leaf-layout docs** after the trace pairing.
5. **`refactor(prover)`: dedup commit pipeline (B)** — extract `commit_plain<C>` + `spill_tree<C>` on `IsStarkProver`, collapsing the near-duplicate main-trace / aux-trace plain-commit code.

## Correctness

- `cargo test -p stark --lib`: **128/128** · with `--features disk-spill`: **133/133**.
- Commits 1, 3, 4, 5 are **byte-identical** to the prior proof output (the test suite is the oracle). Commit 2 is the deliberate proof-format change (validated by the full prove→verify roundtrip suite).
- `clippy` clean on default / `disk-spill` / `instruments`; builds on the combined feature set.
- One profiling-only delta: under `instruments`, the aux commit's timing bucket now includes its (tiny) disk spill — matching what the main path already measured.

## Follow-ups (not in this PR)

- **cuda**: the GPU trace leaf+tree builders (`gpu_lde`) still emit 1-row leaves and must migrate to the row-pair layout before cuda builds pass. The default (non-cuda) build is fully verified here.
- **C** (separate branch): decompose `multi_prove` into named phase helpers; move self-less utilities off the `IsStarkProver` trait.


MauroToscano (changes requested):
This cannot be merged, it breaks CUDA

PR #693, opened 2026-06-19
refactor(prover): centralize u64 limb decomposition into a limbs helper

diegokingston:
## What

Centralizes the repeated `u64 → 16/32-bit little-endian limb` decomposition that was open-coded across the VM trace tables. Adds a small `prover/src/tables/limbs.rs` module and converts ~20 tables to use it.

## The helper (`prover/src/tables/limbs.rs`)

```rust
pub const fn limbs_16(x: u64) -> [u64; 4]   // [x[0..16], x[16..32], x[32..48], x[48..64]]
pub const fn limbs_32(x: u64) -> [u64; 2]   // [lo, hi]
pub const fn limb_16(x: u64, i: u32) -> u64 // (x >> 16*i) & 0xFFFF
pub fn set_limbs_16(data: &mut [FE], col: usize, x: u64) // FE limbs into data[col..col+4]
pub fn set_limbs_32(data: &mut [FE], col: usize, x: u64) // FE limbs into data[col..col+2]
```

Tests live in the dedicated `prover/src/tests/limbs_tests.rs` (no inline tests in the production module).

## Conversions

`set_limbs_16` / `set_limbs_32` replace the `data[base + cols::X_0] = FE::from(v & 0xFFFF); …` blocks; `limbs_16` / `limbs_32` replace the integer `[u32;4]`/`[u64;…]` decomposition arrays. Tables touched: mul, dvrm, decode, cpu, cpu32, commit, load, page, register, memw, memw_aligned, memw_register, lt, branch, eq, store, ec_scalar, ecdas, keccak, keccak_rnd, shift, plus the `trace_builder` bus-value arrays.

Every converted column group was verified consecutive in its `cols` module before conversion.

## Deliberately left as-is

Parametric column indices (`cols_i[..]`, `state_ptr(lane, k)`, `lo_col`/`hi_col` vars), byte (`& 0xFF`) / nibble (`& 0xF`) splits, non-standard groupings (e.g. `DWordWHH`, limbs starting at index 1), sign-fill constants, and the local `half` closure in `trace_builder` (already a local dedup).

## Correctness

Pure refactor — **byte-identical trace output**. `cargo test -p lambda-vm-prover --lib` is unchanged at **322 passed / 104 failed**; the 104 failures are pre-existing and environmental (tests that read `executor/program_artifacts/*.elf`, absent in the sandbox — they fail identically on `main`). `cargo clippy -p lambda-vm-prover` is clean. The 5 new `limbs_tests` cover the helper (LE order, reconstruction, a drift-guard against the open-coded form, and consecutive-column writes).

### DIFF
--- a/prover/src/tables/branch.rs
+++ b/prover/src/tables/branch.rs
@@ -33,6 +33,7 @@ use stark::lookup::{BusInteraction, BusValue, LinearTerm, Multiplicity, Packing}
 use stark::table::TableView;
 use stark::trace::TraceTable;
 
+use super::limbs::set_limbs_32;
 use super::types::{BusId, FE, GoldilocksExtension, GoldilocksField, SHIFT_16, alu_op};
 
 // =========================================================================
@@ -173,18 +174,6 @@ pub fn generate_branch_trace(
     for (row_idx, (op, multiplicity)) in unique_ops.iter().enumerate() {
         let base = row_idx * cols::NUM_COLUMNS;
 
-        // Extract pc as DWordWL: [Word, Word]
-        let pc_0 = (op.pc & 0xFFFF_FFFF) as u32;
-        let pc_1 = (op.pc >> 32) as u32;
-
-        // Extract offset as DWordWL: [Word, Word]
-        let offset_0 = (op.offset & 0xFFFF_FFFF) as u32;
-        let offset_1 = (op.offset >> 32) as u32;
-
-        // Extract register as DWordWL: [Word, Word]
-        let register_0 = (op.register & 0xFFFF_FFFF) as u32;
-        let register_1 = (op.register >> 32) as u32;
-
         // Compute next_pc
         let next_pc_unmasked = op.compute_next_pc_unmasked();
         let next_pc = op.compute_next_pc();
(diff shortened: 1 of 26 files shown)


MauroToscano:
Superseding this PR with two smaller, focused PRs (came out of the review discussion):

- **#697** — centralizes the `u64` limb decomposition into `dword_wl` / `dword_hl` helpers. Same dedup goal as here, with a simpler API: plain functions returning `[FE; N]` that call sites destructure, kept in the prover (not on the generic `Table`) since the limb width is field-size-specific.
- **#698** — makes `Table.data` private and routes the post-build trace writes through `set_main`, which also removes a latent disk-spill hazard (`keccak_rc` was poking the raw buffer, bypassing the mmap-safe accessors).

Both are byte-identical / behavior-preserving (prover lib 416 pass, stark 128 pass; the 5 `ecsm` failures are pre-existing and unrelated) and independent of each other. Suggest closing this in favor of those two.

MauroToscano:
Superseeded by: https://github.com/yetanotherco/lambda_vm/pull/703

PR #695, opened 2026-06-22
spec: Materialize SUB in sub.toml for improved machine-readability

RobinJadoul:


erik-3milabs (approved):
I fixed a small, obscure reference that was easier to solve myself than to explain in words. Aside from that, LGTM 👌

PR #696, opened 2026-06-22
perf(stark): skip fixed 0/1 muls in LogUp fingerprint accumulation

diegokingston:
In the fingerprint hot loop (prover aux-build + constraint-eval + verifier):

- Bus-id term: alpha_powers[0] = alpha^0 = 1, so embed the bus id into the extension field directly instead of multiplying by 1 (drops one F*E mul per interaction per row, hoisted out of the row loop on the aux path).
- Fixed-zero bus elements (the ~235 constant(0) used for bus-width padding) contribute nothing: skip the F*E multiply + accumulate entirely. Variable elements that happen to be zero on a row also benefit.

Value-identical (field addition is exactly associative): stark lib 128/128 (default + parallel), prover bus/logup tests pass, clippy clean. Net effect on prove time is what we want to measure on the 32-core bench.


MauroToscano:
I have been doing a better statistic analysis for this kind of smaller opts, sharing the results, it's a small speedup but it's there. I will make a follow up to update our measurement methods

image

MauroToscano (approved):

PR #725, opened 2026-06-26
feat: no_std prover/verifier + recursion & empty guests

Oppen:
Part 1 of 3 of the recursion stack. **Base: `main`.**

Convert the prover, verifier, executor, stark, ecsm and supporting crates to build under `no_std` + `alloc` so the STARK verifier can run inside a RISC-V guest (required for recursion).

- Gate host-only `std` paths behind a `std` feature (default on for the host, off for guests). Add a `prove` feature so guests pull in only the verifier half of the prover.
- Move VM memory-layout constants (incl. the 64 MiB private-input cap and Keccak tables) into `executor::constants` so the verifier can use them without the full executor.
- Replace std collections with `hashbrown`, `serde_cbor` with `minicbor-serde`, std float math with `libm` on no_std paths.
- Add the `empty` and `recursion` bench guests + `build_recursion_elfs.sh`, and a core smoke test that proves an inner program and runs the verifier inside the recursion guest end-to-end.

### Validation
- `make test` passes (only the 2 CUDA tests fail — no GPU).
- `empty`, `fibonacci`, `recursion` guests build for `riscv64im-lambda-vm-elf`.

> Stacked PR — the repo doesn't support stacked PRs, so PR 2 (`pr/profiling`) targets this branch and PR 3 (`pr/vkey`) targets PR 2.


Oppen:
Actually neither `std`- nor `prove`-gating are necessary, and the changes bring a lot of conflicts. Will rebase the new programs on top of main without it.

PR #729, opened 2026-06-26
Feat/fri early termination

diegokingston:
# FRI early termination — send final-polynomial coefficients

## Motivation

The FRI commit phase folded the deep-composition codeword down to a single
constant (`fri_last_value`), so every fold cost a Merkle root plus one symmetric
opening per query. This PR terminates folding early instead: once the codeword
encodes a polynomial of degree `< 2^k`, the prover sends those `2^k`
coefficients directly (the standard Plonky3 / ethSTARK approach). `k` is a new
option `fri_final_poly_log_degree` (default `7`).

Trade-off: we add `2^k` field elements to the proof once, but drop the last `k`
fold layers — `k` Merkle roots and `k` openings *per query*. For production
query counts this shrinks the proof and cuts commit-phase work, and clamps to a
no-op for traces too small to fold that far.

## Description

- **Option:** new `ProofOptions.fri_final_poly_log_degree: u8` (default `7`).
- **Proof format (breaking):** `StarkProof.fri_last_value` →
  `fri_final_poly_coeffs: Vec<FieldElement<E>>`.
- **New module `fri/terminal.rs`:** pure helpers converting between a terminal
  codeword and its polynomial coefficients (`coeffs_from_terminal_codeword`
  prover-side, `terminal_codeword_from_coeffs` verifier-side).
- **Prover (`fri/mod.rs`, `gpu_lde.rs`):** `commit_phase_from_evaluations` and
  the GPU mirror commit `total_folds - 1` layers, do one final fold to the
  terminal codeword, and append its coefficients to the transcript.
- **Verifier (`verifier.rs`):** a `fri_termination_params` helper centralizes the
  fold schedule (used by the Fiat-Shamir replay and `step_3_verify_fri`); the
  prover derives the identical schedule inline. `step_3_verify_fri` reconstructs the terminal
  codeword and, before the query loop, checks (1) committed-layer count,
  (2) `fri_final_poly_coeffs.len() == 2^expected_k`, (3) every per-query
  decommitment has exactly `num_committed` layers. `verify_query_and_sym_openings`
  compares each folded query against the reconstructed terminal codeword
  (no-fold / single-fold / general cases).
- **Fiat-Shamir:** `fri_final_poly_log_degree` is absorbed into the statement and
  the coefficients replace the final value in the transcript; tags bumped
  `STATEMENT_V2→V3`, `CONTINUATION_EPOCH_V1→V2`.
- **Robustness:** overflow-safe clamp so an out-of-range `k` can't
  divide-by-zero the prover (degrades to no early termination).
- **Cleanup:** removed the dead `number_layers` param and unused
  `Domain.root_order` field.

## Soundness

- The verifier reconstructs the terminal codeword by FFT from exactly
  `2^expected_k` coefficients, so it is low-degree by construction and the
  blowup (rate) is preserved — per-query soundness is unchanged from folding to
  a constant.
- The per-query decommitment vectors are untrusted and not bound in the
  transcript; check (3) pins their length before the fold loop, blocking empty
  (vacuous accept) or padded (skip terminal check) decommitments.

## Bench results

| Metric                 | main      | PR        | Δ       |
|------------------------|----------:|----------:|--------:|
| Verify time (per-side) | 4.977s    | 4.531s    | -8.96%  |
| Proof size             | 256.86 MiB | 228.31 MiB | -11.12% |


jotabulacios (approved):

nicole-graus (approved):

MauroToscano (changes requested):
There are multiple issues wit this

MauroToscano (approved):

PR #753, opened 2026-06-30
Feat/ecsm update

jotabulacios:
## Motivation

The spec received several updates after PR #657 merged. This PR brings the implementation in sync.

## Description

**EC_SCALAR inlined into ECSM**

The separate EC_SCALAR chip is removed. The scalar `k` is now stored as 256 individual bit columns inside ECSM instead of 32 bytes. ECSM directly holds and validates all scalar bits via IS_BIT constraints, and interacts with the BIT bus as a receiver (one per bit). The SERVE_K self-referential bus is eliminated.

**xG range check in the ECSM AIR**

Adds a new column `xG_sub_p` (U256HL) and proves `xG < p` via the same carry-overflow technique already used for `k < N` and `xR < p`. Previously this was only enforced in the executor.

**Disjoint memory timestamps**

Separates the xG and k memory reads into different timestamp slots: xG reads at T, k register read and k dword reads at T+1, xR write at T+2.

**q1[32] IS_BYTE**

The yG quotient's high byte (`q1[32]`) was previously constrained by IS_BIT (polynomial). It is now IS_BYTE (bus lookup), fixing a completeness issue for witnesses where `q1[32] > 1`.

**ECDAS padding simplified**

Padding rows now use `q0 = q1 = q2 = 0` and `op = 1`. The `R·P` term in the λ and xR convolution relations is gated with `μ`, so it vanishes on padding rows and the relations hold at zero carries without needing to pad quotients to `r = 3p`.

**Curve identifier in ECDAS bus**

A curve identifier `id` (Bit, 0 = secp256k1) is prepended to the ECDAS bus tuple in all interactions, enabling future domain separation between curves.


jotabulacios:
### Responses to AI reviewers

#### Council of AIs

- **AI-001** (IS_BIT on q1[32]): Fixed in `53e31bd` — this was a genuine soundness bug
  introduced when removing a reindexing scheme that was later reverted. The IS_BIT
  constraint is back.

- **AI-002** (LEN_K unconstrained): False positive. `len_k` is constrained implicitly by
  the bus balance: the ECSM chip sends exactly 256 BIT interactions for k, so a wrong
  `len_k` would break bus balance and fail verification.

- **AI-003** (256 IS_BIT receivers redundant): False positive. Each of the 256 bit columns
  holds a single bit of k. Without IS_BIT on each one, a malicious prover could set any
  column to an arbitrary field element while still satisfying the bus interaction (which
  only checks the linear combination). The individual IS_BIT constraints are required for
  soundness.

- **AI-004** (XgLtP missing from test): Fixed in `cd41217` — added `OverflowKind::XgLtP`
  to the overflow carry constraint loop in `constraints_hold_on_generated_trace`.

#### Codex

- **Low** (stale doc comments in `trace_builder.rs` and `crypto/ecsm/src/lib.rs`): Fixed
  in `53e31bd` — `collect_ecsm_ops` doc now accurately describes the disjoint timestamp
  scheme (x11+xG at T; x12+k+x10 at T+1; xR at T+2), and `lib.rs` no longer references
  EC_SCALAR.

- **Open question** (executor overlap rationale): Fixed in `53e31bd` — the comment now
  explains the actual reason: overlapping addresses would cause the same memory byte to
  serve as both an xG limb and a k bit, corrupting the scalar multiplication. The overlap
  guard is still intentional and correct.

#### Claude

- **Minor** (stale doc comments): Fixed in `53e31bd` — same as Codex finding above.

nicole-graus (approved):

diegokingston (approved):

PR #756, opened 2026-06-30
ci/verifier bench

jotabulacios:
This PR adds a verifier wall-time benchmark as a separate `/bench-verify` command (`bench-verify.yml` + `scripts/bench_verify.sh`), kept out of `/bench` like `/bench-abba` and `/bench-gpu`.

It uses the same drift-free A/B/B/A paired method as `/bench-abba`: builds the `cli` at the PR head and `main`, proves once, and has both sides verify that same shared proof over N interleaved pairs, then reports whether the PR is faster or slower with a confidence interval and a clear verdict. Verifier changes are usually small (recent ones ~1.4% to ~17.8%), so the pairing is what lets us resolve them: it cancels machine drift by hitting both sides equally. A verify is ~4s and deterministic, so the default 20 pairs runs in ~4 min and resolves ~0.6%.

Usage: comment `/bench-verify` (20 pairs) or `/bench-verify 32` (clamped `[2,40]`). Like `/bench-abba`, the trigger only works once merged to `main`.


scripts/bench_verify.sh
Oppen, line 1: About 1/3 of this script is an inline Python script. Why not just make the whole thing one pure Python script?

Oppen: Alternatively, given that script is most likely the same logic as all other ABBA tests, maybe extract it to a .py and invoke it here and in the other benches.

jotabulacios: good idea, maybe leave it as a follow up pr?

Oppen: Sure.

Oppen (approved):

PR #758, opened 2026-07-01
Fix(Continuations): Make private input private

nicole-graus:
## Motivation
The continuation prover shipped the raw private input inside the proof bundle, and the verifier recomputed the starting memory from it, so private input was exposed to the verifier.

## Description
- Private-input `GLOBAL_MEMORY` pages are now built **non-preprocessed**: their genesis is a committed, bus-enforced column the verifier never recomputes (mirroring the monolithic PAGE table).
- The bundle drops the raw `private_inputs` bytes for a `num_private_input_pages` count, classified count-based from `PRIVATE_INPUT_START_INDEX` (like the monolithic verifier), bound-checked and absorbed into the global proof's Fiat-Shamir statement.
- `verify_continuation` now works from the bundle + ELF alone, it never sees the raw private input.
- `Elf::load` now rejects any loadable segment overlapping the reserved private-input region, so ELF data can never sit on a page whose genesis isn't recomputed from the ELF (closes a pre-existing forgeable-genesis gap shared with the monolithic path).


MauroToscano (changes requested):
Check the following issues:

  1. Plaintext private-input leak through serialized boundaries
      - Severity: High/Critical
      - EpochProof.boundary serializes CellBoundary.init.value, which contains genesis values for touched cells. For private-input reads, those values are the private
        bytes.

      - This defeats the PR’s headline privacy goal: the bundle no longer carries private_inputs, but read private bytes still appear in plaintext.
      - Fix: serialize only verifier-needed boundary data, likely touched addresses/page bases. The verifier uses boundary via b.address; values are redundant.

  2. Unconstrained OFFSET on non-preprocessed private pages
      - Severity: Critical
      - Private GLOBAL_MEMORY pages make OFFSET a prover-controlled main-trace column, while OFFSET is used to derive bus addresses.
      - This lets a malicious prover forge genesis tokens for same-high-32-bit addresses by choosing invalid offsets.
      - Caveats: this appears shared with monolithic private PAGE handling, so it is pre-existing; it is structurally verified but not PoC’d.
      - Fix: keep OFFSET preprocessed for private pages, or constrain it to the row index/range.

  3. Loader reservation misses the honest max private-input spill page
      - Severity: High
      - The loader reserves [START, START + MAX_PRIVATE_INPUT_SIZE), but the wire image is 4 + len.
      - At max legal input length, the 4-byte prefix spills into page index 256 (0x103000000), which can be classified private but is not reserved from ELF segments.
      - Fix: reserve the full page-rounded wire span: ceil((MAX_PRIVATE_INPUT_SIZE + 4) / PAGE_SIZE) pages.

  4. Verifier bound allows one extra private page
      - Severity: High
      - The verifier accepts ceil((MAX_PRIVATE_INPUT_SIZE + 4) / PAGE_SIZE) + 1, allowing page index 257 (0x103040000) even though honest max is 257 total pages,
        indexed 0..256.

      - This is complementary to #3, not redundant: #3 is the honest max spill page; #4 is the extra slack page.
      - Fix: remove the + 1; check both continuation and monolithic verifier bounds.

  5. Overflowing PT_LOAD ranges are accepted
      - Severity: Medium
      - Elf::load uses saturating arithmetic and never rejects p_vaddr + p_memsz overflow.
      - Downstream code uses unchecked base_addr + ..., causing malformed-input acceptance and debug/release divergence.
      - Fix: use checked_add and return the existing ElfError::AddrTooLarge.

  6. Missing regression tests
      - Severity: Medium
      - Missing serialized-bundle byte-absence test; this would fail today because of #1.
      - Missing continuation test for multi-page private input.
      - Missing in-bounds inflated-count tamper test. This guards bundle tampering via Fiat-Shamir divergence, not the self-consistent forgery path in #2.
      - Fix: add all three, with comments that correctly describe what each guards.

  7. Residual privacy claim overstates non-ZK STARK privacy
      - Severity: Low/Medium
      - Even after fixing #1, non-preprocessed private columns are still opened as deterministic STARK trace evaluations unless the proof system is zero-knowledge/
        blinded.

      - Fix: phrase the claim as “raw private input is not bundled/recomputed by the verifier,” unless actual blinding is added.

  8. Stale/misleading docs and comments
      - Severity: Low
      - Several comments now encode false reasoning, including “verifier never sees” private genesis and boundary-value tampering being inconsequential.
      - These claims mask #1 and #2 and should be corrected with the code changes.
      - Fix: update continuation.rs comments and docs/continuations_design.md to match the actual security model.

nicole-graus:
Issues checked:

1. Fixed in f7777b12 (doc follow-up 7f108b30).
2. Not  a soundness bug: OFFSET/INIT/FINI/FINI_EPOCH are unconstrained in the AIR on private pages, but they're pinned by the GlobalMemory LogUp bus: a forged OFFSET either points a genesis token at an address another page/L2G already claims → unmatched token → bus can't balance → rejected, or changes a cell that's written before it's read → overwritten before use → never affects any output.
3. Fixed in 3cab6b6f
4. Fixed in c10ba950
5. Fixed in 852dd0f1
6. Fixed in bab015ba
7. Fixed in dbac7707
8. Fixed in dbac7707

diegokingston (approved):

jotabulacios (approved):

MauroToscano (approved):

PR #761, opened 2026-07-01
flamegraph: fix tail-call misattribution, trie-based fold, addr2line enrichment

Oppen:
## Summary

- Fix a correctness bug where any `dst=0` jump (loop back-edges, if/else,
  jump tables, self-tail-recursion) was unconditionally treated as a tail
  call, corrupting the tracked call stack. Now compares the containing
  function of `current_pc` vs `next_pc` via `SymbolTable::lookup`; only
  mutates the stack on a genuine cross-function tail call.
- Replace the `String`-keyed `HashMap<String, u64>` fold with a call-graph
  trie (`TrieNode` arena, O(1) push/pop/count independent of stack depth).
  Symbol resolution/demangling is deferred to `write_folded`, memoized per
  address, instead of running per instruction — this is what makes
  flamegraph generation viable at all on real workloads (see Benchmark).
- Add a raw hex-address-keyed output mode (`--flamegraph-raw`) plus a
  standalone script (`scripts/enrich_flamegraph.py`) that drives the system
  `addr2line` binary (one process, one pipe) to expand inlined call chains
  and resolve file:line, with no new crate dependency — the executor never
  shells out itself.
- Extract the CLI's execute+flamegraph drive loop into
  `executor::flamegraph` (`run_with_flamegraph`/`drive_with_flamegraph`),
  shared between the CLI and future callers, with `--cycle-budget` and
  `--flamegraph-checkpoint-cycles`.
- Enable `debug = true` for the `recursion-bench` guest so `addr2line` has
  DWARF to resolve against.

## Benchmark

`cli execute recursion.elf --private-input <blob>`, 73-FRI-query
verification of an `empty` inner proof. `/usr/bin/time -l`, same machine,
same guest ELF and input across each pair of rows, only the CLI rebuilt.
"debug-recursion" = guest ELF built with `debug = true` (this PR); "no-debug"
= guest ELF built without it (`main`'s own build config).

| | Execution only | Flamegraph | Overhead |
|---|---|---|---|
| **This branch**, debug-recursion | 192.41s / 3.495 GB | 222.40s / 3.517 GB | +15.6% time / +0.63% mem |
| **This branch**, no-debug | 195.75s / 3.436 GB | 222.78s / 3.437 GB | +13.8% time / +0.05% mem |
| `main`, debug-recursion | 194.64s / 3.496 GB | **>10,493s** (killed, incomplete) | **>52.9x** time |
| `main`, no-debug | 192.99s / 3.436 GB | **>7,139s** (killed, incomplete) | **>36.0x** time |

`main`'s fold calls `format_stack()` (demangles every frame on the call
stack) on *every instruction*, uncached. Both `main` runs were killed
without finishing, each after multiple hours, while this branch completes
the identical run in ~222s — the numbers above are lower bounds, not final
ones. This branch's flamegraph mode is **at least 32-47x faster** than
`main`'s for the identical guest binary and input.

Peak RSS for the two killed `main` runs (0.95 GB / 0.44 GB) is *not*
comparable to the completed rows: RSS was still climbing linearly at kill
time (each run had processed far fewer guest instructions in its ~2-3
hours than the ~200s execution-only baseline processes in full), so it
understates `main`'s true peak rather than showing it using less memory.

Flamegraph examples (branch, debug-recursion.elf):

## Test plan

- [x] `make test-executor`: 166 asm + 20 flamegraph + 29 rust tests, green
- [x] Regression tests: intra-function `jal`/`jalr x0`, cross-function
  mutual tail calls, self-tail-recursion, `dst=1` recursion unaffected,
  zero-size-symbol boundary pinned, raw-mode keys by address not name
- [x] `cargo clippy`/`cargo fmt` clean on `executor` and `cli`
- [x] `--flamegraph-raw` → `scripts/enrich_flamegraph.py` verified
  end-to-end against a real cross-compiled RISC-V guest ELF with
  `llvm-addr2line`
- [x] Benchmarked against `main` on the same machine, same input


Oppen:
Regular flamegraph:
flamegraph_regular

Enriched flamegraph (raw output + addr2line script):
flamegraph_raw_enriched

diegokingston (approved):

MauroToscano:
Correctness

1. Checkpoint writes are non-atomic — a kill during a checkpoint destroys the previous good checkpoint
bin/cli/src/main.rs:313 — write_flamegraph_checkpoint opens the target with File::create, truncating it in place before rewriting. A SIGKILL/OOM-kill landing mid-write is the exact scenario --flamegraph-checkpoint-cycles advertises surviving ("a killed run still leaves usable output on disk"), yet it leaves the file empty or torn, with the last good checkpoint gone.
Fix: write to a temp file in the same directory, then rename over the target.

2. The enrichment script crashes on the torn file that finding 1 produces
scripts/enrich_flamegraph.py:40 — read_raw_stacks does stack, count = line.rsplit(" ", 1) then int(count) with no malformed-line handling. A last line truncated mid-write raises an unhandled ValueError; if the tear lands inside the trailing integer, it instead silently parses a wrong count. The advertised killed-run → enrich workflow dies exactly when it's needed.
Fix: skip-and-warn on malformed lines.

3. All unsymbolized addresses collapse into one ?? (??:0) frame in enriched output
scripts/enrich_flamegraph.py:106 — addr2line emits ??/??:0 for unresolvable addresses, so the parsed frame list is non-empty and the else: frame_names.append(addr) raw-hex fallback is dead code. Distinct unsymbolized call paths merge with summed counts — losing the address information --flamegraph-raw exists to preserve (the Rust write_folded path keeps 0x{addr:x} for unknowns).
Fix: treat a ?? function as unresolved and fall back to the hex address.

4. Flamegraph write can report success while the file is empty
bin/cli/src/main.rs:322 — the BufWriter is never flushed explicitly; flush happens on drop, where errors are swallowed, and outputs under the 8 KiB buffer never touch disk before drop. On ENOSPC or a failing mount the CLI prints "Flamegraph written to …" and exits 0 with a truncated/empty file. Preexisting on main, but the PR carries it into the new helper and multiplies exposure through every checkpoint.
Fix: writer.flush()? (or into_inner()) at the end of write_flamegraph_checkpoint.

5. A guest fault discards the entire accumulated profile, and the new API can't recover it
bin/cli/src/main.rs:390 — on Err from run_with_flamegraph the FlamegraphGenerator is dropped inside the callee, so a multi-hour run that faults near the end (without checkpointing enabled) loses everything. Behavior-parity with main, not a regression — but the old code at least held the generator in CLI scope, and this PR freezes the loss into the API signature while it's still new.
Fix: return the generator alongside the error (or write a final checkpoint on the error path).

Cleanup

6. The budget-capped drive loop now exists in two places, character-for-character
bin/cli/src/main.rs:414 — the non-flamegraph branch of cmd_execute t computation, resume_with_limit call, and both break conditionsthat drive_with_flamegraph (executor/src/flamegraph.rs:340) centralizes; CHUNK_SIZE was made pub solely to feed the duplicate (same crate — pub(crate) would do).
The stop-exactly-at-budget invariant can now drift between the two
Fix: a shared drive_to_budget helper on the executor, or an Option<&mut FlamegraphGenerator> parameter.

7. write_folded_raw/collect_raw/raw_stacks duplicate write_folded's entire walk-fold-sort-write pipeline
executor/src/flamegraph.rs:297 — the two paths differ only in the psolution vs 0x{addr:x}). Verified safe to merge: routing the rawpath through the same String-keyed counts map is a semantic no-op since raw trie paths are unique by construction. One fold(resolve: impl FnMut(u64) -> String) collapses both writers to thin wrappers and deletes raw_stacks/collr). Matters doubly because the Python parser depends on bothwriters' output format staying in sync.

jotabulacios (approved):

PR #763, opened 2026-07-02
feat(stark): field-generic constraint IR + single-body ConstraintBuilder framework

MauroToscano:
## What

The framework half of the single-source constraints redesign: every transition constraint gets **one definition** that serves three consumers — compiled prover evaluation, compiled verifier evaluation (identical code path for the recursion guest), and a captured flat IR for the upcoming GPU constraint interpreter.

This PR is **inert**: nothing here is wired into production paths, and prover/verifier behavior is byte-identical to main. It adds:

- `crypto/stark/src/constraint_ir/` — a field-generic constraint IR (`ConstraintProgram<F, E>`: flat op-DAG with side-table constants), its builder (hash-consing CSE), and a CPU interpreter. This is the artifact the GPU kernel will consume (data residency: evaluate constraints on-device instead of shipping the LDE back to the CPU).
- `crypto/stark/src/constraints/builder.rs` — the `ConstraintBuilder` trait (single-body authoring surface: leaves + operator-overloaded expressions + indexed emit), `ConstraintSet`, `ConstraintMeta`, and its three implementations: `ProverEvalFolder` (compiled per-row eval), `VerifierEvalFolder` (compiled OOD eval), `CaptureBuilder` (expression tree → flattened IR, with per-constraint measured degrees).
- `crypto/stark/src/constraints/zerofier.rs` — the existing zerofier evaluation logic as free functions of `ConstraintMeta` (bodies relocated verbatim; equivalence-tested against the still-live trait defaults).
- Design docs: `thoughts/gpu-constraint-eval/impl-plan-single-source-constraints.md` (the plan this implements, including the decision record) and `survey-constraint-frontends.md` (how Plonky3 / OpenVM / SP1 / risc0 / zisk / airbender solve the same problem — the design follows the production-validated pattern).

## Why

Constraint definitions were about to be duplicated (a hand-written `evaluate` plus a hand-written GPU capture per constraint). One body with multiple interpretations removes that permanently. The CPU paths stay **compiled** (measured: an interpreted CPU path costs ~9% total prove time on ethrex — recorded in the plan), and the interpreter/IR exists for the GPU and as a differential-testing reference.

## Testing

- 31 new tests, 168/168 total in `cargo test -p stark`: hand-built IR programs vs direct field arithmetic (every op/leaf kind), three-way differentials on random rows (prover folder vs verifier folder vs captured-and-interpreted program), degree measurement, emission-completeness and meta-invariant assertions, zerofier free functions vs the untouched trait defaults across all six branch shapes, and a non-Goldilocks field tower to keep the genericity honest.
- `cargo fmt` / `cargo clippy` clean; no `unsafe` anywhere in the new code.

## Next

The follow-up PR converts all tables + the LogUp framework constraints to single bodies, rewires `AirWithBuses`/`AIR`/zerofier grouping onto this framework, and deletes the old per-constraint machinery — gated on byte-identical golden proofs for a fixed ELF set.

### DIFF
--- a/crypto/stark/Cargo.toml
+++ b/crypto/stark/Cargo.toml
@@ -37,6 +37,7 @@ web-sys = { version = "0.3.64", features = ['console'], optional = true }
 serde_cbor = { version = "0.11.1" }
 
 [dev-dependencies]
+math = { path = "../math", features = ["test-utils"] }
 criterion = { version = "0.4", default-features = false }
 env_logger = "*"
 test-log = { version = "0.2.11", features = ["log"] }
--- a/crypto/stark/src/constraint_ir/builder.rs
+++ b/crypto/stark/src/constraint_ir/builder.rs
@@ -1,0 +1,282 @@
+//! Explicit-builder capture front-end.
+//!
+//! Every transition constraint is captured into a flat [`ConstraintProgram`]
+//! through an explicit [`IrBuilder`]: each constraint translates its algebra
+//! into builder calls (`main`, `add`, `sub`, `mul`, ...). No fake field, no
+//! thread-local arena.
+//!
+//! The builder hash-conses every node on `(Op, Dim)` and only emits leaves for
+//! columns the constraint actually reads, so captured programs are minimal.
+//! Field constants live in the [`ConstraintProgram`]'s `base_consts` /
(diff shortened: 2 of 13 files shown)


MauroToscano:
We will just merge the version with everything: https://github.com/yetanotherco/lambda_vm/pull/764

PR #766, opened 2026-07-02
perf(continuations): Verify-and-drop continuation epoch proofs instead of collecting them

nicole-graus:
## Motivation
`prove_and_verify_continuation` proved every epoch and kept all proofs in a `Vec` before verifying, so retained proof memory grew with the number of epochs.

## Description
Verify each epoch inline right after proving it and drop its `MultiProof`, keeping only the small `boundary` / `l2g_root` / `public_output` needed for the final global proof. Retained per-epoch proof memory is now O(1) instead of O(#epochs). The epoch loop is extracted into a shared `EpochDriver` so `prove_continuation` (the serialized bundle producer) is unchanged, and the streaming verifier derives `is_final`/`label` positionally to stay a faithful mirror of `verify_continuation`.


MauroToscano (changes requested):
This is of no use beyond testing, converting prove and verify to a function that drop proofs is not useful for the normal flow. If it's needed for testing, we should not change or documentation on the design because of this, it's a testing detail

PR #767, opened 2026-07-02
perf(stark): borrow trace rows in place for prover transition eval

MauroToscano:
Stacked on #764. Candidate from the perf investigation — bench in isolation before deciding.

## What

The LDE buffers have been **row-major** since the row-major LDE rework, but the evaluator still gather-copied every main + aux column of every transition offset into an owned `Frame` on **each LDE point** (~150–200 element clones per row per table) before the constraint body ran.

This replaces the Prover context's `frame` with `RowFrame`: one borrowed `(main, aux)` row-slice pair per transition offset, taken straight from the row-major storage with the same cyclic row arithmetic. The `ProverEvalFolder` and the IR interpreter read `rows[offset][col]` directly. The per-thread preallocated `Frame` + `fill_from_lde` are deleted (single-row steps are the only shape since virtual columns were removed — asserted). `Frame` stays for the verifier/OOD path and debug validation (bridged via `Frame::as_row_frame`).

**Reads the same values from the same memory — proofs are unchanged.** Gates: stark 165/165 (incl. folder-vs-capture differentials), emit differentials 13/13, release e2e prove→verify, `make lint`.

## How to bench

Isolated against the base branch (recommended — measures only this change):
```
scripts/bench_abba.sh origin/perf/row-frame-view origin/feat/single-source-constraints-switch 20
```
Or `/bench-abba` on this PR (measures vs main, i.e. this change + the whole switch branch).

### DIFF
--- a/crypto/stark/src/constraint_ir/interp.rs
+++ b/crypto/stark/src/constraint_ir/interp.rs
@@ -175,7 +175,7 @@ pub fn eval_program<F, E>(
     E: IsField,
 {
     let TransitionEvaluationContext::Prover {
-        frame,
+        rows,
         rap_challenges,
         logup_alpha_powers,
         logup_table_offset,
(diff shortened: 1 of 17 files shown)


MauroToscano:
Folded into #764 as commit 3638d7a1 (cherry-picked onto the cleaned-up tip — the micro-op bundle this branch originally stacked on has been reverted there; this change is the one that measured a real win, ~0.8% on ABBA). Branch kept for provenance.

PR #769, opened 2026-07-03
perf(verifier): verify STARK proofs in place via rkyv

Oppen:
Verifies STARK proofs **in place** via rkyv, without dropping serde or duplicating the verifier.

## Design

- **One verification implementation, two proof sources.** `crypto/stark/src/proof/view.rs` adds a borrowed `StarkProofView` (and nested `Table`/`FriDecommitment`/`DeepPolynomialOpening`/`PolynomialOpenings` views), each constructible `Owned(&T)` or `Archived(&ArchivedT)`. Scalars are copied out (`to_native()` vs. a plain copy); field-element/commitment arrays stay borrowed either way (`slice_as_native` on the archived side). The verifier (`crypto/stark/src/verifier.rs`) is written once against this view. `multi_verify` (owned) and `multi_verify_archived` (archived) are thin constructors that build the matching view and share every check — no serialize-then-delegate step, no duplicated logic.
- **serde is kept.** `StarkProof`, `MultiProof`, `FriDecommitment`, `BusPublicInputs`, `Table`, `PolynomialOpenings`, and `DeepPolynomialOpening` dual-derive `serde` alongside `rkyv`, so existing serde-based call sites (examples CLI, disk-spill table serialization) are unaffected. Only the STARK proof wire format used by the prover/CLI/recursion guest moved from bincode to rkyv.
- **Recursion guest**: unchanged DECODE/page-commitment + `program_id` design (this branch's approach, not a vkey digest). The private-input blob gets a 12-byte `LVMR` magic/version prefix so the rkyv archive lands 16-aligned in guest memory (the mapped payload itself is only 4-aligned); this prefix is local to the recursion blob and does **not** change the global private-input ABI (`PRIVATE_INPUT_START + 4` is unchanged for every other guest). The guest borrows the blob in place (`get_private_input_slice`) and verifies straight from the archive — no deserialization pass.
- **ethrex quarantine**: ethrex pins rkyv's `unaligned` feature, which cannot coexist with this workspace's `aligned` feature in one resolved dependency graph. Its host-reference tests move to a detached workspace, `tooling/ethrex-tests` (own `cargo test`), isolating Cargo.lock resolution.
- **CLI proof files** switch bincode → rkyv.

## Numbers

Recursion verifier guest, inner `empty` proof, multi-query (blowup=8, 73 queries) preset (`make test-profile-recursion-multi`), measured against current `main`:

| | main | this branch (rkyv) | Δ |
|---|---|---|---|
| multi-query, blowup=8 | 2,071,186,046 cycles; 67.4s | **1,334,810,646 cycles, 28.4s** | **−35.5%,-57.9%** |

> [!NOTE] Numbers below are out of date

### Regression check: regular (non-recursion) verify path

`multi_verify`'s owned path no longer serializes to reach the shared implementation (it used to archive-then-delegate); checked this isn't a wash or regression via `cargo bench -p stark --bench prover_benchmark --features test-utils -- quick`, comparing this branch against `main`:

| config | main | this branch | Δ |
|---|---|---|---|
| verify/fib_2col_16k | 582.6 µs | 553.2 µs | **−5.0%** |
| verify/fib_4col_8k | 513.0 µs | 490.8 µs | **−4.3%** |
| verify/fib_4col_16k | 590.9 µs | 582.7 µs | −1.4% |

No regression on the regular verify path — dropping the serialization step is a small net win there too. `prove` times are unchanged (within noise) on all three configs, as expected (prove was never touched).

## Verification

Full workspace suite green: `cargo test -p stark --lib` (188, +193 with `disk-spill`), `cargo test -p lambda-vm-prover --lib` (509, incl. continuation and recursion smoke tests), `cargo test -p executor`, `tooling/ethrex-tests`. `cargo clippy -D warnings` clean on both stark feature configurations. `cargo build --workspace` clean.


executor/tests/rust.rs
Oppen, line 1: Moved out of the main workspace due to `rkyv/unaligned` feature bubbling up to `lambda_vm`. Not actually deleted.

MauroToscano (approved):

PR #775, opened 2026-07-03
perf(stark): fuse deep-composition reconstruction for both FRI points

Oppen:
## Summary

- Fuses the regular/symmetric-evaluation-point passes of
  `reconstruct_deep_composition_poly_evaluation` (STARK verifier, FRI step)
  into one traversal, sharing the OOD-table walk, the `trace_term_coeffs`
  walk, and both `inplace_batch_inverse` calls per query. The core trick: `coeff*(base-ood)*denom`
  distributes to `denom*(coeff*base - coeff*ood)`, isolating the
  point-independent `coeff*ood` term (computed once, shared between both
  points) and letting base-field columns use the cheap `IsSubFieldOf`
  asymmetric multiply instead of a full extension-field product.
- Follow-up: dedupes the `z*g^k` ladder walk within a query (was computed
  twice, once per evaluation point), hand-rolls the 2-element
  composition-tail batch inverse to avoid a Vec allocation, and marks the
  `IsSubFieldOf<Degree{2,3}GoldilocksExtensionField>` mul/add/sub impls
  `#[inline(always)]` to match their sibling `IsField` impls.
- Trims a doc-comment sentence that duplicated an inline comment.

## Measured impact

Multi-query recursion profile (`make test-profile-recursion-multi`; the
underlying proof has ~40-45K cycles of run-to-run variance from
`stark::grinding::generate_nonce`'s parallel `find_any` search, so all
deltas below are 2-sample-cluster-to-cluster, not point-to-point):

- Pre-fusion baseline: 2,210,366,539 cycles
- After the fusion (`85fa5b3e`): ~1,947,242,710 cycles (**-11.9%**)
- After the follow-up micro-opts (`3ac5e494`): ~1,939,857,910 cycles
  (a further **-0.38%**, **-14.2%** total from baseline)

Single-query and step-3 (FRI) improvements are proportionally similar; see
individual commit messages for per-commit numbers.

## Soundness

Every guard the two-call version had (base/aux-length checks, the
`trace_term_coeffs` sanity check, batch-inverse zero-rejection, the
composition-poly-parts-count check) is reproduced exactly or made strictly
stricter in the fused version — verified by independent adversarial review
agents against the reject-path semantics, the algebraic identity, and the
soundness threat model (a dishonest prover cannot craft a proof this code
accepts that the prior code would have rejected).

## Test plan

- [x] `make test` (workspace-wide; only failure is `math-cuda`, which needs
      a GPU unavailable in this environment — pre-existing, unrelated)
- [x] `make test-ethrex`
- [x] `cargo test -p lambda-vm-prover --lib test_recursion_execute_1query --
      --ignored --nocapture` (in-VM end-to-end verify, output digest
      unchanged across all three commits)
- [x] Two independent adversarial review passes (soundness + correctness,
      then a second round covering performance/simplicity) on the diff


Oppen:
Superseded by #826

PR #776, opened 2026-07-03
perf(stark): hoist query-invariant deep-composition sums out of the FRI query loop

Oppen:
## Summary
- Hoist sums that depend only on `challenges` and proof-level OOD/gamma data (`ood_row_sum`, `number_of_parts`, `z_pow`, `h_sum_zpow`) out of the per-query `reconstruct_deep_composition_poly_evaluation_pair` into a new once-per-proof `compute_query_invariant_deep_terms`, eliminating the O(num_queries × height × width) `coeff * ood` multiplies down to O(height × width).
- Derive `number_of_parts` from `proof.composition_poly_parts_ood_evaluation.len()` (already validated against the AIR's degree bound in `multi_verify_archived`) instead of each query's own opening length, and check both regular/symmetric per-query openings against that fixed value — strictly stronger than the previous regular-vs-symmetric-only guard, since it also rejects a proof whose per-query opening length varies query to query.

## Numbers
| Metric | Before | After | Change |
|---|---|---|---|
| Single-query total cycles | 86,368,803 | 85,810,707 | −0.65% |
| Single-query `step3:fri` | ≈7.97M | 7,807,225 | ≈−2% |
| Multi-query total cycles | 1,951,764,531 | 1,818,365,778 | −6.84% |
| Multi-query `step3:fri` | ≈652.1M | 529,415,199 | ≈−18.8% |

## Test plan
- [x] `cargo check -p stark`
- [x] `cargo test -p stark` (137 tests pass)
- [x] `make test-ethrex`
- [x] `cargo test -p lambda-vm-prover --lib test_recursion_execute_1query -- --ignored --nocapture` (in-VM verify accepted)
- [x] `make test-profile-recursion-single` / `make test-profile-recursion-multi` (cycle counts above)
- [x] Adversarial soundness review (independent Opus pass): confirmed no reject-path was weakened, the `number_of_parts` guard change is a genuine strengthening, and hoisted sums are algebraically identical to the per-query-recomputed values for honest proofs
- [x] Independent correctness, simplicity, and performance reviews: no issues found

### DIFF
--- a/crypto/stark/src/verifier.rs
+++ b/crypto/stark/src/verifier.rs
@@ -85,6 +85,22 @@ where
 
 pub type DeepPolynomialEvaluations<F> = (Vec<FieldElement<F>>, Vec<FieldElement<F>>);
 
+/// Deep-composition sums that are identical across all FRI queries of a
+/// single proof (see `compute_query_invariant_deep_terms`).
+pub struct QueryInvariantDeepTerms<FieldExtension>
+where
+    FieldExtension: Send + Sync + IsField,
+{
+    /// `ood_row_sum[row] = sum_col trace_term_coeffs[col][row] * ood(row, col)`.
+    ood_row_sum: Vec<FieldElement<FieldExtension>>,
+    /// Derived from `proof.composition_poly_parts_ood_evaluation.len()`.
+    number_of_parts: usize,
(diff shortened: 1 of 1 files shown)


Oppen:
Superseded by #826

PR #792, opened 2026-07-07
Make the u32 value domain a compile-time fact for MemwOperation

MauroToscano:
Third follow-up from the #786 review, stacked on #791. Types-only change; produced traces are byte-identical.

#786 shrank `MemwOperation.value`/`old` to `[u32; 8]` but kept `[u64; 8]` constructor signatures, narrowing with `debug_assert!` + `as u32` — release builds would silently truncate an out-of-domain value (a bus-imbalance debugging nightmare if a future caller ever passes a full 64-bit word), and the M1/M3/M5 register fast path cast `as u32` at its call sites with no assert at all, in any build profile.

This moves the domain to the type system instead of guarding it at runtime:

- `MemwOperation::new` and `with_old` take `[u32; 8]` directly; the narrowing maps and their debug_asserts are deleted. (`old_timestamp` stays `[u64; 8]` — timestamps can reach `u64::MAX`.)
- `pack_register_value` returns u32 limbs, so the fast path's bare casts disappear rather than gaining asserts.
- `MemoryState::read_bytes` returns `[u32; 8]` values (they are u8 memory bytes), and the byte-domain call sites (load, store, commit, keccak, ECSM) build `[u32; 8]` at the source.
- `LoadOperation` keeps its existing u64 interface via a widening `map(u64::from)` at the boundary.

Out-of-domain values are now unrepresentable in every build profile — the "error cleanly" version of the guard, at zero runtime cost.

## Testing
- `make lint` passes; `cargo clippy --all-targets` no warnings; `cargo fmt --check` clean.
- Full `cargo test -p lambda-vm-prover --lib`: 503 passed — identical to the pre-change baseline (the 5 failures are the known missing rust-guest ELF fixtures, failing at file-read before any prover code runs; CI builds them and runs the full suite).

### DIFF
--- a/prover/src/tables/bitwise.rs
+++ b/prover/src/tables/bitwise.rs
@@ -517,6 +517,13 @@ impl BitwiseHistogram {
     /// Increment the counter for one lookup.
     #[inline]
     pub fn bump(&mut self, op: BitwiseOperation) {
+        self.bump_n(op, 1);
+    }
+
+    /// Add `n` occurrences of one lookup in a single step (e.g. CPU padding rows,
+    /// which all send identical all-zero lookups).
+    #[inline]
+    pub fn bump_n(&mut self, op: BitwiseOperation, n: u64) {
         let idx = lookup_type_index(op.lookup_type) * NUM_ROWS + row_index(op.x, op.y, op.z);
         // (x, y) are u8, and row_index debug-asserts z < 16, so in debug builds a
         // corrupt op fails loudly here. In release an out-of-domain z would NOT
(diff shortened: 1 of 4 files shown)


MauroToscano:
Rebased on the updated branch. Note on 3b6840e7 (which upgraded the narrowing guards to always-on `assert!`): this PR resolves the same concern one level deeper — with `[u32; 8]` at the constructor boundary the assert becomes tautological and is deleted. Out-of-domain values are unrepresentable in every build profile, with no release-mode panic path in the trace-gen hot loop.

PR #793, opened 2026-07-07
Feed bitwise collectors straight into the histogram; cut reduce allocations

MauroToscano:
Final follow-up from the #786 review, stacked on #792. Applies the PR's own "never materialize a Vec whose only consumer is the multiplicity count" principle to the remaining big sources, and removes the rayon reduce-identity allocations. Multiplicities are unchanged (same counts, same cells).

## Changes
- **Collectors now bump a `BitwiseHistogram` directly** (`Fn(&mut BitwiseHistogram)`) instead of returning `Vec<BitwiseOperation>`:
  - MEMW_R bumps one IS_HALFWORD per register row — previously an ~4 B/op vector over tens of millions of rows existed just to be counted.
  - PAGE bumps one ARE_BYTES per byte of every touched page — previously a full op vector per page sweep.
  - CPU padding collapses to two `bump_n` calls (3n ARE_BYTES[0,0] + 4n IS_HALF[0,0]) instead of an O(padding_rows) vector of identical zero ops (which also over-reserved 2×).
  - Small sources still use their pure `collect_*` functions; the closure folds the transient Vec in and drops it.
- **`reduce_with` replaces `reduce(identity, ..)`** in the rayon tree-reduce: no more zeroed 80 MiB identity histogram per reduce leaf (up to ~16 × 80 MiB of pure alloc_zeroed + merge traffic on many-core machines).
- **`RegRow.address` shrinks to u16** (register index 0..=255): the walk's largest persisted array goes from 40 to 32 B/row.

### Local tests
`cargo build --release --features instruments`, `cli prove --time`, Apple Silicon, best of 2 runs, this branch vs its base (#792):

| workload | p4_bitwise_collect before | after | Δ |
|---|--:|--:|--:|
| fib_iterative_1M | 112.0 ms | 97.8 ms | −13% |
| fib_iterative_4M | 131.1 ms | 83.3 ms | **−36%** |

`prove_total` and peak RSS unchanged within run-to-run noise on this machine (p4 is <1% of total here, and macOS RSS jitters ~1.5 GB between runs). The reduce-identity savings scale with core count, so the effect should be larger on the many-core provers #786 targets. The remaining serial `base.add_ops` over the in-walk lookups was left as-is: parallelizing it adds code for a phase already down to ~85 ms.

### Server

image

## Testing
- Full `cargo test -p lambda-vm-prover --lib` on this change: 503 passed (same 5 known missing-fixture failures, which fail at file-read).
- `make lint` passes; clippy `--all-targets` no warnings; fmt clean.

### DIFF
--- a/prover/src/tables/bitwise.rs
+++ b/prover/src/tables/bitwise.rs
@@ -517,6 +517,13 @@ impl BitwiseHistogram {
     /// Increment the counter for one lookup.
     #[inline]
     pub fn bump(&mut self, op: BitwiseOperation) {
+        self.bump_n(op, 1);
+    }
+
+    /// Add `n` occurrences of one lookup in a single step (e.g. CPU padding rows,
+    /// which all send identical all-zero lookups).
+    #[inline]
+    pub fn bump_n(&mut self, op: BitwiseOperation, n: u64) {
         let idx = lookup_type_index(op.lookup_type) * NUM_ROWS + row_index(op.x, op.y, op.z);
         // (x, y) are u8, and row_index debug-asserts z < 16, so in debug builds a
         // corrupt op fails loudly here. In release an out-of-domain z would NOT
@@ -524,7 +531,7 @@ impl BitwiseHistogram {
         // mis-count both cells — the proof then fails verification instead of the
         // prover crashing. What actually upholds the invariant is that every
         // `BitwiseOperation` constructor masks or debug-asserts z < 16.
-        self.counters[idx] += 1;
+        self.counters[idx] += n;
     }
 
     /// Fold a slice of lookups into the histogram.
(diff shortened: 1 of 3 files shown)


MauroToscano:
Rebased and composed with 3ba4aac7: kept its `par_chunks` cap (peak memory no longer scales with core count), and layered this PR's two changes on top — collectors bump the chunk histogram directly (no per-source op vectors for MEMW_R / PAGE / padding), and `reduce_with` removes the remaining per-leaf 80 MiB identity allocations that `reduce(identity, ..)` still paid. Full suite re-run on the rebased stack: 502 passed (one test fewer than before — the flaky privacy byte-scan test was removed upstream in #789), same 5 known missing-fixture failures.

PR #802, opened 2026-07-11
perf(crypto): wire keccak_permute ecall into merkle/transcript/grinding

Oppen:
## Summary

- Route the STARK verifier's Keccak-256 usage (Merkle tree backends, Fiat-Shamir transcript, FRI grinding) through the `keccak_permute` precompile on the riscv64 guest, via a `PlatformKeccak256` digest wrapper around `lambda_vm_syscalls::keccak::Keccak256`; falls back to software `sha3` on host.
- Excludes the riscv-only `syscalls` crate from the root workspace (`exclude = ["syscalls"]`): making it reachable from `crypto/crypto` via a target-gated path dependency caused Cargo to auto-adopt it as an implicit workspace member, so `cargo test` tried to run its bare-metal unit tests on host and aborted.
- Adds an end-to-end guest test (`keccak_precompile`) exercising the ecall-backed sponge against known Keccak-256 vectors (empty, rate-1, exactly-rate, multi-block), covering the padding edge cases.

## Cycle comparison (recursion guest, blowup8 preset)

| | before | after | speedup |
|---|---|---|---|
| single query | 5,234,002,718 | 1,722,846,582 | 3.04x |
| multi query | 22,494,934,472 | 7,687,016,693 | 2.93x |

## Test plan
- [x] `make test-profile-recursion-single` / `-multi` before and after (cycles above)
- [x] `test_recursion_execute_1query` — in-VM verify accepts with the ecall-backed hasher
- [x] new `test_keccak_precompile` guest test — known-answer vectors incl. padding edge cases
- [x] `make lint`
- [x] `make test` — all pass except `math-cuda` CUDA tests (expected: no GPU on this machine)


diegokingston (approved):

jotabulacios (approved):

diegokingston (dismissed):
Statement is using still sha3

MauroToscano (changes requested):
We should patch sha3 crate in a fork to use sha3, then only swap the version of sha3 on the guest, not do this

MauroToscano (approved):

MauroToscano:
/bench-verify

> We should patch sha3 crate in a fork to use sha3, then only swap the version of sha3 on the guest, not do this

After further discussion we decided to keep it like this, lgtm

PR #814, opened 2026-07-14
Trivial cleanups from PR #769 review

MauroToscano:
Stacked on #769 (`perf/rkyv-serialization`). Six low-risk, trivial-to-review cleanups surfaced while reviewing that PR — all either doc/comment-only or behavior-preserving refactors with no change to what proofs are accepted, no unsafe-contract semantics, no Merkle bytes, and no CI wiring. The higher-risk findings from the same review (OOD-table width guard, the `NativeArchived` value-validity contract, two-slice leaf hashing, ethrex CI wiring, the view-symmetry guard, the private-input length cap) are intentionally **not** here — they need individual review and are being handled separately.

## Contents

- **chore(prover): drop unused `bincode` dev-dependency.** Nothing under `prover/` references bincode after the rkyv migration; other crates keep their own declarations.
- **docs: fix stale bincode/serde references.** The continuation bundle derives rkyv (not serde) and round-trips via rkyv; `examples_cli` no longer mirrors `bin/cli`'s (now rkyv) proof format.
- **docs(stark): note the proof serde derives are kept only for examples/tests.** rkyv is authoritative; the serde derives survive solely for `examples/examples_cli.rs` and the `serde_cbor` round-trip tests. (We deliberately did **not** remove them — three consumers depend on them and it removes no dependencies.)
- **refactor(syscalls): dedupe `get_private_input`** via `get_private_input_slice().to_vec()`, so the raw-pointer read and its single `unsafe` block live in one place.
- **refactor(stark): avoid cloning the proof in `Verifier::verify`.** `multi_verify` only wraps owned proofs into `StarkProofView::Owned` and delegates; call `multi_verify_views` directly with a borrowed view instead of deep-cloning into a throwaway `MultiProof`.
- **refactor(stark): unify owned/archived `Table::dimensions_consistent` and `into_frame`.** Give owned `Table` a `dimensions_consistent` method (reading length through `row_major_data()` so it stays correct under `disk-spill`) with both `StarkTableView` arms delegating, and collapse the two verbatim `into_frame` copies into one `StarkTableView::into_frame`. No behavior change.

## Verification

`cargo check -p stark` (default + `disk-spill --tests`), `cargo check -p lambda-vm-prover`, and full `make lint` (fmt + all clippy passes) are clean. The syscalls change is guest-only (`riscv64`-gated) and was compile-checked on the guest target.


Oppen:
Had to cherry-pick the commits by hand because my local git seems broken, but it's already merged to #769
Leaving it open so you can check, when you're satisfied close this one.

MauroToscano:
Superseded — these six commits were incorporated directly into `perf/rkyv-serialization` (rebased onto current main). Verified the content is identical for all touched files; closing as merged-by-cherry-pick.

PR #817, opened 2026-07-15
perf(transcript): sample FRI query indices via a masked bit reservoir

diegokingston:


MauroToscano (changes requested):
Putting this on pause for now: measured at the production regime (see the "Measured" section in the description), the improvement is too small right now — −0.32% guest cycles / −0.47% keccak — to prioritize a Fiat–Shamir sampling-schedule change. The approach and numbers check out; we'll revisit when the bigger levers have landed.

PR #825, opened 2026-07-15
bench(recursion): measure the verifier at real query counts, over real blocks

MauroToscano:
## What

The recursion cycle benchmark previously measured one regime: `MIN_PROOF_OPTIONS` (blowup=2, **1 FRI query**) verifying a proof of the **empty program**. That's a useful diagnostic floor, but the production verifier runs 219 queries at blowup 2, and the real inner object is a multi-epoch continuation bundle of an actual block. This PR makes the benchmark able to measure the real target.

### Presets & guests
- `Preset::Blowup2` (219 q) and `Preset::Blowup4` (110 q) — the realistic base-layer regimes — alongside `Min`/`Blowup8`; one guest ELF per preset (`RECURSION_VERIFIER_PRESETS` in the Makefile).
- New `continuation` guest feature → `recursion-cont-{min,blowup2,blowup4}.elf`: verifies a whole `ContinuationProof` bundle in-VM instead of a monolithic `VmProof`, so the inner prove is memory-bounded (any block size proves on a bounded-RAM box).
- `verify_continuation_with_roots` + `continuation_precomputed_commitments`: supplied DECODE / page-genesis roots skip the in-VM FFT+Merkle recomputes. Trust model is identical to the monolithic guest: roots are folded into the attestation's `program_id`; the binding is the consumer's recompute+compare.

### Benchmark plumbing
- `test_dump_recursion_input` env knobs: `RECURSION_DUMP_PRESET`, `RECURSION_DUMP_INNER_ELF/INPUT` (any inner program), `RECURSION_DUMP_EPOCH_LOG2` (continuation mode).
- `bench_recursion_cycles.sh`: preset-aware dump (fails loudly for refs predating it), `--release` test invocations, regime labels; `/bench-verify` now posts **min + blowup2 + blowup4** tables (builds cached; each extra regime ≈ a blob dump + one ~1–2 min execute per ref).
- New `bench_recursion_scaling.sh`: block-size × preset ladder over **committed** ethrex fixtures (1/4/8/16 transfers, ~13–30 KB each, un-ignored explicitly); preset-major sweep so the headline curve completes first.
[...]

PR #826, opened 2026-07-15
perf(stark): fuse and hoist deep-composition reconstruction for both FRI points

Oppen:
## Summary
- Reimplements the `perf/deep-hoist` branch's deep-composition fuse+hoist (originally against an older `verifier.rs`) on top of current `main`'s `StarkProofView`/precomputed-column-split API.
- Fuses the regular/symmetric FRI-point reconstruction into one traversal, and hoists the point-invariant OOD/gamma sums out of the per-query loop into a once-per-proof `compute_query_invariant_deep_terms`.
- Adds a stricter guard: per-query opening length is now checked against the proof-level `composition_poly_parts_ood_evaluation` length, not just regular-vs-symmetric agreement.

## Numbers

The headline figure is the **base-layer regime**: blowup 2 / 219 FRI queries, verifying a real ethrex block (4 transfers, 2 epochs @ 2^21) through the continuation guest. Measured with the PR #825 ladder (`scripts/bench_recursion_scaling.sh`, `TXS=4 PRESETS=blowup2`), base vs head differing only by this PR's `verifier.rs`:

| | Guest cycles |
|---|---|
| before (`a8a3d870`) | 13,364,620,265 |
| after (`2af84167`) | 10,649,255,581 |
| Δ | **−20.32%** (1.255×) |

Keccak calls are identical on both (3,152,604) — the win is field-op restructuring, not hashing. Sanity check: the `before` figure is +0.94% over the recorded post-#769 ladder value for this cell (13.24B), which is #815's own cost.

The other numbers on this PR measure different regimes and should not be read as the Ethrex-relevant win:

| Regime | Δ | What it is |
|---|---|---|
| blowup2 / 219q / real block | **−20.3%** | the base-layer regime (above) |
| `make test-profile-recursion-multi` | −30.8% total, −63.2% step 3 | `Preset::Blowup8` over the **empty** guest — final-wrap params, tiny tables, so deep-composition holds a larger share of the total and the % overstates |
| `/bench_verifier` guest cycles | −5.34% | `Preset::Min` = **1 FRI query**, empty program — the per-query loop this PR optimizes is barely exercised, so the % understates |
| `/bench_verifier` verify time | −3.95% | native host verify, not in-VM |

For reference, the `test-profile-recursion-multi` figures:

| | Total cycles | step 3 (FRI) |
|---|---|---|
| before | 1,325,335,927 | 635,418,644 |
| after | 916,824,543 | 233,869,693 |
| Δ | **-30.8%** | **-63.2%** |

Note that `test-profile-recursion-multi` cannot express the base-layer regime: its presets are only `Min` (blowup 2, 1 query) and `Blowup8`. The blowup2/219q preset lands with #825.

## Test plan
- [x] `cargo test -p stark --lib` — 190/190 passing
- [x] `cargo clippy -p stark --all-targets` — clean
- [x] `make test-profile-recursion-multi` before/after (numbers above)
- [x] #825 ladder, `TXS=4 PRESETS=blowup2 EPOCH_LOG2=21`, base vs head (headline numbers above)

### DIFF
--- a/crypto/stark/src/verifier.rs
+++ b/crypto/stark/src/verifier.rs
@@ -84,6 +84,22 @@ where
 
 pub type DeepPolynomialEvaluations<F> = (Vec<FieldElement<F>>, Vec<FieldElement<F>>);
 
+/// Deep-composition sums that are identical across all FRI queries of a
+/// single proof (see `compute_query_invariant_deep_terms`).
+pub struct QueryInvariantDeepTerms<FieldExtension>
+where
+    FieldExtension: Send + Sync + IsField,
+{
+    /// `ood_row_sum[row] = sum_col trace_term_coeffs[col][row] * ood(row, col)`.
+    ood_row_sum: Vec<FieldElement<FieldExtension>>,
+    /// Derived from `proof.composition_poly_parts_ood_evaluation().len()`.
+    number_of_parts: usize,
+    /// `challenges.z.pow(number_of_parts)`.
+    z_pow: FieldElement<FieldExtension>,
+    /// `sum_j composition_poly_parts_ood_evaluation[j] * challenges.gammas[j]`.
+    h_sum_zpow: FieldElement<FieldExtension>,
+}
+
 // The verifier reads proofs in place from their rkyv archive; archived field
 // elements are viewed as native ones, which is only valid on little-endian.
 #[cfg(not(target_endian = "little"))]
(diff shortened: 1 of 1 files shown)


crypto/stark/src/verifier.rs
```diff
@@ -764,68 +842,96 @@ pub trait IsStarkVerifier<
...
+        // proof-level `number_of_parts`, not just a regular/symmetric mismatch.
+        if lde_composition_poly_parts_evaluation.len() != number_of_parts
+            || lde_composition_poly_parts_evaluation_sym.len() != number_of_parts
+        {
+            return None;
+        }
+        let z_pow = &query_invariant_terms.z_pow;
+
+        // A malformed proof can make evaluation_point == z_pow, reject.
+        let mut denom_composition_pair = [evaluation_point - z_pow, evaluation_point_sym - z_pow];
```
diegokingston, line 918: I think we can do this even better, by computing all the terms for all the pairs from the start, and doing just one batch inversion, which is significantly cheaper than doing number_of_queries*batch_inversion

MauroToscano:
I confirmed these numbers with the real Ethrex program proof: −20.32% (1.255×) │

crypto/stark/src/verifier.rs
MauroToscano: This is something that I think would be included with your PR, we can skip it for this one

MauroToscano (approved):

PR #829, opened 2026-07-16
perf(verifier): factor shared zerofier + batch DEEP inversions

diegokingston:
Two extension-field arithmetic redundancies on the STARK verify path, each turning a per-item inversion into shared work. **Stacked on #823** (base is `feat/logup-acc-current-row`, so the diff is just these two changes; the DEEP one builds on that PR's pruned loop). Retarget to `main` once #823 lands.

## Why

Profiling the ABBA/ethrex verify (204 MiB, ~3.8s) showed the wall-clock is ~85%+ **Keccak Merkle-path hashing, single-threaded** — these arithmetic wins won't move that number. But they cut *serial extension-field inversions*, which is exactly what the **recursion-guest cycle count** pays for (the guest runs this same `verifier.rs`, and a cubic-extension inversion is a long addition chain). So they're aligned with the g·z-pruning cycle-reduction work, not the native 3.8s.

## step_2 — shared zerofier

Every transition constraint's OOD zerofier is `1/(zᴺ − 1)` × an end-exemptions correction `∏(z − rᵢ)` that depends only on `end_exemptions`. Before, each constraint recomputed `zᴺ` (a `pow`) **and a fresh extension inversion**, and the sum multiplied every term by its own denominator. Now: compute `1/(zᴺ − 1)` once, group constraints by `end_exemptions` (almost always the single group `{0}`), accumulate `Σ βᵢ·evalᵢ` per group, and factor the shared inverse + each group's correction out. Removes ~C `pow`s and ~C inversions per table (C = #transition constraints) — mirroring the prover's existing grouped-zerofier dedup. Also turns a `zᴺ = 1` hit from a panic into a clean rejection.

## step_3 — batched DEEP inversions

`reconstruct_deep_composition_poly_evaluation` ran two tiny per-query inversions (a batch-inverse over ~2 denominators + a lone composition-denominator inversion), `2·num_queries` times per table — defeating Montgomery amortization. The `z·gᵏ` row points and `z^parts` are query-independent, so hoist them once, collect every query point's `(h+1)` denominators, and run **one** batch inverse for the whole proof. Collapses ~`4·num_queries` extension inversions/table → ~1. The inner function now takes the pre-inverted denominators (its `evaluation_point`/`primitive_root` args drop out). Malformed-proof rejection is preserved — a denominator on an OOD point fails the single batch inverse.

## Tests

Full `stark` lib suite green (198), including multi_prove roundtrips, soundness negatives (tampered/truncated OOD + composition), and the archived read-in-place path. clippy clean.


nicole-graus (approved):

MauroToscano (commented):

Two refactors: step_2 regroups Σ βᵢ·evalᵢ into `end_exemptions` buckets scaled by ∏(z−rᵢ) and a shared 1/(zᴺ−1), and step_3 batches the DEEP denominators across queries. Both are algebraically equivalent to what they replace, no transcript absorb or squeeze moves, and both `.inv()` / `inplace_batch_inverse(...).ok()?` fail closed. Two issues.

**crypto/stark/src/verifier.rs:338** — the −12.23% guest-cycle headline is the `min` preset: one FRI query at blowup 2, i.e. the regime that flatters this PR most. The step_2 saving (one `pow` plus one extension inversion per transition constraint, per table) is query-invariant, so its absolute value is fixed while the baseline grows roughly linearly with queries; at ~219 queries for an inner proof its share drops by about two orders of magnitude. The DEEP half scales the other way (~2 inversions per query per table), but the net will not be ~12%. The bench baseline `68a120a6` is six commits behind, and the run itself reported that blowup2/blowup4 full-query regimes needed `RECURSION_DUMP_PRESET` on `origin/main`. #846 landed `RECURSION_DUMP_PRESET` and `RECURSION_DUMP_EPOCH_LOG2` on main on 07-21, so a rebase plus a fresh `/bench-verify` can run the full-query and real-block regimes. Book that number, not the single-query one.

**crypto/stark/src/constraints/zerofier.rs:142** — `end_exemptions_correction` re-derives the backward root walk line for line from `end_exemptions_roots` (same `decrement = g^(N−1)` seed, same `current = current * decrement` step, lines 31-37). The old `evaluate_zerofier` reused the shared helper; the rewrite inlines it, and the new `correction_matches_direct_exempt_product_over_cubic_extension` test exists only to catch the two copies drifting. The inlining saves one `Vec` of ≤2 elements per distinct `end_exemptions` per table — negligible against the extension products. `end_exemptions_roots` has no callers outside this file, so it can take `end_exemptions: usize` instead of `&ConstraintMeta` (its only use of `meta` is `meta.end_exemptions`, and `end_exemptions_lde_evaluations` already has that value), with the correction becoming a fold over its result — `roots.iter().fold(one, |acc, r| acc * -(r.clone() - z.clone()))`, the deleted body. One root derivation, regression test still applies.

Nits:
- verifier.rs:915 — `for (i, _iota) in challenges.iotas.iter().enumerate()` binds an iota only to discard it; `num_queries` is in scope at line 853 and is the quantity both the openings guard and the `denominators` capacity are stated in, so `for i in 0..num_queries` reads better.
- verifier.rs:960 — the call site computes `primary_base`/`sym_base` and hand-slices four arguments into a 16-parameter function that re-checks the trace-slice lengths at 1041-1046, splitting the `[trace rows | composition]` stride between caller and callee; passing the two per-point `stride`-sized chunks and splitting at `ood_height` in the callee drops two parameters and the offset arithmetic.
- The PR body's "~4·num_queries extension inversions/table → ~1" is the pre-fuse shape: on this base the function is already the fused pair (one call per iota, two real inversions), so it is 2·num_queries as the comment at line 882 states, and the 4· figure re-claims the halving #826 already shipped.

PR #833, opened 2026-07-16
fix(stark): review follow-ups for OOD pruning (docs, saturating_sub, hard assert)

MauroToscano:
Review follow-ups for #823 (forward LogUp accumulation + next-row OOD opening pruning). Docs and consistency only; no behavior change.

- **Stale doc in `logup_single_source_tests` (`lookup.rs`)**: the module doc claimed the 2-absorbed branch "reads `aux(1, ·)` next-row cells". That described pre-#823 backward accumulation; after forward accumulation both branches' only next-row read is `acc_next`, and absorbed operands read offset 0. Replaced with the branch's real distinction — it folds two absorbed interactions (degree 3).
- **Narrow doc in `ood.rs` module header**: reworded the OOD-table illustration so it is no longer pinned to `offsets [0, 1]`. The layout is generic over `transition_offsets.len()` (repo has `[0]` and `[0,1,2]` example AIRs): offset 0 is the current-row block, every later offset contributes a next-row block.
- **Consistency nit in `verifier.rs`**: `num_eval_points - step_size` → `num_eval_points.saturating_sub(step_size)`, matching the shared `ood.rs` helper for the identical quantity. Purely defensive; the underflow is unreachable (would require an AIR with empty `transition_offsets`, which does not exist and is not proof-controlled).
- **Doc-only precondition note in `build_pruned_trace_term_coeffs` (`ood.rs`)**: added a `Precondition:` line to the doc comment stating that `powers.len()` must equal `num_surviving_trace_openings(...)` for the same layout args (both AIR-metadata-derived, invariant I3), checked in debug builds. Code is unchanged: the per-slot `if p < powers.len()` guards and the `debug_assert_eq!` power-count check stay. Rationale: this is a shared helper the verifier calls, and the verifier must never contain a panic path — an invalid proof is a false proof, not a crash — so the release-mode bounds check is intentional and stays.

**Also in this PR** (clears the base branch's latent `Lint` failure, which is inherited here):
- Clippy fixes in the #823-added forward-accumulation tests (`lookup.rs` `accumulated_column_is_forward_and_circular`, `logup_gpu.rs` `reference_accumulate`): iterate slices directly instead of `0..n` (needless_range_loop) and deref instead of `.clone()` on `Copy` field elements (clone_on_copy). `make lint` runs `cargo clippy --workspace --all-targets`; `--all-targets` covers these test targets. The `logup_gpu.rs` one only surfaces under the cuda-feature clippy pass.
- A one-line rustfmt reflow of an over-width test call in `ood.rs`, the only `cargo fmt --check --all` violation on the base.

Validation: `make lint` (fmt `--check --all` + all four `--workspace --all-targets` clippy passes incl. cuda) green; `cargo test --release -p stark` passes `ood::` (4), `logup_single_source` (7), and `bus_tests` (75).

### DIFF
--- a/crypto/stark/src/logup_gpu.rs
+++ b/crypto/stark/src/logup_gpu.rs
@@ -989,7 +989,7 @@ mod tests {
         let mut out = Vec::with_capacity(num_rows);
         for row in 0..num_rows {
             // Forward accumulation: acc[0] = 0, fold the current row afterwards.
-            out.push(acc.clone());
+            out.push(acc);
             let mut rs = FieldElement::<E>::zero();
             for c in cols {
                 rs = &rs + &c[row];
--- a/crypto/stark/src/lookup.rs
+++ b/crypto/stark/src/lookup.rs
@@ -2432,11 +2432,11 @@ mod logup_single_source_tests {
         let n_fe = Fp3::from(n_rows as u64);
         for i in 0..n_rows {
             let mut row_sum = Fp3::zero();
-            for c in 0..n_term_cols {
-                row_sum = row_sum + &term_columns[c][i];
+            for col in &term_columns {
+                row_sum = row_sum + &col[i];
             }
-            let acc_i = trace.get_aux(i, acc_col_idx).clone();
-            let acc_next = trace.get_aux((i + 1) % n_rows, acc_col_idx).clone();
+            let acc_i = *trace.get_aux(i, acc_col_idx);
+            let acc_next = *trace.get_aux((i + 1) % n_rows, acc_col_idx);
             let lhs = (acc_next - acc_i) * &n_fe;
             let rhs = row_sum * &n_fe - &l;
             assert_eq!(lhs, rhs, "forward circular recurrence broken at row {i}");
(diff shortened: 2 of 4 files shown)


MauroToscano:
Second commit (009f6de3) clears the `Lint` job, which was latently red on the base branch: `make lint` runs `cargo clippy --workspace --all-targets`, and the `--all-targets` flag covers test code that the initial crate-scoped lib clippy skipped. Five `-D warnings` errors in tests added on the #823 branch:

- `lookup.rs` `accumulated_column_is_forward_and_circular`: iterate `&term_columns` instead of `0..n_term_cols` (needless_range_loop, two sites) and deref the two `get_aux(..).clone()` reads (clone_on_copy — `FieldElement<Degree3GoldilocksExtensionField>` is `Copy`).
- `logup_gpu.rs` `reference_accumulate`: `out.push(acc.clone())` → `out.push(acc)` (clone_on_copy). This one only surfaces under the cuda-feature clippy pass (`make lint`'s 4th clippy invocation), which the earlier passes stop short of.

Verified locally by running the exact CI `make lint` (fmt --check --all + all four `--workspace --all-targets` clippy passes incl. `--features lambda-vm-prover/cuda`) to green, and re-running the release-mode `ood::`, `logup_single_source`, and `bus_tests` suites (plus the rewritten test) — all pass. Test semantics are unchanged.

PR #842, opened 2026-07-16
perf(verifier): compute shared LogUp α-powers once, slice per table

diegokingston:
Stacked on #830.

## What
`step_2_verify_claimed_composition_polynomial` rebuilt the LogUp α-power vector `[1, α, α², …]` (up to the table's bus width) **once per table** via `compute_alpha_powers`, even though the LogUp α challenge is a **single value sampled once** in `multi_verify_views` (Round 1 Phase B) and shared across every table. Each table's vector is just a prefix of the same geometric sequence.

This computes the sequence **once**, right after sampling the shared challenges, to the global-max bus width across all AIRs, and each table's `step_2` slices the prefix it needs.

## Why
The verifier doubles as the serial recursion guest, so per-table recompute costs real cycles:
- Saves `(Σ_tables max_bus_elements − max_t max_bus_elements)` extension-field multiplications per proof (E×E ≈ 9 base muls each).
- Saves `N−1` heap allocations per proof — each heap alloc is a proven memory-table write in the guest.

## Correctness
Byte-identical: the sliced prefix `&shared_alpha_powers[..air.max_bus_elements()]` equals the vector each table computed before. When the proof carries LogUp challenges, `shared_alpha_powers.len()` is the global max, so every table's slice is in range; otherwise both old and new paths yield an empty slice. Comment at `lookup.rs:96` documents that α *must* be shared across all AIRs.

## Testing
- 204 stark tests pass (real prove→verify roundtrips, incl. multi-table LogUp).
- clippy clean; prover crate builds.
- Recursion-cycle benchmark: server.

### DIFF
--- a/crypto/stark/src/verifier.rs
+++ b/crypto/stark/src/verifier.rs
@@ -196,6 +196,7 @@ pub trait IsStarkVerifier<
             && next.height() == expected_next_height
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn step_2_verify_claimed_composition_polynomial(
         air: &dyn AIR<Field = Field, FieldExtension = FieldExtension, PublicInputs = PI>,
         proof: StarkProofView<'_, Field, FieldExtension, PI>,
@@ -209,6 +210,10 @@ pub trait IsStarkVerifier<
         // split below.
         ood_full: &Table<FieldExtension>,
         step_size: usize,
+        // Powers of the shared LogUp α challenge, precomputed once by the caller to
+        // the global-max bus width across all tables (α is sampled once and shared),
+        // then sliced here to this table's width — no per-table recompute or alloc.
+        shared_alpha_powers: &[FieldElement<FieldExtension>],
     ) -> bool {
         crate::profile_markers::step_marker::<
             { crate::profile_markers::STEP_VERIFY_CLAIMED_COMPOSITION_POLYNOMIAL },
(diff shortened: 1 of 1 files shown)


MauroToscano (commented):

Hoists the α-power sequence into `multi_verify_views` and slices it per table, replacing the per-table `compute_alpha_powers` rebuilds of prefixes of one geometric sequence. The transform is equivalent: same α, same prefix, no absorb or squeeze moved, and the build-site guard (`lookup_challenges.len() > LOGUP_CHALLENGE_ALPHA`) and the slice-site guard (`challenges.rap_challenges.len() > …`) read the same vector because `replay_rounds_after_round_1` stores `rap_challenges` unmodified. Two items before merge.

**.github/scripts/run_recursion_bench.sh:18** (and `.github/workflows/bench-verify.yml:111`) — the posted `−5.4M cycles (−12.44%)` is not this change's win. Both benches hardcode `origin/main` as the baseline, but this PR is based on `perf/verifier-amortize-tier01`, so REF_B `a8648320` is the merge-base and the delta also contains `4d4d8868` (shared zerofier + batched DEEP inversions) and `1ea78769` (reused trace root + shared Q FRI points) — both on per-query paths that run 110–219 times per proof, versus this change's once-per-table work. This commit's own saving is `(Σ_tables max_bus_elements − max_t max_bus_elements)` extension muls plus ~N−1 small allocations over ~31 tables: tens of thousands of guest cycles, under the script's own stated ±100k noise floor. The native `−0.09% BORDERLINE` matches that. Re-run against `1ea78769`, or state in the body that the headline is stack-cumulative so it isn't credited again when #830 lands. The workflow should use the PR's base ref (or merge-base with it) rather than hardcoded `origin/main`, otherwise every stacked verifier PR inherits this.

**crypto/stark/src/verifier.rs:306** — `&shared_alpha_powers[..air.max_bus_elements()]` is a panicking index on a verifier-executed path, justified by a comment about an invariant maintained two frames up. The invariant holds today (one call site for `verify_rounds_2_to_4`, one empty trait impl, both lengths derived from the same `airs` slice, `max_bus_elements` is AIR metadata and not proof-derived), so it is not attacker-reachable. But after the #833 revert the bar is no panic on a verifier path even for AIR-metadata invariants, since a panic in the recursion guest kills execution instead of yielding a rejectable proof. Keeping the property local costs nothing:

```rust
let logup_alpha_powers: &[FieldElement<FieldExtension>] =
    shared_alpha_powers.get(..air.max_bus_elements()).unwrap_or(&[]);
```

Behaviourally identical on every reachable state — when no challenges were sampled the shared vector is empty and `get` yields `&[]`, exactly the old `else` arm — it deletes the branch and fails closed. `match … { Some(p) => p, None => return false }` inside the existing guard works equally well if rejecting loudly is preferred.

Nits:
- verifier.rs:302 — "when the proof carries LogUp challenges" implies the prover decides whether challenges exist; they are verifier-sampled and gated on `needs_lookup_challenges = airs.iter().any(|air| air.has_aux_trace())`, pure AIR layout. Same phrasing in the PR body.
- verifier.rs:1379 — `lookup_challenges.clone()` sits directly above the new argument and is a 2-element Vec allocation per table, the same per-table alloc cost this PR removes; fixing it needs a lifetime on `Challenges`, but it caps the allocation win being claimed.

PR #843, opened 2026-07-16
fix(syscalls): raise the private-input clamp to 512 MiB

Oppen:
## Summary
- 64 MiB was sized for the old naive-recursion `VmProof` case. Real proofs at production options (blowup=2, 219 queries) are much larger — ethrex/20 proves to ~231 MiB monolithic, and a continuation bundle carries one such proof per epoch.
- Bumps both sides of the clamp together: the executor's cap (`executor::vm::memory::MAX_PRIVATE_INPUT_SIZE`) and the guest's mirrored constant (`syscalls::syscalls::MAX_PRIVATE_INPUT_SIZE`), so an honest length prefix stays within bound on both — a prior version of this change bumped only the host side and silently truncated real guest inputs.

## Test plan
- [x] `cargo test -p lambda-vm-prover --lib` (continuation page-count regression test updated: 257 → 2049 pages)
- [x] `cargo test -p executor --test flamegraph test_run_with_flamegraph_returns_generator_on_executor_new_failure`


Oppen:
Addressed the two cheap review findings (67bc664a):
- `test_verify_rejects_num_private_input_pages_exceeds_max` now uses `page::max_private_input_pages() + 1` instead of a hardcoded `1000`, which the 512 MiB bump had made a false-pass (1000 pages is now within bound).
- Fixed the stale "64 MiB" comment in `continuation.rs`.

**Known issue, tracked separately (not fixed in this PR):** the reviewers are right that `Memory::cells` (a `HashMap<u64, [u8;4]>`) backing the private-input region doesn't scale to the new 512 MiB cap — measured peak RSS for a single `store_private_inputs` call at exactly `MAX_PRIVATE_INPUT_SIZE` is **~9.7 GB** on this per-word hashmap design. We prototyped 5 alternative backing strategies (dense Vec, paged HashMap with Vec/inline-array/Box-array values) — all bring peak RSS down to ~1-2 GB and store time from ~2.2s to 30-90ms. Follow-up PR against `main` will land the chosen redesign. Merging this PR as-is since it unblocks other work and the clamp/wire-format correctness is unaffected — the memory-overhead issue is a resource-usage risk, not a soundness bug.

jotabulacios (approved):

Oppen:
Memory explosion will be fixed in a follow-up PR. For now it's rather theoretical.

MauroToscano (approved):

PR #844, opened 2026-07-16
feat(recursion): supply DECODE/global-memory-genesis roots via private input for continuation verify

Oppen:
## Summary
Mirrors #782's monolithic mechanism for the continuation path: a caller (the recursion guest) can supply the DECODE preprocessed root and each touched data page's genesis root instead of the verifier recomputing them from the ELF, skipping the in-VM FFT + Merkle build. Supplied roots are used verbatim; the binding shifts to the consumer's recompute-and-compare of the folded identity, exactly like the monolithic path.

- `verify_global`'s genesis-page classification (ELF-backed vs zero-init) is derived from ELF segment address ranges instead of materializing a full byte-level image when roots are supplied, since the real bytes are never read once a root covers that page.
- Page lookups are keyed by page number (`page_base >> log2(page_size)`) rather than the raw page-aligned address, since the low bits are always zero.
- `verify_continuation` keeps its existing signature (trustless recompute); `verify_continuation_with_roots` is the new supplied-roots entry point, and `continuation_precomputed_commitments` lets a caller derive the roots to supply for a given bundle.

This PR does not change the wire format or add a guest-facing API — that's `perf/continuation-rkyv-zero-copy`, stacked on top of this one.

## Test plan
- [x] `cargo test -p lambda-vm-prover --lib` (515 passed)
- [x] New test: `test_verify_continuation_with_supplied_roots` — supplied-roots verify matches the trustless recompute, and a tampered supplied root is rejected
- [x] `cargo clippy --all-targets -- -D warnings` / `cargo fmt --check`


Oppen:
On Codex feedback:
1. We do that everywhere AFAIK
2. No real program has no zero-init pages: they all use the stack, and most have at least some zero-init global, not a realistic problem
3. We control the blowup and match it to real usage

MauroToscano (approved):

PR #846, opened 2026-07-16
bench(recursion): measure the verifier at real query counts, over real blocks

Oppen:
## Summary
Stacked on #845. The recursion cycle benchmark only measured `MIN_PROOF_OPTIONS` (blowup=2, 1 FRI query) verifying a proof of the empty program — a diagnostic floor, not the production verifier cost (blowup=2 at 128-bit is 219 queries, and real inner proofs are multi-epoch continuation bundles of real blocks).

Presets and guests:
- `Preset::Blowup2` (219 q) / `Blowup4` (110 q) — the realistic base-layer regimes — alongside the existing Min/Blowup8; one guest ELF per preset.
- `bench_vs/lambda/recursion`'s `continuation` guest feature: verify a whole `ContinuationProof` bundle in-VM (`recursion-cont-<preset>.elf`), built on #844/#845's continuation supplied-roots verify path and zero-copy guest API.

Benchmark plumbing:
- `test_dump_recursion_input` gains env knobs: `RECURSION_DUMP_PRESET`, `RECURSION_DUMP_INNER_ELF`/`INPUT` (any inner program, e.g. ethrex blocks), `RECURSION_DUMP_EPOCH_LOG2` (continuation mode).
- `scripts/bench_recursion_cycles.sh`: preset-aware dump, `--release` test invocations, per-preset regime labels; `/bench-verify` now posts min + blowup2 + blowup4 tables.
- `scripts/bench_recursion_scaling.sh`: block-size × preset ladder over the committed ethrex fixtures (1/4/8/16 transfers).
- A real-block detailed profile test (blowup=4, 4 transfers) plus a `make test-profile-recursion-block` target, since cycle counts across every preset/block size are already covered by the scripts above.

**Note**: this branch currently includes a cherry-picked copy of #843's commit (`fix(syscalls): raise the private-input clamp to 512 MiB`) so the real-block tests are runnable now (real blocks exceed the old 64 MiB cap). That commit will disappear from this diff once #843 merges and this branch is rebased onto the updated base.

## Test plan
- [x] `cargo test -p lambda-vm-prover --lib` (516 passed)
- [x] `test_recursion_profile_blowup4_block` run end-to-end against a real 4-tx ethrex block (not just compiled) — produced the full cycle/step/function breakdown
- [x] `cargo clippy --all-targets -- -D warnings` / `cargo fmt --check`
- [x] `make -n test-profile-recursion-block` resolves the full build graph correctly


Oppen:
Note: `MAX_PRIVATE_INPUT_SIZE` findings come from and are handled in #843

nicole-graus (approved):

PR #854, opened 2026-07-21
spec: `REG(W)` templates

erik-3milabs:
This PR introduces templates for the short-hand notation of register reads and writes.

Note: this PR is awaiting `spec/streaming` to get merged first.


spec/src/commit.toml
```diff
@@ -81,35 +81,35 @@ ref = "commit:c:receive_ecall"
...
 
 [[constraints.read_input]]
-kind = "interaction"
-tag = "MEMW"
-input = [1, ["cast", ["*", 2, 254], "DWordWL"], ["arr", ["+", "index", ["cast", "count", "BaseField"]], 0, 0, 0, 0, 0, 0, 0], "timestamp", 0, 0, 0]
-output = ["arr", "index", 0, 0, 0, 0, 0, 0, 0]
-multiplicity = "first"
+kind = "template"
+tag = "REG"
+input = [254, ["arr", ["+", "index", ["cast", "count", "BaseField"]], 0], "timestamp"]
```
erik-3milabs, line 111: this conversion is a touch tricky, since it now writes to both limbs, instead of just the lower. should be fine though. let me know what you think.

spec/reg.typ
```diff
@@ -0,0 +1,43 @@
...
+#let write = raw(reg_write.name)
+
+#show: book-page("REG")
+
+We provide the #read and #write templates.
+These templates act as short hand notation for `MEMW` interactions pertaining to operations in the `register` domain ($= 1$, see @memory).
+Interactions on this domain can denote more concisely since
++ the register address can be represented using a one-column `Byte` rather than a two-column `DWordWL`, and
++ registers accesses (almost) always have a width of 2.
+Moreover, the factor 2 that must be included in converting from register-number to the register-address has led to several hard-to-catch errors in the past; introducing this abstraction will reduce the likelihood of these mistakes.
```
RobinJadoul: This feels more like a PR description than part of a spec.

(not entirely convinced by my own suggestion yet either, but maybe)
```suggestion
Moreover, the factor 2 that must be included in converting from register-number to the register-address can be error-prone, so this reduces the likelihood of these mistakes in writing chips.
```

spec/reg.typ
```diff
@@ -0,0 +1,43 @@
...
+
+The necessity for two templates follows from the existence of two subtly different `MEMW` signatures (see @signatures): there is both a read-write and an write-only interaction.
+Here, the former is wrapped by #read, while the #write template encapsulates the latter.
+
+For both templates hold: when not provided, `cond` defaults to `1`.
```
RobinJadoul: Weird sentence

spec/src/memw_register.toml
```diff
@@ -121,14 +121,14 @@ ref = "regw:c:write_val"
 name = "output"
 
 [[constraints.output]]
-kind = "interaction"
-tag = "MEMW"
-input = [1, ["arr", ["cast", ["*", 2, "address"], "Word"], 0], ["arr", ["idx", "val", 0], ["idx", "val", 1], 0, 0, 0, 0, 0, 0], "timestamp", 1, 0, 0]
-output = ["arr", ["idx", "old", 0], ["idx", "old", 1], 0, 0, 0, 0, 0, 0]
-multiplicity = ["-", "μ_read"]
+kind = "template"
```
RobinJadoul: This feels like a strange thing to do.
It's correct, and probably readable, I'm just not sure if the providing side of a signature should be a template.

spec/src/commit.toml
erik-3milabs: @RobinJadoul

spec/src/memw_register.toml
erik-3milabs: agreed. reverted it.

spec/reg.typ
erik-3milabs: I rewrote the paragraph. let me now whether you agree that it is better now.

spec/reg.typ
erik-3milabs: Agreed. Better now?

spec/src/commit.toml
RobinJadoul: Yeah, should be fine

RobinJadoul (approved):

PR #856, opened 2026-07-21
fix(bench-verify): Measure each recursion ref's guest cycles with a CLI built from that ref

nicole-graus:
## Motivation

`/bench-verify`'s recursion cycle bench builds a single measuring CLI from the workflow checkout, which, in the `issue_comment` flow, is the default branch (`main`), and runs it against both the baseline and PR guest ELFs. That works only while every guest uses syscalls `main` already knows. The moment a PR's guest emits a new syscall the main executor aborts with `ExecutionError(UnknownSyscall(...))` and the cycle bench fails, even though the PR is correct. PR #818 (wiring the recursion verifier's Fp3 mul-add through the FEXT accelerator) is the first to hit this: its guest emits `FEXT_LOAD (u64::MAX - 19 = 18446744073709551596)`, which `main` doesn't recognize. It will recur for any future PR that adds an accelerator/syscall.

## Description

Build the measuring CLI per ref instead of once: the baseline guest is measured by a `cli` built from the baseline ref and the PR guest by a `cli` built from the PR ref, each inside the worktree the script already creates for that ref's guest build. Each VM therefore understands exactly its own guest's syscall, robust for PRs that add or remove syscalls in either direction. This mirrors the existing per-side build in `scripts/bench_verify.sh`. The result-cache key drops its counter-SHA component (the ref SHA now identifies the counter) and the CLI build shares `HOST_TARGET_DIR` with the blob-dump build so deps stay warm. Only `scripts/bench_recursion_cycles.sh` changes behavior; the workflow gets a comment update.

### DIFF
--- a/.github/workflows/bench-verify.yml
+++ b/.github/workflows/bench-verify.yml
@@ -36,9 +36,11 @@ jobs:
        startsWith(github.event.comment.body, '/bench-verify') &&
        contains(fromJSON('["MEMBER","OWNER","COLLABORATOR"]'), github.event.comment.author_association))
     runs-on: [self-hosted, bench]
-    # Job cap. On a cold runner the recursion BUILDS dominate: MEASURE_CLI once, plus
-    # per ref a guest build and a prover-test build. Cached in /tmp; build-std / host
-    # cargo target shared across ref worktrees (see the recursion step's env).
+    # Job cap. On a cold runner the recursion BUILDS dominate: per ref a guest build, a
+    # prover-test build, and a measuring-CLI build (the CLI is built FROM each ref so it
+    # understands that ref's own guest syscalls). Cached in /tmp; build-std / host cargo
+    # target shared across ref worktrees (see the recursion step's env), so the second
+    # ref's native builds mostly reuse the first ref's compiled deps.
     timeout-minutes: 90
     steps:
       - name: Acknowledge (react + occupancy notice)
--- a/scripts/bench_recursion_cycles.sh
+++ b/scripts/bench_recursion_cycles.sh
@@ -122,12 +122,19 @@ prune_worktree_cache() {
     git worktree remove --force "$wt" >/dev/null 2>&1 || rm -rf "$wt"
     rm -f "$WORK"/result_"${s8}"_*.txt "$WORK"/blob_"${s8}"_*.bin \
           "$WORK"/build_guest_"${s8}".log "$WORK"/dump_"${s8}"*.log \
-          "$WORK"/measure_"${s8}"*.err "$WORK"/measure_cli_"${s8}" \
+          "$WORK"/measure_"${s8}"*.err "$WORK"/measure_cli_"${s8}"* \
           "$WORK"/build_cli_"${s8}".log
   done <<< "$stale"
   git worktree prune >/dev/null 2>&1 || true
 }
 prune_worktree_cache
+
+# One-time sweep of the retired single-CLI scheme's fixed-name artifacts. Before this
+# script measured per ref it built one shared counter at $WORK/measure_cli (+ its .sha
+# marker and build_measure_cli.log). Those are never written or read anymore, and their
+# fixed names escape the per-SHA prune globs above, so on the long-lived bench runner
+# they would linger forever. Drop them so the disk-bounding claim actually holds.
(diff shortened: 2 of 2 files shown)


jotabulacios (approved):

MauroToscano (commented):

The measuring CLI is now built per-ref instead of once from head, mirroring `bench_verify.sh`, because a main-built executor cannot run a guest whose syscall it does not know. The mechanics check out (no stray `MEASURE_CLI`/`HEAD_SHA`, atomic `.tmp`+`mv -f` inside `$WORK`, no race across the four sequential presets, prune globs and result-cache key covering both new and retired names); two issues remain.

**scripts/bench_recursion_cycles.sh:354** — none of the new code has executed, and the green bench comment is not evidence that it has. That comment came from run `29846483185`, an `issue_comment` run whose checkout step has no `ref:`, so main's copy of the script ran: its log reads `==> Building MEASURE_CLI (cli, release) from 528a8411bb` / `==> Reusing cached MEASURE_CLI (528a8411bb)`, and its raw counts (`41726358`, `1470209700`) match the posted table exactly — the retired single-CLI scheme measured both sides. The only `workflow_dispatch` runs of `bench-verify` are three from the deleted `bench/recursion-bench-infra` on Jul 20, the newest also on the old scheme; none for `fix-bench-verify`. Benched SHA `05e2be02` predates head `09e68c33`, so the tree-grep commit is untested too. Step 2c2, the new cache key, the pre-#807 guard and the startup sweep have therefore never run. Blast radius is bounded (`continue-on-error` recursion step; every new failure path is a loud `exit 1` with a log tail, so a break reads as "regime unavailable" rather than a wrong number), but the workflow header prescribes a `workflow_dispatch` of `bench-verify` from `fix-bench-verify` for exactly this kind of change, with the log showing `building measuring CLI (cli, release) @ <sha8>` for both refs and a sane table. `scripts/bench_recursion_cycles.sh <#818-head-sha> origin/main min` on the bench box would additionally demonstrate the motivating `UnknownSyscall` case end to end.

**scripts/bench_recursion_cycles.sh:402** — syscall-adding PRs become measurable but not reportable. The parse keeps only `^Cycles:` and `^Keccak calls:`, so any extra counter a ref's own CLI prints is dropped from the table (486), the raw block (497-502), the cached record (414-419), and from the log too, since `$out` is echoed only on parse failure. #818, the PR this unblocks, adds `Fext load calls:` / `Fext fma calls:` / `Fext store calls:` in `bin/cli/src/main.rs` and moves Fp3 mul/fma behind those ecalls, each retiring as one guest cycle while doing its work in a new prover table — so #818's comment will show a guest-cycle drop with no number for the ecalls that absorbed the removed software work, precisely the blind spot the Keccak companion column exists to close (header, lines 20-22). Pre-existing parse, so follow-up-shaped rather than a regression, but cheap here: capture every counter-shaped line (`^[A-Za-z][A-Za-z ]* calls: [0-9]+$`) into the result record and the raw `<details>`, and flag when the two refs emit different counter sets — that asymmetry marks the cycle delta as "work moved", not "work eliminated". Skipping counters that are 0 on both sides preserves the deliberate Ecsm exclusion.

Nits:
- **scripts/bench_recursion_cycles.sh:369-383** — the `HOST_TARGET_DIR` if/else duplicates the `cargo build`, its 4-line error handler and the `cp` when only the target dir varies; `local host_target="${HOST_TARGET_DIR:-$wt/target}"` plus one `( cd "$wt" && CARGO_TARGET_DIR="$host_target" cargo build --release -p cli )`, one handler and one `cp "$host_target/release/cli" "$measure_cli.tmp"` halves it with identical behavior (`$wt/target` is cargo's default inside `$wt`) and kills the else-branch edge where a caller-exported `CARGO_TARGET_DIR` redirects the build while the `cp` still reads `$wt/target/release/cli`; the dump block at 333 has the same shape.
- The per-ref/`UnknownSyscall` rationale is stated in full at 25-35, 153-158 and 354-358 — keep it in the header and make the other two one-line pointers (step 1 still needs a tombstone; 2c2's unique content — HOST_TARGET_DIR sharing, copy-out, atomic tmp+mv, SHA-named cache — stays).

ColoCarletti (approved):

PR #858, opened 2026-07-23
perf: build PAGE trace from dense memory store

ColoCarletti:
## What
Generate the PAGE trace tables by reading each cell's final `(value, timestamp)` straight from the dense per-page memory store, instead of first building a sparse `FinalStateMap` and doing one (mostly-miss) lookup per page offset.

## Why
Those per-offset sparse-map lookups were the dominant cost of PAGE generation.
Reading the dense page slice directly is one indexed read per offset, no hashing.

## Impact (ethrex 5-tx block, RTX 5090)
- PAGE table generation: 658ms → 260ms
- Trace generation: 2.63s → 2.25s (**−15%**)
- Total prove: 12.8s → 12.5s (−2.4%)

Produces the exact same PAGE trace — `prove` + `verify` pass.

## Changes (3 files)
- `prover/src/paged_mem.rs`: add `PagedMem::page_data()` (dense per-page slice).
- `prover/src/tables/page.rs`: add `generate_page_trace_from_dense()`.
- `prover/src/tables/trace_builder.rs`: `generate_page_tables` uses the dense path.

### DIFF
--- a/prover/src/tables/page.rs
+++ b/prover/src/tables/page.rs
@@ -284,6 +284,64 @@ pub fn generate_page_trace(
     trace
 }
 
+/// Like [`generate_page_trace`] but reads each offset's final `(value, timestamp)`
+/// straight from the dense per-page memory store ([`crate::paged_mem::PagedMem::page_data`])
+/// — one indexed read per offset, no hashing. Equivalent to looking each byte up in a
+/// `FinalStateMap` built from the same cells, but avoids the sparse map and its
+/// `page_size` (mostly-miss) lookups per page — the dominant cost of PAGE generation.
+///
+/// `final_page` is `None` for a page with no runtime cells (every offset falls back to
+/// init, ts 0). `exclude_touched` drops runtime-written cells (ts > 0) so PAGE
+/// self-cancels them — the continuation-epoch case where the L2G table owns them.
+///
+/// `ts == 0` marks an offset that is either untouched or an initial-image byte; either
+/// way its final value equals its init value, so we emit `(init, 0)` (matching the
+/// `FinalStateMap`-miss branch of [`generate_page_trace`]).
+pub fn generate_page_trace_from_dense(
+    config: &PageConfig,
+    final_page: Option<&[(u8, u64)]>,
+    exclude_touched: bool,
+) -> TraceTable<GoldilocksField, GoldilocksExtension> {
+    let page_size = DEFAULT_PAGE_SIZE;
+    assert!(
+        config.page_base.is_multiple_of(page_size as u64),
+        "Page base must be page-aligned"
+    );
+    if let Some(page) = final_page {
+        debug_assert_eq!(page.len(), page_size, "dense page slice must span the page");
+    }
+
+    let num_rows = page_size;
+    let mut trace = TraceTable::new_main(
+        crate::tables::types::zeroed_fe_vec(num_rows * cols::NUM_COLUMNS),
+        cols::NUM_COLUMNS,
+        1,
+    );
+    let table = &mut trace.main_table;
+
+    for offset in 0..page_size {
+        table.set_u64(offset, cols::OFFSET, offset as u64);
+
+        let init_value = config
+            .init_values
+            .as_ref()
+            .and_then(|v| v.get(offset).copied())
+            .unwrap_or(0);
+        table.set_byte(offset, cols::INIT, init_value);
+
+        let (value, timestamp) = final_page.map_or((0u8, 0u64), |p| p[offset]);
+        let (fini_value, timestamp) = if timestamp == 0 || (exclude_touched && timestamp > 0) {
+            (init_value, 0)
+        } else {
+            (value, timestamp)
+        };
+        table.set_byte(offset, cols::FINI, fini_value);
+        table.set_dword_wl(offset, cols::TIMESTAMP_LO, timestamp);
+    }
+
+    trace
+}
+
 // =========================================================================
 // Preprocessed commitment
 // =========================================================================
--- a/prover/src/tables/page.rs
+++ b/prover/src/tables/page.rs
@@ -295,6 +295,14 @@ pub fn generate_page_trace(
 /// (`collect_bitwise_from_page`), so the two cannot drift and the AreBytes bus
 /// stays balanced.
 pub fn page_final(init: u8, value: u8, timestamp: u64, exclude_touched: bool) -> (u8, u64) {
+    // The `ts == 0 → (init, 0)` collapse only matches the sparse path when every
+    // ts==0 cell holds its init byte: image bytes are seeded `(init, 0)` from the
+    // same image `init_values` is read from, and runtime writes carry `ts >= 4`.
+    // Pin it here so both call sites inherit the check; debug-only, free in release.
+    debug_assert!(
+        timestamp != 0 || value == init,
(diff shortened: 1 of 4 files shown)


jotabulacios (approved):

prover/src/tables/page.rs
```diff
@@ -284,6 +284,78 @@ pub fn generate_page_trace(
...
+    exclude_touched: bool,
+) -> TraceTable<GoldilocksField, GoldilocksExtension> {
+    let page_size = DEFAULT_PAGE_SIZE;
+    assert!(
+        config.page_base.is_multiple_of(page_size as u64),
+        "Page base must be page-aligned"
+    );
+    if let Some(page) = final_page {
+        debug_assert_eq!(page.len(), page_size, "dense page slice must span the page");
+    }
```
Oppen, line 338: These could be enforced at the type system level instead:
1. Make page not a slice but a reference to a `[(u8, u64); PAGE_SIZE]` array;
2. Make `page_base` refer to the right-shifted address, as pages are powers of 2.

prover/src/tables/page.rs
```diff
@@ -284,6 +284,78 @@ pub fn generate_page_trace(
...
+/// `final_page` is `None` for a page with no runtime cells (every offset falls back to
+/// init, ts 0). `exclude_touched` drops runtime-written cells (ts > 0) so PAGE
+/// self-cancels them — the continuation-epoch case where the L2G table owns them.
+///
+/// `ts == 0` marks an offset that is either untouched or an initial-image byte; either
+/// way its final value equals its init value, so we emit `(init, 0)` (matching the
+/// `FinalStateMap`-miss branch of [`generate_page_trace`]).
+pub fn generate_page_trace_from_dense(
+    config: &PageConfig,
+    final_page: Option<&[(u8, u64)]>,
```
Oppen, line 328: This is 43.75% wasted space. A pair of slices would be more efficient: `(&[u8], &[u64])`.

Oppen (approved):

prover/src/tables/page.rs
ColoCarletti: Fair point, but that (u8, u64) is the existing `PagedMem<MemoryCell>` layout, not something this PR adds. Making it SoA means changing `PagedMem`, which the whole walk hits on every access. Better as a follow-up, and it'd help the whole walk, not just page gen.

prover/src/tables/page.rs
ColoCarletti: Agreed, both are cleaner, but they need PagedMem/PageConfig changes that ripple through the whole walk. Keeping this PR focused and opening a follow-up.

MauroToscano (commented):

This PR makes the dense store the single PAGE path, with `page_final` as the sole source of FINI/TIMESTAMP and the ARE_BYTES multiplicities. The `PagedMem` invariants behind the binary search hold, both call sites derive bases from `page_bases()`, and nothing moves on the verifier/transcript/CUDA side. Two items before merge, plus nits.

**prover/src/tables/page.rs:297** — `page_final` collapses every `ts == 0` cell to `(init, 0)`, equivalent to the sparse path only under the unenforced invariant that every `ts == 0` cell in `memory_state.cells` holds exactly the init byte for its offset. It holds today (`MemoryState::from_image` seeds `(byte, 0)` from the same `initial_image` that `build_init_page_data`/`PageConfig` read, trace_builder.rs:104-110/4244; unset offsets read back as fill `(0,0)` against a zero init; every runtime write carries `op.timestamp = i*4 + 4 >= 4`, trace_builder.rs:347), but nothing checks it and the new test assumes it rather than probing it. Breakage is silent: per `bus_interactions` (page.rs:517-597) PAGE receives `memory[0, addr, 0, init]` and sends `memory[0, addr, ts, fini]`, so for an untouched cell with init `I`, stored `V`, `I != V`, the old path emitted INIT=I/FINI=V/TS=0 — two non-cancelling tuples, a Memory-bus imbalance that failed the proof — while the dense path emits INIT=I/FINI=I/TS=0, which self-cancels. The trace stays provable while PAGE's final memory disagrees with executor state and with the image carried across a continuation chain. Not a soundness break (FINI is committed main trace, bus-constrained only), but the only signal that caught divergence is gone; `debug_assert!(timestamp == 0 => value == init)` inside `page_final` restores it and is inherited by `collect_bitwise_from_page`.

**prover/src/tables/page.rs:237** — `generate_page_trace`, `FinalStateMap` (page.rs:101) and `FinalByteState` (page.rs:93) lose their last non-test callers here; only `prover/src/tests/page_tests.rs` still references them, yet all three stay `pub` in the public module tree (`tables/mod.rs:47`) and `generate_page_trace`'s doc still reads "Generates a PAGE trace table for a single page." with no hint it is retired (the dense fn, the differential test and both trace_builder comments do say which path is live). Mark it as the reference oracle for `generate_page_trace_dense_matches_sparse`, or better gate it and the two types behind `#[cfg(test)]` — feasible since `lib.rs:27-28` declares `#[cfg(test)] pub mod tests;`. Retargeting page_tests.rs:33/65/96 at the dense generator is polish, not a coverage gap.

Nits:

- **prover/src/tests/page_tests.rs:262** — the `#[cfg(not(feature = "disk-spill"))]` gate rests on a false rationale and can be dropped: table.rs:268-269 provides `#[cfg(feature = "disk-spill")] impl<F: IsField> PartialEq for Table<F>` (element-wise via `.get()`, unbounded in `F`), while the `not(disk-spill)` derive adds an implicit `F: PartialEq` that `GoldilocksField` satisfies. The second sentence is correct and should stay: `TraceTable`'s derive (trace.rs:22) adds `E: PartialEq`, which `Degree3GoldilocksExtensionField` (extensions_goldilocks.rs:281) lacks. No CI gain — the disk-spill job is name-filtered (pr_main.yaml:233).
- **prover/src/tables/trace_builder.rs:2131** — comment says "`page_final` computes `(init, fini)`", but `page_final(init, value, timestamp, exclude_touched) -> (u8, u64)` takes `init` as input and returns `(fini_value, fini_ts)`, with call site 2144 dropping the ts half and `init` computed at 2141; reword to say it derives `(fini, ts)` exactly as `generate_page_trace_from_dense` does, which is the real justification for ARE_BYTES matching PAGE's FINI column.
- **prover/src/tables/page.rs:350** — the `None` arm (`final_page.map_or((0u8, 0u64), |p| p[offset])`) is unreachable: both callers (trace_builder.rs:2637, page_tests.rs:292) pass `Some`, since bases come from `page_bases()`, exactly what `page_data()` searches. One comparison in the existing loop against `generate_page_trace(&config, &FinalStateMap::new())` covers the `(0u8, 0u64)` constant and passes as written; note the fallback is duplicated at page.rs:350 and trace_builder.rs:2143.
- **prover/src/paged_mem.rs:110** — `page_data` alone does not normalize its address (`get`/`set` go through `Self::split(addr)`; this binary-searches the raw `base`), so an unaligned argument can never match and returns `Err(_) => None`, indistinguishable from "no set cell in this page". Unreachable today, but the unguarded consumer is `collect_bitwise_from_page`, where a spurious `None` skews ARE_BYTES multiplicities and unbalances the bus — add `debug_assert!(base.is_multiple_of(DEFAULT_PAGE_SIZE as u64))` rather than normalizing via `Self::split(base).0`, which would hide the misuse.

PR #863, opened 2026-07-24
perf(prover): halve GPU continuation proving time 

ColoCarletti:
## Result

ethrex 10tx `--continuations` on RTX 5090 + 16 cores, interleaved A/B/B/A (12 pairs): **main 22.58s → 11.05s = 2.04x (−51.1%, sd 0.88%)**. Proofs cross-verify in both directions against main's verifier.

## What changed

- **Epoch schedule**: execution/trace-build pipelined ahead of proving, K concurrent epoch provers, global prove overlapped with the epoch queue, trace builds on a builder pool.
- **Kernels**: constraint-interpreter rewrite (dim-split + liveness-reused slots): 3x kernel time, 8-35x less scratch; preprocessed tables committed through the fused GPU pipeline.
- **Device residency** (biggest win): the d=2 composition pipeline never crosses PCIe — H stays resident, on-device decompose + slab LDE, parts handle feeds R4 DEEP, FRI folds from the device-resident bit-reversed codeword. ~6 codeword-sized PCIe trips per table per epoch become one drain.
- **Shared caches**: domain/twiddles/OOD constants/FRI twiddles/zerofier inverses computed once per domain per process instead of per table per epoch; constraint-program lowering cached by content hash.
- **Fix**: intermittent deadlock — parallel batch inversion inside OnceLock initializers could starve against rayon workers parked on the same cell; initializers are now sequential and pre-filled at domain construction.

## Validation

Per-change interleaved ABBA + bidirectional cross-verify + full stark cuda suite (212/212) at every step; clippy clean; 20-prove stress run for the deadlock fix; GPU↔CPU parity bit-exact on the constraint kernels.

## Notes

- `--epoch-size-log2 22` + `LAMBDA_VM_EPOCH_CONCURRENCY=1` gives a further −18% on 32GB/16-core boxes (~27GB RAM per prove) — left at defaults, documented for deployment tuning.


MauroToscano (commented):

This batch adds an overlapped epoch pipeline for continuations, CUDA LDE/commit work, and a profiling toolkit. The test shape is right — the dim-split slot-file interpreter is differentially tested against the CPU model, the fused preprocessed-commit path has a CPU parity check, and `fini_from_final_state` pins the register fini to the trace — and the ABBA numbers back the perf claim. Two blocking issues, plus cleanup.

**prover/src/continuation.rs:1366** — the new epoch pipeline deadlocks instead of returning `Err` on any downstream failure. `rx` and `build_rx` are `Mutex<Receiver>` locals declared *outside* `std::thread::scope` (1130-1131), so they stay alive for the whole scope and a `send` can never observe disconnection; `tx.send(...).is_err()` (1237) and `build_tx.send(Ok(job)).is_err()` (1366) are therefore dead branches, contradicting their own "the downstream side hung up, stop quietly" comments.

The chain: a worker's `prove_epoch` `Err` (1174-1177, reachable from `Prover::multi_prove`) or a builder error forwarded as `tx.send(Err(e))` (1241-1244) sets `first_err`; every worker then exits at 1148-1150 and every builder at 1196-1198 after at most one more message. Nothing drains `build_rx` after that, and the producer never consults `first_err` — its only exits are `is_final` and the dead `send` check — so once the ~3 jobs of drain slack are consumed it blocks forever in `build_tx.send`, `producer.join()` (1442) never returns, the scope cannot exit, and the overlapped global thread stays wedged in `boundary_rx.recv()` (1399) because `boundary_tx` was moved into the stuck producer. Net: a failure with ~3+ epochs still pending turns a clean `Err` into a permanent hang of the main continuation entry point.

Scope limits: producer-originated errors (e.g. `HaltInNonFinalEpoch` from `Traces::collect_epoch`) shut down fine because the producer drops `build_tx`; a failure on or near the final epoch also returns normally; a single panicking prover with `workers >= 2` still completes.

Fix: check `first_err` at the top of the producer loop and return early (draining in the workers alone unblocks the scope but makes the producer execute the whole remaining program first), and have workers/builders keep draining to close rather than returning on the error path. This path has zero coverage — the only `Err`-expecting continuation test returns before the pipeline is built, and `trace_builder_tests.rs:1009` calls `Traces::from_image_and_logs` directly — so a negative test is needed: `cfg(test)` fault injection in `prove_epoch` or `build_from_collected`, enough epochs to exceed the slack, asserting `prove_continuation` returns `Err`, wrapped in a timeout so a regression fails instead of wedging CI.

**crypto/math-cuda/src/device.rs:632** — `PendingD2H` holds the staging slot's `MutexGuard` but has no `Drop`, so every `?` between enqueue and wait releases the slot with `cuMemcpyDtoHAsync` still writing into the pinned slab. Three sites do exactly that: `coset_lde_row_major_inner` (pending at 592-602, then `launch_row_to_col_major(...)?` at 606 which allocates, plus `take_event()?`/`record(..)?` at 609-610, wait only at 616), `coset_lde_row_major_split_trees` (759 → 764 → 770), and `coset_lde_batch_base_into_with_merkle_tree_inner` (1541, held across `d2h_bytes_via_pinned_hashes(...)?` at 1543, whose nested `ensure_capacity` can fail on pinned-host OOM).

Slots are per-rayon-worker (529-536), so the next user is usually the same worker on a later call — on a different stream, since `next_stream` round-robins, so nothing orders them — and slot 0 is shared by every non-rayon thread. That next user either grows the slot (`ensure_capacity` `cuMemFreeHost`s the old pointer at 63-69, and the in-flight DMA writes into a freed mapping) or repacks the slab for its own H2D (`as_mut_slice` + `copy_from_slice`, lde.rs:1170-1180) and the stale DMA silently corrupts that upload. These `Err`s are meant to be survivable — `gpu_lde.rs` falls back to CPU on `Err` at 467-471, 1042, 1114, 1656, and the crate ships a `test-cuda-faults` injector — so a recoverable OOM leaves the process running with a live DMA into an unowned slab. `?` does not poison the mutex; only the panic path is accidentally safe. Fix: `impl Drop for PendingD2H` doing a best-effort `self.staging.sync_event()` (ignoring the error) before the guard releases, so the invariant holds by construction. That also covers the lower-risk holds in deep.rs:249, fri.rs:223, logup.rs:205 and constraint_interp.rs:140, where the intervening `?` happens to be a sync/`memcpy_dtoh` that drains the copy on success.

**prover/src/continuation.rs:1120 / :1433** — `LAMBDA_VM_EPOCH_CONCURRENCY` defaults to 3 at 1120, but the sizing comment at 1433 says "the default is a conservative 2". Both lines are new in this PR, so one is stale on arrival, and it is not only a comment: main proves epochs strictly sequentially, while the new pipeline holds up to ~7 epoch-sized working sets concurrently per the comment at 1103-1104 (1 collected + 2 building + 1 queued + 3 proving), on CPU-only builds too. Nothing sizes this from available memory — `unwrap_or(3)` is unconditional — and the per-prove VRAM admission control does not coordinate across concurrent proves: `crypto/stark/src/prover.rs:2548-2552` hands each prove `backend().vram_budget_bytes()`, which is 80% of *total* device memory (`crypto/math-cuda/src/device.rs:252-275`), so all three plan `plan_table_chunks` against the whole card. 24 GB cards look safe (the measured ~15 GB peak is already the 3-worker config; the standalone ~9 GB per-epoch peak does not stack on top of it), but ≤16 GB cards, larger `EPOCH_SIZE_LOG2`, and CPU host RAM are exposed. Pick a number and justify it in one place — default 2 with 3 opt-in, or derive it from the detected budget — and delete whichever comment loses.

**scripts/profiling/README.md:149** — four doc sites promise math-cuda entry-point NVTX ranges that do not exist. Nothing under `crypto/math-cuda/src/` references `nvtx` except `lib.rs:18` and `nvtx.rs` itself; the ~40 `crate::nvtx::Range::fmt(...)` guards present at a48f5d42 (e.g. `lde.rs` `lde_tree_base[n=.. m=.. bf=..]`, `barycentric.rs` `bary_base[..]`) were dropped when 530f423f rewrote those files. They are still claimed at README.md:149 ("every `math-cuda` public entry point pushes a range with its shape"), `crypto/math-cuda/Cargo.toml:40`, `crypto/stark/Cargo.toml:59`, and `crypto/math-cuda/src/nvtx.rs:2`. The toolkit is not broken — instruments spans mirrored via `instruments.rs:89` and the `LAMBDA_VM_NSYS_CAPTURE_SPAN` gating at 92/112 both work, so the README's "verify ranges appear" check passes — but `nsys_phase_busy.py:318`'s "## math-cuda entry points (innermost NVTX range)" table computes `fine_key = base_name(chain[-1])`, which now resolves to the innermost *instruments* span, so that section is mislabeled and mostly duplicates the coarse table above it. Either restore the ranges behind the `nvtx` feature (zero-cost when disabled) or fix all four doc sites and drop that section from the report.

Nits:
- ~59 statements across `crypto/math-cuda` are wrapped in braces that bind nothing and scope no borrow (`let out = { stream.clone_dtoh(&out_dev)? };`, `{ stream.synchronize()?; }`, multi-line `let (a, b, c) = { ( ..?, ..?, ..? ) };`) where main has the plain form (`barycentric.rs:48`, `:75-77`); densest in barycentric.rs (22), merkle.rs (10), lde.rs (~9 inert plus 5 real borrow scopes), logup.rs (8), inverse.rs (5), ntt.rs and fri.rs (3), deep.rs (2), constraint_interp.rs (1) — semantically identical, but they read as if a guard were deleted, and `inverse.rs:71`, `logup.rs:202`, `deep.rs:246`, `deep.rs:399` still say "the labelled sync" with no label present; unwrap them (keeping lde.rs's five) or fill the holes with the `nvtx::Range` guards this PR already ships.
- `scripts/profiling/__pycache__/{nsys_phase_busy,nvml_sampler,phase_table,timeline_to_perfetto}.cpython-313.pyc` are checked in (~35 KB, first tracked `.pyc` files in the repo) with no `__pycache__`/`*.pyc` rule in any `.gitignore` — timestamp-invalidated so no stale-bytecode risk, but they persist in history and show as modified for any 3.13 dev importing these modules; `git rm -r --cached scripts/profiling/__pycache__` plus a root `.gitignore` entry.

MauroToscano (approved):

jotabulacios (approved):

PR #867, opened 2026-07-27
bench: default CI benches to ethrex 100tx continuations; widen rkyv pointers for >2 GiB proofs

ColoCarletti:
## What
- `/bench-gpu` and `/bench-abba` now default to ethrex 100tx `--continuations` (the production proving mode). New syntax: `/bench-gpu [N] [cont[TX]|mono[TX]]` — `mono5` keeps the legacy monolithic workload.
- rkyv now uses `pointer_width_64`: the default 32-bit offsets cap an archive at ~2 GiB, and the 100tx continuation proof is 3.13 GiB — proving worked, writing the file panicked. Proof size +2%, proving/verify time unchanged.

## Breaking
Proof files serialized before this change can't be read by new binaries (and viceversa). `RECURSION_INPUT_VERSION` bumped 1→2 so stale recursion blobs are rejected cleanly. No proofs are persisted in-repo or in CI.

## Validation (16-core + RTX 5090)
- 100tx cont: GPU prove 203 s, CPU 410 s, proof 3.13 GiB, verify OK (30.7 s)
- 10tx cont: 22.3 s (same as before the change), proof +2.04%

## Notes
- Comment-triggered workflows run from main, so this takes effect post-merge.
- PR branches older than this fix must rebase before a `cont` bench ≥40tx; the  config step now fails fast with a clear message instead of burning the build.

### DIFF
--- a/.github/workflows/bench-abba.yml
+++ b/.github/workflows/bench-abba.yml
@@ -2,8 +2,18 @@ name: Bench ABBA tiebreaker
 
 # Drift-free paired (A/B/B/A) prover benchmark for resolving small (~1%) deltas the
 # cheap PR benchmark can't confirm. It builds both binaries and runs ~20 interleaved
-# pairs, so it OCCUPIES THE SINGLE BENCH SERVER FOR ~30-40 MIN. For that reason it
+# pairs — at the default workload that OCCUPIES THE SINGLE BENCH SERVER FOR SEVERAL
+# HOURS (a 100tx continuations prove is ~7 min on a 16-core box). For that reason it
 # NEVER auto-triggers -- it runs only on an explicit `/bench-abba` comment on a PR.
+#
+# Syntax: "/bench-abba [N] [cont[TX]|mono[TX]]" (N = pair count, default 20). The
+# workload defaults to ethrex 100tx with --continuations — the production proving
+# mode at a size where per-epoch effects amortize (63 epochs). "cont10" gives the
+# small continuation workload (~45s/prove — good when a coarse answer fast matters
+# more than resolution); "mono[TX]", e.g. "mono5", the legacy monolithic prove. The
+# ~3.1 GiB cont100 proof requires rkyv pointer_width_64 on BOTH sides (see
+# prover/Cargo.toml), so PR branches older than that fix must rebase before a cont100
+# bench.
 on:
   issue_comment:
     types: [created]
@@ -26,9 +36,12 @@ jobs:
       startsWith(github.event.comment.body, '/bench-abba') &&
       contains(fromJSON('["MEMBER","OWNER","COLLABORATOR"]'), github.event.comment.author_association)
     runs-on: [self-hosted, bench]
-    # Generous ceiling so a hang/OOM can't strand the single bench runner; the
-    # workload itself is ~30-40 min at the default 20 pairs (clamped to <=40).
-    timeout-minutes: 120
+    # Generous ceiling so a hang/OOM can't strand the single bench runner — not the
+    # expected duration. Measured on a 16-core box: a 100tx-continuations CPU prove
+    # is ~7 min, so the default 20 pairs (40 proves) runs ~4.5-5 hr and the 40-pair
+    # clamp ceiling ~9.5 hr; builds add ~10-20 min. (The old 5tx-monolithic workload
+    # was ~30-40 min total at 20 pairs.)
+    timeout-minutes: 600
     steps:
       - name: Acknowledge (react + occupancy notice)
         uses: actions/github-script@v7
@@ -41,10 +54,10 @@ jobs:
             await github.rest.issues.createComment({
               owner: context.repo.owner, repo: context.repo.repo,
               issue_number: context.issue.number,
-              body: '⏳ **ABBA tiebreaker started** on the bench server (~30–40 min). The bench server is occupied until it finishes.'
+              body: '⏳ **ABBA tiebreaker started** on the bench server (~4-5 hr at the default 20 pairs of ethrex 100tx continuations; pass a smaller pair count or `cont10` for a quicker, coarser run). The bench server is occupied until it finishes.'
             });
 
-      - name: Resolve PR head + pair count
+      - name: Resolve PR head + bench config
         id: cfg
         env:
           GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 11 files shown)


Oppen (approved):

MauroToscano (commented):

This PR turns on rkyv's `pointer_width_64` across the proof-format crates, bumps `RECURSION_INPUT_VERSION`, and adds a CI pre-flight that blocks bench runs on branches predating the change. Both halves are internally consistent: the feature is on all six crates that declare `rkyv`, host and guest compile `RECURSION_INPUT_VERSION` from the same constant, every proof read stays on validated `rkyv::access`/`from_bytes` so old-format bytes yield an `Err` rather than a panic or unchecked access, the now-64-bit archived table dimensions remain gated by `checked_mul` plus the exact AIR-derived equality checks, and token parsing validates digits-only before every `[ -lt ]`. The findings below are workflow-side, plus one missing guard on the invariant the PR turns on.

**.github/workflows/bench-abba.yml:102-112** (duplicated at **benchmark-gpu.yml:135-146**) — the pre-flight does not fail open on a failed fetch, contrary to its comment ("Skip if the fetch fails"). `gh api` prints HTTP error bodies to stdout, so on any 404/403/5xx `SIDE` holds a JSON error blob rather than the empty string: `[ -n "$SIDE" ]` passes, `grep -q pointer_width_64` misses, and the job hard-exits with `PR branch predates the rkyv pointer_width_64 fix` without having read anything. Reproducible against this repo — `gh api "repos/$GITHUB_REPOSITORY/contents/prover/Cargo.toml?ref=<bad-sha>" -H "Accept: application/vnd.github.raw" 2>/dev/null || true` returns a `{"message":"No commit found for the ref ...","status":"404"}` body on stdout. A rate-limited token or a transient contents-API failure therefore turns the default cont100 bench into a hard block that blames the author for a stale branch. Fix: branch on gh's exit status (`if SIDE=$(gh api ...); then ... fi`) and/or require the payload to look like the file (e.g. it also contains `rkyv`) before treating a missing string as evidence the branch is stale. Secondary: the check is purely textual on one file's contents, so it also matches the feature name appearing only in the explanatory comment added above the dep; the whole check is deletable once every open branch postdates this merge.

**.github/workflows/bench-abba.yml:38** — the Acknowledge step still runs before the `cfg` step (52-124), the reverse of benchmark-gpu.yml, which this PR restructured so config (71) precedes ack (163) and the notice interpolates `steps.config.outputs.pairs`/`workload`. Two effects. The notice is a hardcoded literal, so `/bench-abba 4 cont10` still gets "~4-5 hr at the default 20 pairs of ethrex 100tx continuations; pass a smaller pair count or `cont10` for a quicker run" — hedged by "at the default" so not false, but uninformative; pre-existing, widened by the new workload axis. New here: the fail-fast at line 110 fires *after* the "bench server is occupied until it finishes" comment, which — unlike the GPU workflow's marker-based update-in-place — is a plain `createComment` that is never retracted, and it is the default path (`cont100` ⇒ `TX_COUNT >= 40`) for any un-rebased branch. The `always()` Post result step then posts a second comment headed ``` `<sha>` vs `main` ( pairs, ethrex)``` — `pairs`/`workload` are written at 118-123, after the `exit 1`, while `head_sha` lands at 63 before it — whose "Last log lines" block tails `/tmp/abba_out.txt` from a *previous* run, since the bench runner is persistent self-hosted and nothing cleans `/tmp`. Fix: move Acknowledge after `cfg` and interpolate `steps.cfg.outputs.pairs`/`workload`; the cfg step costs seconds, and the fail-fast then exits before any occupancy claim is posted. Optionally write the outputs block before the fail-fast (or `rm -f /tmp/abba_out.txt` at job start) so the failure comment is not half-blank with a stale log.

**prover/Cargo.toml:27-29** — the `pointer_width_64` invariant is held only by the "Keep in sync across all proof-format crates" comment plus five copies of it. Nothing in code or CI pins it: no const assert, no test, and `RECURSION_INPUT_VERSION` cannot help, since host and guest compile it from the same source (prover/src/lib.rs:214, checked at :265) — a rebuilt-but-misconfigured binary still advertises v2. rkyv `compile_error!`s on *conflicting* pointer-width features, but with none enabled it silently defaults to `u32`. Feature unification makes dropping it from a single main-workspace crate harmless; the two cases that bite are losing it from a whole graph (hoisting rkyv into `[workspace.dependencies]`, or a version bump that drops the feature list) and the separate non-member `bench_vs/lambda/recursion` guest workspace drifting from the host. Both stay green in CI, since every existing test round-trips archives that fit 32-bit offsets, and only resurface as a serialization failure at the end of a >2 GiB prove. Mirror the `const _: () = { assert!(...) }` pattern already at prover/src/lib.rs:216-229: `const _: () = assert!(size_of::<rkyv::primitive::ArchivedUsize>() == 8, "proof format requires rkyv pointer_width_64");`. That compiles into both the host and the riscv64 guest graph (the guest path-depends on lambda-vm-prover, whose rkyv dep is non-optional), so it covers the drift the CI grep cannot see; placing it in crypto/math or crypto/stark under their `rkyv` feature would additionally cover standalone builds of the crates that define the archived types.

Nits:

- **prover/src/lib.rs:257** — no test covers the prefix rejection path (`None` on a short blob / wrong magic / wrong version, surfaced as `Error::Execution("recursion blob: bad magic or version")` at :286, continuation twin at recursion.rs:312); the three tests in prover/src/tests/recursion_smoke_test.rs (600, 649, 711) all leave the 12-byte prefix intact. Pre-existing (only the constant's value changes) and bounded to a less legible bytecheck `Err`, but worth closing since the breaking-change note leans on clean rejection and blobs are persisted (`test_dump_recursion_input` writes `/tmp/recursion_input.bin`, prebuilt `recursion-*.elf` guests are cached); `recursion_archive_bytes` is pure, so a test needs no proving — fabricate a 12-byte prefix plus dummy bytes and assert `None` for (a) version patched to `1u32` LE at bytes 4..8, (b) a flipped magic byte, (c) a blob shorter than `RECURSION_INPUT_PREFIX_LEN`.
- **prover/src/lib.rs:256** — the doc comment claiming an `access_unchecked` follows is stale; there is none in tree.

jotabulacios (approved):

MauroToscano (approved):

PR #869, opened 2026-07-27
  perf: make bump the default guest allocator

jotabulacios:
The guest allocator now uses a monotonic bump allocator by default instead of `embedded-alloc`'s TLSF heap. The `dlmalloc-alloc` feature remains available for executions with unbounded allocation churn; it reuses freed allocations on top of a page-aligned bump provider. Bump allocation is constant-time, makes `dealloc` a no-op, and skips the `alloc_zeroed` memset because guest memory is zero-initialized. TLSF is removed from the guest allocator configuration.

## Motivation

TLSF spends guest instructions on free-list bookkeeping on every allocation, and a guest instruction is a trace row. Guest allocation patterns are dominated by short-lived buffers within one execution, so paying for reuse is paying for something the workload barely uses.

## What changed

- Default guest allocator is a monotonic bump allocator: `alloc` moves a cursor, `dealloc` is empty, `realloc` grows the top block in place when the block being grown is the one the cursor sits on, and `alloc_zeroed` skips the memset.
- `dlmalloc-alloc` puts Doug Lea's malloc on a bump "system" provider that hands it page-aligned segments. Its footprint is bounded by live bytes rather than total bytes ever allocated, and it can grow a buried block in place.
- `embedded-alloc` and the `tlsf-alloc` feature are gone, along with the guest lockfile churn they carried.

The `alloc_zeroed` memset skip depends on freshly bumped memory reading as zero. That holds at three independent layers: executor misses return `unwrap_or_default()`, the prover's replay fills `(0, 0)`, and page genesis is `zero_init` bound to the static zero-page root, which the verifier rebuilds from the ELF rather than trusting the prover. The cursor is monotonic by construction — `realloc` never rewinds it on shrink, which would re-expose written bytes and break exactly that invariant.

## Measurements

All figures are post-#861. Thin LTO inlines the per-allocation bookkeeping dlmalloc and TLSF pay and bump avoids by construction, which closed about two thirds of the gap the same comparison showed before it; any pre-LTO number for this change reads larger than it should.

Bump against dlmalloc, on the ethrex transfer fixtures and a real Hoodi block:

| Metric | Result |
|---|---|
| Guest cycles, 20-transfer block | -7.0% |
| Guest cycles, 150-transfer block | -6.2% |
| Guest cycles, real Hoodi block | -2.9% |
| Prove time, monolithic | -2.6% |
| Peak RSS, monolithic | -3.9% |
| Prove time, continuations at 2^21 and 2^22 | -1.2% |

The in-place `realloc` takes a further 0.57..1.34% off guest cycles across four ethrex fixtures; it has not been re-run against dlmalloc.

Where bump loses, both deterministic:

- Proof bundle is 0.6..1.0% larger at every epoch. Memory that is never reused spans more pages, and every page touched pays PAGE rows.
- At epoch 2^20 dlmalloc proves ~2.3% faster, in 8 of 9 paired rounds: eight epochs amplify that page cost. Larger epochs are ~26% cheaper in absolute terms, so the configuration anyone runs is the one bump wins.

Scope of the prove-time numbers: they come from the synthetic transfer fixtures on the monolithic path. **On the real block this change is not resolvable.** The bench box swings ~15% run to run on 8-minute runs; the CPU real-block run reported no significant change (-40 MB peak heap, -1.9% prove time, within spread), and the GPU ABBA run at n=32 came back INCONCLUSIVE with a point estimate of -0.01% and a 95% CI of [-0.24%, +0.22%]. The win is claimed on the fixtures, not there.

## Bump's ceiling

Bump never reclaims, so the quantity that matters is cumulative allocation, against ~3 GiB of `[_end, MAX_MEMORY_SIZE)`. Measured execute-only over eight ethrex fixtures spanning 0.42M to 63M gas:

| Gas | Cumulative alloc | Marginal B/gas |
|---|---|---|
| 2.43M (real, contract-heavy) | 9.4 MB | — |
| 4.24M (real, contract-heavy) | 15.1 MB | — |
| 31.5M (1500 transfers) | 72.1 MB | 2.211 |
| 63M (3000 transfers) | 142.0 MB | 2.219 |

`alloc = 2.55 MB + 2.213 B/gas × gas`, R² = 0.999993, with the marginal rate flat between 2.175 and 2.289 across a 150× range of gas. Allocation is linear: there is no superlinear term, which was the only route to exhaustion within one block.

The quantity that binds is bytes per gas, not transactions per gas, and transfers roughly minimise it. At the highest rate measured — 3.87 B/gas, on the contract-heavy blocks — a 60M-gas block lands at ~232 MB, a 13.9× margin, and exhaustion needs ~832M gas. Both contract-heavy fixtures are small blocks (2.4M and 4.2M gas), so that rate still carries the ~2.5 MB constant inside its average and no gas-full contract-heavy block has been measured: treat the margin as an extrapolation. If a block ever exceeds it, `dlmalloc-alloc` is a one-flag fallback.

A guest that processes many blocks in one execution has no per-block bound at all, which is the case `dlmalloc-alloc` exists for. Continuations do not stress it: they split proving into epochs, not execution.

## Known limitation, not fixed here

Heap exhaustion does not fail cleanly. `alloc` returns null, `handle_alloc_error` panics, and the guest `#[panic_handler]` is `loop {}`, so execution spins instead of aborting; nothing on the proving path bounds cycles, since `--cycle-budget` is opt-in and only on `execute`.

This is not shipped as part of this change, deliberately:

- It is preexisting. TLSF also returned null on exhaustion and hung through the same panic handler, and any guest panic hangs the prover the same way. The allocator choice moves the threshold, not the failure mode.
- The fix is not allocator-local. `HALT` constrains the exit code to zero at the constraint level (`prover/src/tables/halt.rs:16,194`, spec `halt:c:read_zero_exit_code`), so a nonzero exit is not a provable state. A clean abort needs either a committed failure marker or a non-provable abort ecall. `#[alloc_error_handler]` is not the lever: it is unstable, and the default handler already panics.

Two follow-ups: a host-side cycle bound on the `prove` path, which is cheap and removes most of the operational risk without touching the VM, and then the panic-to-abort design above.

## How to test

```
make lint
cd syscalls && cargo test && cargo test --features dlmalloc-alloc
make compile-programs-rust
```

`pr_main.yaml` covers both allocator profiles. To exercise the fallback in a guest, build it with `--features lambda-vm-syscalls/dlmalloc-alloc`; nothing in the repo selects it today, so the riscv64 path has no consumer yet.


nicole-graus (approved):

MauroToscano (dismissed):
Approving. The three doc issues raised in the earlier review pass are resolved, and the rewrite goes further than what was asked.

**Resolved:**
- *"Never came out behind on any deterministic metric on any workload"* is gone. The header now states the trade honestly — bump "pays for that with a 0.6..1.0% larger proof bundle at every epoch and a loss at epoch 2^20" — and says plainly that the real block does not resolve the difference.
- The stale gas figure is gone, along with the assertion it supported. `1500 transfers (31.5M gas) allocate 72.1 MB` is now presented as one measured fixture rather than as a gas-full block.
- `dlmalloc-alloc` now records that nothing selects it: "CI builds and tests the feature on host, but no guest manifest or Makefile rule turns it on, so the riscv64 `#[global_allocator]` below is a fallback with no consumer yet."

**Better than asked.** The earlier pass concluded the gas-linear fit gave a ~13.9x safety margin. The current text correctly declines that framing — "no gas rule bounds that, so the fit below describes honest blocks and is not a safety margin" — and then shows why, with the adversarial path: every CALL copies its argument region into a fresh, never-reclaimed heap buffer while memory expansion is charged once as `max(args, retdata)`, reaching ~561 B/gas, ~145x the honest contract-heavy rate. That is a stronger and more useful statement than the margin it replaces, and it reframes the operative limit as prover cost (PAGE rows per touched page, plus a per-page GLOBAL_MEMORY table on the continuation path that does not reset per epoch) rather than the ~3 GiB address range.

**Exhaustion behaviour is now documented precisely**, including that std guests never reach a panic handler at all — `__rust_alloc_error_handler` -> `unimp`, and execution spins. Shipping that as a documented known limitation is a reasonable call now that it is written down rather than implicit.

**Scope of this approval:** I verified the doc claims against the measurements and the code, and the earlier pass found no correctness bugs — the zero-init invariant that the `alloc_zeroed` memset skip depends on holds at three independent layers, and `realloc` now has a real override with a test. I did not re-review the work merged in from `main` since that pass (#914, #904, #909, #876 among others); each was reviewed on its own PR.

MauroToscano (approved):
Approving, on a review of the current head this time.

Context for the dismissal above: I had approved on the strength of a review pass against `3bb45e03` after reading only the allocator's doc comment. That head and this one differ by 62 files / +6596, and the gap included `b3699df1` "Grow the top bump block in place on realloc" — a change to the exact mechanism an earlier pass had flagged as its elevated risk. That approval wasn't earned, so it was withdrawn. This one replaces it.

**The zeroing invariant holds.** `alloc_zeroed` is `{ self.alloc(layout) }` — no memset — which is sound only if every byte handed out is already zero. The invariant that delivers it: every byte in `[HEAP_POS, HEAP_END)` has never been part of a returned block, and `HEAP_POS` never decreases. All three mutators preserve it — `alloc` returns a region ending at the new cursor, the in-place grow fires only when `ptr + layout.size() == HEAP_POS` so it claims only `[HEAP_POS_old, ptr+new_size)`, and shrink/`dealloc` never move the cursor.

Tested rather than argued: ~6M randomized ops across 16K/64K/256K heaps with realloc biased onto the top block, using a **served-byte bitmap** so a byte handed out twice is caught even if nothing ever writes it. `double_serves = 0`, no aliasing, no dirty `alloc_zeroed`, no lost realloc bytes, no cursor rewind. Counters: `inplace_grow=157806 copy_grow=57614 shrink_inplace=52790 shrink_copy=18490 zero_size_inplace_grow=7457`.

The zero-size case does fire — all 42,158 zero-size allocs land exactly at the cursor and 7,457 grow in place — and is safe: such a block owns no bytes, and the equality is self-invalidating, since whichever candidate grows first advances the cursor and knocks the other onto the copy path. It's also unreachable from Rust guest code, as `Global` short-circuits size-0 before `GlobalAlloc`. Concurrency is unreachable three ways: target spec `"singlethread": true`, no A extension (`-C passes=lower-atomic`), and no thread/clone/futex ecall.

**The recursion-guest churn question was adjudicated and refuted.** The structural half is real — the per-epoch loop does run inside the guest (`bench_vs/lambda/recursion/src/main.rs:97` → `verify_continuation_and_attest` → `verify_continuation_archived` → `verify_continuation_view`), and `Makefile:247` selects no `dlmalloc-alloc`. But the consequence doesn't follow: measured heap is ~5.02–5.16 MB/epoch at blowup4, exactly linear, so crossover against ~3 GiB is ~590 epochs against a real block's ~112. And the binding constraint sits upstream — `MAX_PRIVATE_INPUT_SIZE = 512 MiB` is enforced in `Executor::new` before any cycle runs, and clamped again guest-side in `get_private_input_slice`. Since heap and blob are both linear in epoch count, the ceiling is a closed form independent of N: ~143 MB at blowup4. A 112-epoch bundle (~3.9 GB) is rejected as a clean error, not a hang. `get_private_input{,_slice}` are a guest's only input channels, so there is no streaming path around the cap.

**CI coverage is sound.** `make test-syscalls` runs both profiles for the default arm; a dedicated job runs `dlmalloc-alloc` in both; and the init-guard test is profile-split, with a `#[cfg(not(debug_assertions))]` case for the release path where the `debug_assert!` is compiled out — which is the profile guests actually build in.

**Non-blocking follow-ups**, none of which I'd hold the merge for:
- The hyperfine trigger added here can't produce a before/after comparison: any `syscalls/**` change alters all four syscalls-linked bench ELFs, so they land in the `modified` arm, which benchmarks head only. This PR's own run shows every row at `Relative 1.00` with no base column — the allocator swap produced zero comparative data.
- The new in-place realloc path has no guest-side coverage; `allocator.elf` does one `alloc`, never a `realloc`.
- The PR body's "Continuations do not stress it" is true of the inner guest but doesn't address the outer recursion guest, whose execution does loop.
- The `~561 B/gas` adversarial figure in the module doc is untraceable — absent from the PR description, and levm is a pinned git dep. Every other header number has a table row behind it.
- `hint_min`/`hint_multi` lockfiles still list the removed `embedded-alloc` (merge-order artifact from #876; cosmetic, nothing uses `--locked`).

**Scope of this approval:** verified directly — the guest call chain, both private-input clamps, the CI wiring, the lockfile state, and the TLSF removal; plus the allocator stress and both test arms run locally. Not verified — benchmarks were not re-run, riscv64 × `dlmalloc-alloc` is compiled nowhere so that arm's guest build is argued rather than proven, and the prosecutor side of the churn pair never reported, so that finding rests on the defense's measurements plus my own check of the call chain and the cap.

PR #879, opened 2026-07-29
Perf/ecsm affine selector

jotabulacios:
## What
Adds an affine variant of the ECSM ecall (`ECSM_AFFINE_SYSCALL_NUMBER = u64::MAX - 11`): input
`xG‖yG` (64 B), output `xR‖yR` (64 B), against the x-only variant's 32 B in/out. An `IS_AFFINE`
selector column lets one prover serve both ABIs, so existing x-only consumers (`ecsm.elf`, bench,
ASM `test_ecsm*`) keep working unchanged.

With x-only output the guest could not learn `y(k·P)`, so it recovered it from a second query
`x((k+1)·P)` plus the chord-addition law — two accelerator calls and a batch of field arithmetic
per point. ethrex's ecrecover now takes `y` from the chip: two ECSM queries per 2-term linear
combination instead of four, and `solve_y` is gone.

## Details
- **The ecall** (`executor/src/vm/instruction/execution.rs`): `ECSM_AFFINE_SYSCALL_NUMBER` gets
  its own handler branch (`scalar_mul_xy_with_y`). Both operand buffers are 64 B, so the
  address-limb guard spans offset 63 for `xG`/`xR` (31 for the scalar) and the `xG`/`k`
  disjointness guard widens to 64 B. Validates `0 < k < N`, `xG < p`, `yG < p` and that
  `(xG, yG)` is on the curve.
- **`IS_AFFINE` selector** (`prover/src/tables/ecsm.rs`, 667 → 684 columns, 413 → 423
  constraints): the `yG`-read (4 dwords at `addr_xG + 32`, ts) and `yR`-write (4 dwords at
  `addr_xR + 32`, ts+3) MEMW buses fire with `mult = IS_AFFINE`, so they are inert on x-only rows
  and on padding. The x-only path is untouched: `yG` stays the canonical even lift and `yR` is
  witnessed only for ECDAS.
- **`yR < p`** (`OverflowKind::YrLtP`, +16 halfword columns and its carry chain): publishing `yR`
  is what makes its representation observable, and the byte range checks only bound it below
  `2^256`. `xR` was already forced canonical by `XrLtP`; this gives the other half of the output
  the same treatment.
- **Guest side**: `ecsm_mul_affine` in `syscalls/src/syscalls.rs`, and `lincomb2_with_oracle` in
  `crypto/ethrex-crypto/src/lib.rs` now does two oracle queries instead of four and drops
  `solve_y` entirely. It still needs one field inverse for the final affine addition.
- **New test guest** `ecsm_affine`, plus a forged-`yR` rejection test.

## Measured (ethrex bench_20, bench server, baseline = the hint branch this sits on)
3 interleaved pairs, each side with its own prover and its own guest ELF, since this touches
both. All proofs verify.

| Metric | base | with affine | Delta |
|---|---:|---:|---:|
| Prove (median) | 17.217 s | 16.589 s | **−3.6%** |
| Peak heap | 33,345 MB | 31,808 MB | **−4.6%** |
| Guest cycles | 5,087,142 | 4,906,209 | −3.6% |
| ECSM calls | 80 | 40 | **−50%** |

| Prove, growth series | base | with affine | Delta |
|---|---:|---:|---:|
| 4 transfers | 9.458 s | 8.847 s | −6.5% |
| 8 | 10.892 s | 10.389 s | −4.6% |
| 12 | 13.267 s | 12.674 s | −4.5% |
| 16 | 15.962 s | 14.739 s | −7.7% |
| 20 | 17.450 s | 16.618 s | −4.8% |

Halving the accelerator calls is the structural part; prove time moves less because ECSM/ECDAS is
a fraction of the trace. An earlier measurement of this work reported −10.2%, on a baseline that
sat 14,466 cycles above `8 × 524,288` and therefore carried a ninth CPU chunk the affine side
dropped — a boundary artifact, not the change. Pairs are interleaved rather than run as
sequential blocks, because cross-session drift on that machine exceeds the effect being measured.

## Soundness
Returning `y` makes two things observable that were not:

- **Input parity.** The AIR constrains `yG` only through `yG² ≡ xG³ + b`, which holds for both
  `+yG` and `−yG`. On the x-only path that freedom is invisible (`x(k·P) = x(k·(−P))`), but once
  `yR` is published a witness could pick `−yG` and get a correctly-computed multiple of a
  *different* point — indistinguishable to both the AIR (an on-curve check passes either way) and
  the guest (it cannot know the parity of `k·P` without doing the multiplication it delegated).
  The `yG` MEMW read closes it by pinning the witnessed column to the caller's own buffer.
- **Output canonicality.** The quotient columns absorb a multiple of `p`, so without `YrLtP` a
  witness could publish `yR + p` whenever `yR < 2^256 − p` (~`2^32`), and such points are
  constructible: `3 | p−1` makes cubing 3-to-1, so a small target `y` has a cube-root preimage
  about a third of the time.

`IS_AFFINE` itself is pinned by the `Ecall` receiver, whose syscall word is
`xonly + IS_AFFINE·(affine − xonly)`; the CPU sends the real `a7`, so a row claiming the wrong
mode leaves the bus unbalanced. `IS_BIT(IS_AFFINE)` and `IS_AFFINE·(1 − µ) = 0` keep it a bit
that is zero on padding. The written `yR` inherits its byte range checks through the ECDAS bus,
which carries the coordinate byte by byte and range-checks its own columns.

Removing the reconstruction also retires an argument rather than adding one: the x-only path had
to establish that the wrong sign of `ya` was excluded by the scalar guards, which is what made
`k = 1` and `k = N−1` degenerate. With `y` supplied by the chip those scalars are ordinary —
secp256k1 has cofactor 1 and prime `N`, so `k·P ≠ O` for every `k ∈ (0, N)`.


nicole-graus (approved):

diegokingston (approved):

MauroToscano (changes requested):
Formally verify it, and make a spec for 3mi

PR #881, opened 2026-07-31
feat(cli): prover config file

ColoCarletti:
Adds `cli prove --config prover.toml` (also on verify, for blowup): a single TOML profile for the prover's tuning knobs, which today are spread across env vars and hardcoded constants. Proofs are unaffected, these knobs only shape how the work is scheduled and split.

- [prove] blowup and epoch_size_log2; 
- [tables] per-table row caps in log2 (previously compile-time constants); 
- [scheduler] table_parallelism and vram_budget_mb; 
- [gpu] thresholds, mempool and kill switches.

The file is a diff over the code defaults: every key is optional, machine-computed defaults (cores, VRAM) stay computed, and unknown keys are rejected so typos fail loudly.
- Precedence: explicit CLI flag > env var > file > code default, so existing scripts and env-based tooling keep working unchanged. When a set env var shadows a file key, the CLI prints a warning naming the variable, so a stale export can never silently ignore the file.
- Ships docs/prover.example.toml fully commented with the defaults annotated, plus a full-coverage parse fixture in tests.


diegokingston (approved):

bin/cli/src/main.rs
```diff
@@ -259,33 +273,75 @@ fn main() -> ExitCode {
...
+                }
+            };
+            file_cfg.warn_env_shadowing();
+            file_cfg.install_runtime_overrides();
+            let blowup = blowup.or(file_cfg.prove.blowup).unwrap_or(2);
+            let epoch_size_log2 = epoch_size_log2.or_else(|| {
+                file_cfg.prove.epoch_size_log2.map(|v| {
+                    parse_epoch_size_log2(&v.to_string()).unwrap_or_else(|e| {
+                        eprintln!("config [prove].epoch_size_log2: {e}");
+                        std::process::exit(2);
```
Oppen: Shouldn't this be an `ExitCode::FAILURE`?

Oppen (approved):

bin/cli/src/main.rs
ColoCarletti: Yes, fixed in: https://github.com/yetanotherco/lambda_vm/pull/881/changes/c2b621a6ed80d43697ce6f5587524cea1cb4e421

PR #884, opened 2026-07-31
fix(tooling): review fixups for the real ethrex block fixture

MauroToscano:
## What this became

Started as review fixups for #880; grew into the wiring that PR was missing. #880 built a validated real-block fixture but nothing benchmarked with it — every perf number still came from the synthetic 20-transfer block its own motivation section calls unrepresentative. This branch closes that gap and swaps the block for a measured, screened mainnet one. The branch includes a merge of current `main` (@ 9ccdaf2), so all numbers below are main-vintage.

## The block: mainnet 25368371

2,428,684 gas, 29 txs — selected via Dune sweeps against measured mainnet composition (baseline: 3.5M txs + a 6-month weekly profile). It is the **only** block in 90 days in the 1.6–2.6M gas band with 2 heavy txs, no whale tx (top tx 22.5% of gas), and a mainnet-like transfer mix (heavy-gas 44.3% vs mainnet ~39%, p50 tx gas 41,297 vs ~38k, tx/Mgas 11.9). Verified: executes statelessly to its post-state root on the guest's precompile surface (no KZG 0x0a — the no-KZG screen was live during the check), free of transient-contract dependence (predates the FWA lottery deployment), 1 EIP-7702 authorization.

| workload | cycles (main vintage) | keccak | ECSM | measured prove |
|---|---|---|---|---|
| `ethrex_bench_20` (synthetic, growth sweep only) | 8,734,622 | 411 | 80 | ~25 s (CPU bench runner) |
| **mainnet 25368371 (the `/bench` workload)** | **50,781,557** | **10,478** | **116** | **158.8 s CPU runner · 59.87 s RTX 5090** (`--epoch-size-log2 22`) |

Real/synthetic ≈ **5.8×** by cycles. Both prove times are measured (see "First measured runs" below), not estimated.

Measurement rules learned the hard way, recorded in `tooling/ethrex-block-converter/README.md`:
- **Cycles/gas is not constant** (26–38 across measured blocks) — candidates must be measured, never extrapolated from gas.
- **Cycle counts are guest-build-sensitive**: thin-LTO (#861) is worth −5.5% on this real block vs −2.3% on toy fixtures, and the guest's C objects (secp256k1) make counts clang-major-sensitive (the Makefile pins target flags, not the compiler). Numbers must be labelled with their ELF vintage and compiler; the pre-LTO figures (74.8M) are superseded and marked as such.
- ECSM = exactly 4 × signature recoveries (116 = 4×29), confirming the per-ecrecover cost model.

## Benchmark tiers

- **`/bench`** — **proves the real block, and nothing else**: 3 sampled runs vs the cached 3-run baseline published on push to main (`/bench N` adjusts, 1–5; missing baseline renders the PR side alone with a notice). Verdict bands are honest about resolution: 3–10% deltas report as *unresolved*. Measured job time ~12 min. The synthetic sampled screen was removed deliberately: it was the only monolithic-path coverage (a path considered vestigial) and its EC-dense mix (8.83 ECSM/Mcycle vs 1.3–1.6 on real blocks) is not a workload that will ever be real. `benchmark-pr.yml` records which block it measured and refuses to diff across a repoint.
- **`/bench-gpu`** — the real block (golden target), epoch 2^22 per calibration (28.9% VRAM headroom on a 32 GiB card; 2^23 needs ~44 GiB). A/B/B/A paired runs on a rented box — absolute times are per-host (the prover is partly host-CPU-bound); the deltas are what travel. Takes an optional pair count (`/bench-gpu [N]`) and nothing else.
- **`/bench-growth`**, **`/bench-abba`**, **`/bench-verify`** — unchanged specialist commands (the growth sweep still uses the synthetic fixture family — a scaling curve needs a family of comparable blocks, which real blocks can't provide).
- `bench_verify.sh` / `perf_diff.sh` / `bench_abba.sh`: `WORKLOAD=synthetic|real`, defaulting to synthetic. `bench_verify.sh` composes with #878's monolithic/continuation split: `WORKLOAD` picks the block, #878's knobs pick how it's proven; under `WORKLOAD=real` the monolithic arm is skipped (a real block monolithically is ~240 GB of heap — it can only OOM).

Continuations are mandatory for real blocks, not tuning: monolithic peak heap grows ~3–5 GB/Mcycle (measured, R²≈0.998). `prove --continuations` also now prints `Peak heap` (it printed nothing; the benchmark parser would have hard-failed).

## Fixture mechanism

Fetched by URL + sha256 (the `prepare-sysroot` pattern: `.tmp` + verify + move, digest re-checked every invocation), from the [`bench-fixtures-v1` pre-release](https://github.com/yetanotherco/lambda_vm/releases/tag/bench-fixtures-v1) — both fixtures plus their ethrex-replay provenance caches, all digest-pinned. Regenerating the fixture from the hosted cache reproduces it byte-for-byte. No binaries in git. The converter (`tooling/ethrex-block-converter`, renamed from `ethrex-real-block` — it converts any block, the name now says so) moves off the per-PR path into a path-filtered workflow (`ethrex-block-converter.yml`); its ~335-package build, cache download and rev pin leave the required gate. `no_kzg_backend_linked` stays in the gate (microseconds, and it's the property the usability screen rests on). Converter pins stay on Hoodi — the one upstream-hosted cache — decoupled from the benchmarked block; repointing the benchmark is five Makefile lines + one test constant.

## Review fixups (original scope)

- Stale-cache bug: an `ETHREX_REPLAY_REV` bump silently kept converting the old cache (now guarded).
- CI double-build: k256 `expose-field` came only from a dev-dependency, so `cargo run` and `cargo test` resolved different feature sets and rebuilt the ethrex tree twice per job.
- Chain-config assert independent of the digest — it's what forced the deliberate `HOODI_CHAIN_ID → MAINNET_CHAIN_ID` flip during the repoint instead of letting a substituted config hide behind a regenerated hash.
- Measured doc corrections (`ethrex_bench_20` is 32,766 B in `distinct` mode, not 17 KB; rkyv note; "cheapest first" ordering).

## Known items

- `/bench-abba`'s escalation text points from a 20-tx monolithic result to a 100-tx continuations tiebreaker — pre-existing mismatch, flagged not fixed.
- `tooling/ethrex-tests/Cargo.lock` gains one line (`ecsm`/`num-integer`): pre-existing staleness; `cargo build --locked` in that workspace currently fails on main. Separable on request.
- The two alternate candidate blocks keep their pre-LTO measurements, in a separate table explicitly labelled with vintage, pending re-measurement.
- Guest builds inherit whatever `clang` is on PATH (moves cycle counts ~0.1–2% between clang majors); a ready pin lives on the unpushed-to-PR branch `fix/pin-guest-clang` — optional hygiene, the bench runner builds its own ELF so bench numbers never depended on it.
- Epoch-size defaults are encoded from measured calibration: GPU flow 2^22, bench runner 2^22, 128 GiB-class servers 2^23 (tables in the converter README); the CLI's laptop default stays 2^20.

## First measured runs (branch dispatches, pre-merge)

**CPU bench runner** ([run](https://github.com/yetanotherco/lambda_vm/actions/runs/30669692578)): real block, 3 sampled runs at epoch 2^22 — **median 158.8 s, spread 2.8%** (156.2/158.8/160.7), peak heap ~52 GB, total job ~12 min including the growth sweep. Two findings: the runner proves the real block at **3.13 s/M cycles vs 5.3–5.6 on synthetic** — real-world keccak/trie work is cheaper per cycle than the synthetic EC-heavy mix, so all prior time projections were pessimistic; and peak heap is machine-dependent (52 GB here vs 32 GiB on a 32-core calibration box at the same epoch — table parallelism scales with cores), leaving ~19% headroom on the runner. Stable across all three runs.

**GPU** ([4-pair A/B/B/A branch-vs-main dispatch](https://github.com/yetanotherco/lambda_vm/actions/runs/30669856290), rented RTX 5090, epoch 2^22): PR 117.9 s vs main 118.3 s — **−0.35%, 95% CI [−1.38%, +0.67%], verdict inconclusive-at-zero**, which is the pass condition for a tooling-only branch. Full chain validated (rental → provision → dual build → real-block ABBA → paired-t/Wilcoxon) in ~20 min. Note the per-run absolute (~118 s) vs the calibration box's 59.87 s on the same card class: the prover is partly host-CPU-bound, and rented hosts vary — absolute GPU times are per-box; only same-box A/B deltas are comparable, which is what this workflow measures.

Note: PR-comment triggers run the workflow definition from main (GitHub semantics), so these numbers came from `workflow_dispatch` on the branch; after merge, plain `/bench` comments exercise this flow directly, and the merge push publishes the first baseline.

## How to test
```
make ethrex-real-block-fixture   # fetch + digest-verify (~1 MB)
make test-ethrex-offline         # host-side suite, no network
```
Comment-triggered `/bench` exercises this workflow only after merge (comment triggers run main's workflow definition); pre-merge validation was done via `workflow_dispatch` on the branch — both runs linked above.


tooling/ethrex-real-block/Cargo.toml
```diff
@@ -24,16 +24,32 @@ ethrex-config = { git = "https://github.com/lambdaclass/ethrex.git", rev = "156c
...
+#
+# Used only by the tests, but declared here rather than in `[dev-dependencies]` on
+# purpose: it is the only thing that asks for k256's `expose-field`, and under
+# resolver 3 a dev-dependency's features are not unified into builds that exclude
+# dev-dependencies. As a dev-dep, `cargo run` (the `make ethrex-real-block-fixture`
+# step) and the `cargo test` that follows it in CI resolve two different k256
+# feature sets, so k256 and every ethrex crate above it compile a second time in
+# the same job. Confirm with:
+#   cargo tree -e features,no-dev -i k256   # must list expose-field
+#   cargo tree -e features -i k256          # must match the above
```
Oppen: Wouldn't it make more sense to just set the k256 expose-field feature here?

Oppen (approved):

tooling/ethrex-real-block/Cargo.toml
MauroToscano: Solved

PR #888, opened 2026-08-03
perf(prover): device-only preprocessed tables and GPU commits for mid-size tables

ColoCarletti:
- Preprocessed tables (BITWISE/DECODE) were excluded from the device-only gate, so every epoch drained their full LDE (and their aux) to host. The split-tree commit now honors `retain_host_lde`, R4 openings serve both subsets (multiplicity range and precomputed range) from the device row gather, and the gate no longer excludes preprocessed tables.
- Mid-size tables sat below the GPU-commit threshold, committed on CPU, and — having no device handle — re-uploaded their LDE on every R2-R4 dispatch (~40 GB per prove). The default threshold drops from 2^19 to 2^14, the sweep optimum: 2^14 beats 2^15..2^19, and also beats "everything on GPU", where sub-2^14 tables lose to launch overhead.

Also in this round, wall-neutral on this box (the scheduler already hides transfer time) but cutting per-prove PCIe from 187 GB to ~70 GB and freeing GPU busy for smaller-lane setups:

- R3 OOD barycentric kernels evaluate every OOD point in one pass over the LDE on a cols x chunks grid (the old kernels re-read the LDE per point from a single block per column).
- Domain coset points are cached on device instead of re-uploaded per table per epoch (~19 GB of identical data).
- The epoch builder thread pre-uploads the big main traces during its idle slack, so the R1 commit D2D-copies instead of paying the H2D in its chain. BITWISE is excluded (`prove_epoch` edits its multiplicities post-build) and `update_multiplicities` drops any stale pre-upload defensively.


diegokingston (approved):

diegokingston (commented):
This conflicts with 914

MauroToscano (approved):

PR #914, opened 2026-08-07
fix(gpu): survive transient VRAM pressure — recover resident-table declines, close an R2 corruption race

ColoCarletti:
Main today fails ~1/3 of real-block proves at epoch 2^22 on a 32 GB GPU (reproduced: 2/6 runs on 8b88a8d6), and under sustained VRAM pressure can silently emit an invalid proof (~1/15 runs). This PR fixes the three failure modes without touching the perf envelope.

**Hard failures under transient CUDA OOM.** The resident-aux LDE and the device R2 path had no fallback for a device-only table: the aux decline was a hard prove error ("resident aux LDE failed; host aux trace is empty") and an R2 device miss was a hard abort that kills the prover thread. Now the aux decline drains the device and retries once (the transient peak is usually gone, so the table stays resident); if it still declines, the resident aux trace and the main LDE are downloaded and the table continues host-backed. The R2 miss recovers the same way (materialize the resident LDEs, continue on the host evaluator). Both paths log the downgrade with the table name.

**Silent invalid proofs.** Concurrent device R2 windows under VRAM pressure can transiently produce a fully wrong composition H for one or two tables while every resident input stays correct (rerunning the same chain matches the host recompute), yielding a proof that fails the verifier's composition check. Serializing only the constraint-eval + decompose window across tables eliminates it (30/30 clean cycles vs ~5 failures/30 without); commits and host arms stay parallel and the windows overlap rarely enough that the lock is near-free (~0% on real block, ~1-2% on small synthetic workloads). LAMBDA_VM_GPU_SERIALIZE_R2=0 lifts the lock. The underlying race is still being hunted; the lock is the safe default until then.

**Device-only gate tightening.** The device R2 path only exists for the d=2 composition split; tables with any other bound (DECODE proves with num_parts == 1) now never enter device-only mode instead of aborting downstream. Mixed states (GPU aux commit with a CPU main commit) are also excluded.

**Diagnostics (env-gated, zero cost when off).** LAMBDA_VM_GPU_XCHECK=1 runs the verifier's composition consistency check inside the prover per table (~µs) and, on a failure, a post-mortem that recomputes each device stage on host, reports the corruption shape and reruns the device chain to distinguish a transient race from corrupted resident inputs. LAMBDA_VM_GPU_FORCE_DOWNGRADE=1 exercises the recovery paths end to end. These are how the fixes above were found and validated.

Validation on a 5090 (32 GB): 30/30 real-block e22 prove+verify cycles under reduced VRAM budget with zero corruption and all transient OOMs recovered (vs main failing 2/6 and the pre-fix branch failing ~5/30); full GPU test suite green (898 tests).


crypto/stark/src/gpu_lde.rs
```diff
@@ -54,6 +54,27 @@ fn gpu_lde_threshold() -> usize {
...
+    static LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+    if *ENABLED.get_or_init(|| !std::env::var("LAMBDA_VM_GPU_SERIALIZE_R2").is_ok_and(|v| v == "0"))
+    {
+        // The guarded state is (), so a panic while holding the lock carries
+        // no information — recover instead of burying the original panic
+        // under a cascade of PoisonErrors from every other table.
+        Some(LOCK.lock().unwrap_or_else(|e| e.into_inner()))
+    } else {
+        None
+    }
```
Oppen: Consulting the env var is not that expensive, and unless we set it inside the code it won't change during the lifetime of the binary:
```suggestion
    static LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
    if std::env::var("LAMBDA_VM_GPU_SERIALIZE_R2") != Ok("0") {
        // The guarded state is (), so a panic while holding the lock carries
        // no information — recover instead of burying the original panic
        // under a cascade of PoisonErrors from every other table.
        Some(LOCK.lock().unwrap_or_else(|e| e.into_inner()))
    } else {
        None
    }
```

diegokingston (approved):
We should check whether it makes sense to land 911 before

MauroToscano (approved):

PR #917, opened 2026-08-10
spec: HINT chip

nicole-graus:
## Description:

Specifies the hint `ECALL` implemented in #876, as `ECALL` number `-31`.

The chip constrains where the hinted value lands and that it is 32 bytes, but never the value itself — a prover may return anything. Soundness therefore rests on an obligation on the calling program: verify the hint, and recompute in software when the check fails. That obligation is declared as `HNT-A1` and discussed in the chapter.


erik-3milabs (changes requested):
see above

PR #931, opened 2026-08-11
spec: DMA memcpy accelerator

nicole-graus:
## Description

Specifies the `DMA` memcpy table added by #874. It has one row per eight copied bytes (one on the tail), rows chained over a new `DMA_NEXT` bus, `ECALL` number -3. Adds the `ADDNW` template (`ADD` without wraparound) used for the address updates, and lists -3 in the ECALL overview.

One deliberate deviation from #874's Rust: 31 columns rather than 32, because the spec models `timestamp` as a `Word`, as every other chapter does.


erik-3milabs (changes requested):
see above

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+
+#show: book-page("dma.typ")
+
+#let config = load_config()
+#let chip = load_chip("src/dma.toml", config)
+#let dma = raw(chip.name)
+
+The #dma chip copies a range of bytes from one location in memory to another, that is, it performs a `memcpy`.
+#footnote([Linux man-page on `memcpy`; man7.org. #link("https://man7.org/linux/man-pages/man3/memcpy.3.html")[[src]]])
+The guest performs such a copy with a RISC-V loop --- per copied doubleword a load, a store, two pointer increments and a branch, each of them a `CPU` row (@cpu) together with its memory operations.
```
erik-3milabs, line 21: Well, not any more. 😄 
Rephrase to indicate that this is how this would be done if this chip did NOT exist, i.e., if you were to have the CPU do this "manually"

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+This accelerator replaces all of that with a single row per eight copied bytes.
```
erik-3milabs, line 22: I'd recommend rewriting this in higher-level terms; after all, the reader does not immediately know how a CPU row compares to one of these rows.

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+An accelerated `memcpy` is expected to conform to the Ethereum Foundation's Accelerated Memory Operations standard, which fixes the semantics the symbol must keep and how it must win linking.
```
erik-3milabs, line 23: The standard is set, right? In that case, it either conforms or it does not. Either we can get rid of "expected", or the entire sentence can be removed.

erik-3milabs: Also, what does the following mean:

> ... the semantics the symbol must keep and how it must win linking

?

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+#footnote([Accelerated Memory Operations; eth-act/zkevm-standards. #link("https://github.com/eth-act/zkevm-standards/tree/main/standards/accelerated-memory-operations")[[src]]])
+Both obligations fall on the guest-side stub and on the link, outside this chapter; what the standard asks of the chip itself is that it assume no particular alignment of `dst`, `src` or `count`, which @dma:c:tail guarantees by choosing the width from `count` alone.
```
erik-3milabs, line 25: I would move the analysis of why this adheres to the eth spec to a separate subsection somewhere at the bottom; this is the chapter header and things should be as concrete/short as reasonably possible here.

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+#let nr_columns = total_nr_instantiated_columns(chip, config)
+#let nr_interactions = compute_nr_interactions(chip)
+
+The #dma chip is comprised of #nr_variables variables that are expressed using #nr_columns columns and leverages #nr_interactions interactions:
+#render_chip_variable_table(chip, config)
+
+= Assumptions
+#render_chip_assumptions(chip, config)
+
+The obligations on `src`, `dst` and `count` concern the _first_ row of a copy sequence only:
```
erik-3milabs, line 38: "obligations" => "assumptions"

"On the first row they are discharged by the register file, outside
this chapter."
Don't understand what that is supposed to mean. Consider rephrasing.

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+every subsequent row receives all three over the `DMA_NEXT` bus, whose sender range-checks them (@dma:c:range_src_incr, @dma:c:range_dst_incr and @dma:c:range_count_decr).
+On the first row they are discharged by the register file, outside this chapter.
+
+`timestamp` is different: it is forwarded verbatim by @dma:c:send_copy_next_chunk and never range-checked here.
+The first row's `timestamp` is pinned by @dma:c:receive_ecall to the `CPU`'s preprocessed timestamp column (@vars), and every later row inherits it through the bus.
+That column also carries $#`timestamp` = 4 dot (i + 1)$, which is what keeps $#`timestamp` + 2$ from leaving the `Word` range in @dma:c:write_value --- `IS_WORD` alone would not.
```
erik-3milabs, line 44: what does it mean for a column to carry a timestamp?

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+
+= Variables
+#let nr_variables = total_nr_variables(chip)
+#let nr_columns = total_nr_instantiated_columns(chip, config)
+#let nr_interactions = compute_nr_interactions(chip)
+
+The #dma chip is comprised of #nr_variables variables that are expressed using #nr_columns columns and leverages #nr_interactions interactions:
+#render_chip_variable_table(chip, config)
+
+= Assumptions
```
erik-3milabs, line 35: This section is bloated; please trim the excess detail.

spec/about_ecalls.typ
```diff
@@ -32,6 +32,7 @@ Negative numbers (represented as 2s complement 64-bit numbers), are used for our
 / 93: `exit` (@halt)
 / -1: `SHA256` (@sha256)
 / -2: `KECCAK` (@keccak)
+/ -3: `DMA`/`memcpy` (@dma)
```
erik-3milabs, line 35: The -1 through -10 range is (implicitly) reserved for future hash function accelerators. Please start a new "group", e.g., -30?

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+  render_constraint_table,
+  render_chip_assumptions,
+  render_chip_padding_table,
+)
+
+#show: book-page("dma.typ")
+
+#let config = load_config()
+#let chip = load_chip("src/dma.toml", config)
+#let dma = raw(chip.name)
```
erik-3milabs, line 17: Consider renaming the chip from `dma` to `memmove`; that more accurately describes what this chip actually does.

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+
+The #dma chip copies a range of bytes from one location in memory to another, that is, it performs a `memcpy`.
+#footnote([Linux man-page on `memcpy`; man7.org. #link("https://man7.org/linux/man-pages/man3/memcpy.3.html")[[src]]])
```
erik-3milabs, line 20: this chip does `memmove`, not (just) `memcpy` (which does not work for overlapping data ranges)

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+The `memcpy` operation has the following signature:
+
+```c
+void *memcpy(void dest[restrict count], const void src[restrict count], size_t count);
+```
+
+That is to say,
+- `A0` contains the address of the first byte to write,
+- `A1` contains the address of the first byte to read, and
+- `A2` contains `count`.
```
erik-3milabs, line 62:
```suggestion
- `A2` contains `count`; the number of bytes to copy.
```

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+
+@dma:c:read_dst, @dma:c:read_src and @dma:c:read_count read these three registers.
+Each of them writes back the value that was read, so the copy leaves the registers untouched;
+the guest is responsible for producing `memcpy`'s return value.
```
erik-3milabs, line 66: there is no return value? it is a void* function.

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+Again, these memory interactions only take place when this is the `first` call in the recursion tree.
+#render_constraint_table(chip, config, groups: "read_input")
+
+== Chunk width
+A row copies eight bytes whenever at least eight bytes remain, and a single byte otherwise.
+@dma:c:tail pins that choice to `LT` (@lt): the prover cannot select a convenient partition of the requested range.
+A copy of $n$ bytes is therefore laid out as $floor(n \/ 8)$ eight-byte rows, followed by $n mod 8$ one-byte rows, followed by one terminal row.
+
+Because a row is the unit in which this chip charges for a copy, an unbounded `count` would let a single guest instruction append an unbounded number of rows to the trace.
+@dma:c:bound therefore proves $#`count` < 257$ on the first row of every sequence, which caps a sequence at $39$ rows --- attained at $n = 255$, not at $n = 256$, since a byte short of the bound trades one eight-byte row for seven one-byte rows.
```
erik-3milabs, line 76: Why limit the chip to copying 256 bytes at a time? We have the `IS_HALF` and `IS_B20` range checks that can assert `count < 65536` or even `count < 1048576` at a smaller cost than calling the LT chip.

erik-3milabs (commented):
first batch of comments; I'll review the remainder on Monday

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+
+== Performing the copy
+The bytes are read from `src` at $#`timestamp` + 1$ and written to `dst` at $#`timestamp` + 2$.
+Both interactions are expressed over the _same_ `value` variable, which is what makes the copied bytes equal:
+there is nothing to constrain, since there is only one set of columns.
+The read carries `value` as both its input and its output, so `value` is pinned to whatever the memory argument (@memory) says resides at `src`.
+
+Splitting the two accesses over two consecutive timestamps is what gives an overlapping copy well-defined semantics:
+the memory argument orders accesses to an address by timestamp, so _every_ read of this `ECALL` observes memory as it was before the copy started.
+A single `ECALL` hence behaves like `memmove` rather than like a byte-by-byte forward copy.
```
erik-3milabs: bloated explanation: please rephrase to be more concise.

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+
+Which memory chip the two accesses reach is decided by their addresses, and this chip constrains neither.
+An eight-byte access lands on `MEMW_A`, the read-size aligned fast path (@memw), when its address is eight-byte aligned and all eight bytes were last touched at one timestamp: that chip stores a single old timestamp, so it needs one `LT` row (@lt) to order the access.
+Anything else falls to the general `MEMW`, which stores one old timestamp per byte and needs eight, on a row that is the wider of the two to begin with.
+The read and the write are routed independently, so a copy can take the fast path at one end and not at the other.
+And since @dma:c:tail picks the width from `count` alone, an end that begins misaligned stays misaligned for every chunk of the sequence --- this chip never emits a short leading row to realign one.
+A misaligned copy therefore commits strictly more cells than an aligned copy of the same length while producing exactly the same number of #dma rows, which is the reason the row counts this chapter reasons about are not on their own the cost of a copy.
```
erik-3milabs: bloated explanation; I don't need a summary here of the different types of MEMW chips here; I will look at the spec for that.

The reason I keep hammering on bloated explanations is that they often go stale / are not updated when a chip changes. Let less bloat, the more consistent the spec will be.

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+there is nothing to constrain, since there is only one set of columns.
+The read carries `value` as both its input and its output, so `value` is pinned to whatever the memory argument (@memory) says resides at `src`.
+
+Splitting the two accesses over two consecutive timestamps is what gives an overlapping copy well-defined semantics:
+the memory argument orders accesses to an address by timestamp, so _every_ read of this `ECALL` observes memory as it was before the copy started.
+A single `ECALL` hence behaves like `memmove` rather than like a byte-by-byte forward copy.
+
+#render_constraint_table(chip, config, groups: "copy")
+
+@dma:c:tail_lanes canonicalises a one-byte row.
```
erik-3milabs: no clue what canonicalises means in this context. What does it mean?

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+Such a row addresses `MEMW` with $#`write2` = #`write4` = #`write8` = 0$, i.e. it presents a single-byte access.
+`MEMW` gates every memory interaction for lane $i >= 1$ on those same width flags (@memw), so the seven unused lanes never reach the memory argument at all;
+what they do reach is the `MEMW` tuple itself, and pinning them to zero is what keeps that tuple the canonical encoding of a single-byte access rather than one carrying seven free field elements.
```
erik-3milabs: unnecessary explanation

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
++ A copy of zero bytes is a single row with $#`first` = #`end` = 1$: it accepts the `ECALL` and reads the three registers, but emits no memory operations and starts no recursion.
+
+== Chaining the rows
+When this was not the last chunk of this sequence, we recursively copy the next chunk over the `DMA_NEXT` bus, specifying the timestamp, the addresses to continue reading and writing at, and the number of bytes that still have to be copied (@dma:c:send_copy_next_chunk).
+Since that certainly won't be the `first` call in the sequence, we read `src_incr`, `dst_incr` and `count_decr` from the previous recursion level into `src`, `dst` and `count`, and continue copying.
+#render_constraint_table(chip, config, groups: "lookups")
+
+Both tuples carry the `timestamp`, and that is what separates one copy from another:
+without it, rows belonging to two different copies could be spliced into each other's sequences while the bus still balances.
+Since the CPU's timestamps strictly increase per instruction, no two #dma `ECALL`s share one.
```
erik-3milabs: superfluous

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+#render_constraint_table(chip, config, groups: "bits")
+
+= Padding
+To pad this chip, use the below data.
+#render_chip_padding_table(chip, config)
+
+Note that this padding row is not all-zero.
+@dma:c:count_decr is unconditional, so a padding row has to satisfy it too: $#`tail` = 1$ makes $#`step` = 1$, which $#`count` = 1$ and $#`count_decr` = 0$ then satisfy.
+The two address updates are conditioned on $#`μ` - #`end`$ and so do not forbid a wraparound here, but their low-limb carry is constrained on every row (@addnw:c:carry), so a padding row must satisfy that relation too; $#`src_incr` = #`dst_incr` = 1$ is the assignment that does so with a zero carry.
+It is not the only one --- @dma:c:range_src_incr and @dma:c:range_dst_incr, which would pin the limbs, carry multiplicity $#`μ`$ and are inert here --- but a padding row feeds no interaction either way.
```
erik-3milabs: We don't typically clarify why the padding works; we have the type check system to validate it for us.

@RobinJadoul, do we perhaps need to add to the type checker that padding rows don't send any interactions?

spec/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+
+= Notes/optimizations
+- The copy is a `memmove` per `ECALL`, but _not_ per guest-level `memcpy`: a copy larger than 256 bytes is chunked into several `ECALL`s at distinct timestamps, and chunk $k+1$ reads what chunk $k$ has already written.
+  This is in-contract for `memcpy`, whose buffers may not overlap, but the `memmove` property must not be claimed at the guest level.
```
erik-3milabs: This is a very important point; it should be somewhere earlier than here

spec/chapters/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+Again, these memory interactions only take place when this is the `first` call in the recursion tree.
+#render_constraint_table(chip, config, groups: "read_input")
+
+== Chunk width
+A row copies eight bytes whenever at least eight bytes remain, and a single byte otherwise.
+@dma:c:tail pins that choice to `LT` (@lt): the prover cannot select a convenient partition of the requested range.
+A copy of $n$ bytes is therefore laid out as $floor(n \/ 8)$ eight-byte rows, followed by $n mod 8$ one-byte rows, followed by one terminal row.
+
+Because a row is the unit in which this chip charges for a copy, an unbounded `count` would let a single guest instruction append an unbounded number of rows to the trace.
+@dma:c:bound therefore proves $#`count` < 257$ on the first row of every sequence, which caps a sequence at $39$ rows --- attained at $n = 255$, not at $n = 256$, since a byte short of the bound trades one eight-byte row for seven one-byte rows.
```
nicole-graus: You're right.
I measured the copy sizes on the mainnet block fixture (block 25368371) to pick between your two options, and I got 56,443 `memcpy` calls, all of them ≤ 24,504 bytes except one, which is 1,110,156 bytes (the private input being read in a single copy). 
So I'd go with `IS_HALF` (`count < 65536`), replacing the LT call. It gets all of the calls except one down to one ECALL.`IS_B20` wouldn't help with the one that matters since 1,110,156 exceeds 2^20 as well, so that copy stays chunked either way, while the worst case per call grows 16× (~131k rows instead of ~8.2k).
Tell me if it makes sense to you or if you would prefer `IS_B20`.

spec/chapters/about_ecalls.typ
```diff
@@ -32,6 +32,7 @@ Negative numbers (represented as 2s complement 64-bit numbers), are used for our
 / 93: `exit` (@halt)
 / -1: `SHA256` (@sha256)
 / -2: `KECCAK` (@keccak)
+/ -3: `DMA`/`memcpy` (@dma)
```
nicole-graus: Sure. I fixed it here [52a39f6](https://github.com/yetanotherco/lambda_vm/pull/931/commits/52a39f6b5b697f154d25f462e1c2766ead9842fe).

spec/chapters/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+#let nr_columns = total_nr_instantiated_columns(chip, config)
+#let nr_interactions = compute_nr_interactions(chip)
+
+The #dma chip is comprised of #nr_variables variables that are expressed using #nr_columns columns and leverages #nr_interactions interactions:
+#render_chip_variable_table(chip, config)
+
+= Assumptions
+#render_chip_assumptions(chip, config)
+
+The obligations on `src`, `dst` and `count` concern the _first_ row of a copy sequence only:
```
nicole-graus: - "obligations" => "assumptions": changed in [d0a39fe](https://github.com/yetanotherco/lambda_vm/pull/931/commits/d0a39fe90d53473fc9ff6a0c97612fbd60e31781).
- "On the first row they are discharged by the register file, outside this chapter." It means the register file already guarantees `src`, `dst` and `count` are valid `Word` limbs, so this chip does not re-check them. Rephrased in [d0a39fe](https://github.com/yetanotherco/lambda_vm/pull/931/commits/d0a39fe90d53473fc9ff6a0c97612fbd60e31781).

spec/chapters/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+every subsequent row receives all three over the `DMA_NEXT` bus, whose sender range-checks them (@dma:c:range_src_incr, @dma:c:range_dst_incr and @dma:c:range_count_decr).
+On the first row they are discharged by the register file, outside this chapter.
+
+`timestamp` is different: it is forwarded verbatim by @dma:c:send_copy_next_chunk and never range-checked here.
+The first row's `timestamp` is pinned by @dma:c:receive_ecall to the `CPU`'s preprocessed timestamp column (@vars), and every later row inherits it through the bus.
+That column also carries $#`timestamp` = 4 dot (i + 1)$, which is what keeps $#`timestamp` + 2$ from leaving the `Word` range in @dma:c:write_value --- `IS_WORD` alone would not.
```
nicole-graus: 'Carries' should have been 'holds'. It's the CPU's preprocessed timestamp column, whose row `i` holds `4*(i+1)`. Rewritten in [d0a39fe](https://github.com/yetanotherco/lambda_vm/pull/931/commits/d0a39fe90d53473fc9ff6a0c97612fbd60e31781).

spec/chapters/memmove.typ
```diff
@@ -0,0 +1,192 @@
...
+
+= Variables
+#let nr_variables = total_nr_variables(chip)
+#let nr_columns = total_nr_instantiated_columns(chip, config)
+#let nr_interactions = compute_nr_interactions(chip)
+
+The #dma chip is comprised of #nr_variables variables that are expressed using #nr_columns columns and leverages #nr_interactions interactions:
+#render_chip_variable_table(chip, config)
+
+= Assumptions
```
nicole-graus: Ok, done it here [d0a39fe](https://github.com/yetanotherco/lambda_vm/pull/931/commits/d0a39fe90d53473fc9ff6a0c97612fbd60e31781).

spec/chapters/memmove.typ
```diff
@@ -0,0 +1,192 @@
...
+```
+
+That is to say,
+- `A0` contains the address of the first byte to write,
+- `A1` contains the address of the first byte to read, and
+- `A2` contains `count`.
+
+@dma:c:read_dst, @dma:c:read_src and @dma:c:read_count read these three registers.
+Each of them writes back the value that was read, so the copy leaves the registers untouched;
+the guest is responsible for producing `memcpy`'s return value.
```
nicole-graus: Not sure I follow. Do you mean the part that says "the guest is responsible for producing `memcpy`'s return value"? .

spec/chapters/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+
+#show: book-page("dma.typ")
+
+#let config = load_config()
+#let chip = load_chip("src/dma.toml", config)
+#let dma = raw(chip.name)
+
+The #dma chip copies a range of bytes from one location in memory to another, that is, it performs a `memcpy`.
+#footnote([Linux man-page on `memcpy`; man7.org. #link("https://man7.org/linux/man-pages/man3/memcpy.3.html")[[src]]])
+The guest performs such a copy with a RISC-V loop --- per copied doubleword a load, a store, two pointer increments and a branch, each of them a `CPU` row (@cpu) together with its memory operations.
```
nicole-graus: You're right. Rephrased in [4efbb06](https://github.com/yetanotherco/lambda_vm/pull/931/commits/4efbb06a167b54ac27db18eafad7533e1c21b287).

spec/chapters/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+This accelerator replaces all of that with a single row per eight copied bytes.
```
nicole-graus: Yes. Fixed in [4efbb06](https://github.com/yetanotherco/lambda_vm/pull/931/commits/4efbb06a167b54ac27db18eafad7533e1c21b287).

spec/chapters/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+An accelerated `memcpy` is expected to conform to the Ethereum Foundation's Accelerated Memory Operations standard, which fixes the semantics the symbol must keep and how it must win linking.
```
nicole-graus: Both fixed in [4efbb06](https://github.com/yetanotherco/lambda_vm/pull/931/commits/4efbb06a167b54ac27db18eafad7533e1c21b287).

1. Got rid of "expected".
2. It meant two requirements, both about the guest side, not the chip: 
- The accelerated memcpy must behave identically to C's.
- The linker must pick it and not the C library's, which the standard gets by requiring a strong definition, or --whole-archive.

spec/chapters/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+#footnote([Accelerated Memory Operations; eth-act/zkevm-standards. #link("https://github.com/eth-act/zkevm-standards/tree/main/standards/accelerated-memory-operations")[[src]]])
+Both obligations fall on the guest-side stub and on the link, outside this chapter; what the standard asks of the chip itself is that it assume no particular alignment of `dst`, `src` or `count`, which @dma:c:tail guarantees by choosing the width from `count` alone.
```
nicole-graus: Sure. Moved it into a new section in [4efbb06](https://github.com/yetanotherco/lambda_vm/pull/931/commits/4efbb06a167b54ac27db18eafad7533e1c21b287).

spec/chapters/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+  render_constraint_table,
+  render_chip_assumptions,
+  render_chip_padding_table,
+)
+
+#show: book-page("dma.typ")
+
+#let config = load_config()
+#let chip = load_chip("src/dma.toml", config)
+#let dma = raw(chip.name)
```
nicole-graus: Agreed. Renamed in [d891b9e](https://github.com/yetanotherco/lambda_vm/pull/931/commits/d891b9e4b4e4a36f30501027474f2fc2dfeacaad).

spec/chapters/dma.typ
```diff
@@ -0,0 +1,192 @@
...
+
+The #dma chip copies a range of bytes from one location in memory to another, that is, it performs a `memcpy`.
+#footnote([Linux man-page on `memcpy`; man7.org. #link("https://man7.org/linux/man-pages/man3/memcpy.3.html")[[src]]])
```
nicole-graus: Fixed in [d891b9e](https://github.com/yetanotherco/lambda_vm/pull/931/commits/d891b9e4b4e4a36f30501027474f2fc2dfeacaad).

spec/chapters/dma.typ
```diff
@@ -16,11 +16,8 @@
...
-The guest performs such a copy with a RISC-V loop --- per copied doubleword a load, a store, two pointer increments and a branch, each of them a `CPU` row (@cpu) together with its memory operations.
-This accelerator replaces all of that with a single row per eight copied bytes.
-An accelerated `memcpy` is expected to conform to the Ethereum Foundation's Accelerated Memory Operations standard, which fixes the semantics the symbol must keep and how it must win linking.
-#footnote([Accelerated Memory Operations; eth-act/zkevm-standards. #link("https://github.com/eth-act/zkevm-standards/tree/main/standards/accelerated-memory-operations")[[src]]])
-Both obligations fall on the guest-side stub and on the link, outside this chapter; what the standard asks of the chip itself is that it assume no particular alignment of `dst`, `src` or `count`, which @dma:c:tail guarantees by choosing the width from `count` alone.
+Without this chip the guest would copy the range with a RISC-V loop --- per doubleword a load, a store, two pointer increments and a branch, five instructions the proof pays for.
+This accelerator replaces the loop with a single `ECALL` that copies eight bytes per row.
```
erik-3milabs: * _up to_ eight bytes per row

spec/chapters/dma.typ
erik-3milabs: > while the worst case per call grows 16×

What do you mean by this?

erik-3milabs: Note that memmove should never be chuncked; if the tail of the source memory and head of the destination memory overlap, you end up duplicating the first chunk of the overlapping section throughout the destination, rather than properly moving the contents.

Solutions: split `count `over 2 columns. Either 2x B20 (2 lookup, 2^40 range), or 1x B20 + 1x Bit (1 lookup, 2^21 range).

PR #932, opened 2026-08-11
spec: ECSM affine ecall variant

nicole-graus:
## Description

Specs the affine ECSM variant added in #879.

An `is_affine` selector lets one chip serve both ecalls: it picks the ECALL number the chip answers to, and gates the `yG` read (`addr_xG + 32`, at `ts`) and the `yR` write (`addr_xR + 32`, at `ts + 3`). Everything else is shared. Two checks come with it — the `yG` read pins the input point's parity, which becomes observable once `yR` is published, and `yR < p` forces the output canonical.

ECSM goes 37 → 42 variables, 708 → 757 columns.

**Numbering change:** the ecall number is now `-11 - 2·id - is_affine`, so secp256r1 moves from -12 to -13. #879 ships affine secp256k1 at -12, and secp256r1 has no implementation.

**Not spec'd:** #879's address-limb LT bounds. Those close a gap in `ecsm.rs`, which builds dword bases in the low limb only; the spec derives every address with a full 64-bit `ADD`.


erik-3milabs (changes requested):
see above

PR #943, opened 2026-08-21
spec: Recursion

erik-3milabs:


spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
...
+#let comm(x) = $commit(#x)$
+
+#let program = $f$
+#let relation = $cal(R)$
+
+#let verifierSpace = $cal(V)$
+#let (prove, verify) = ($italic("p")$, $italic("v")$)
+
+// Mathematical symbols
+#let (zero, one) = ($bb(0)$, $bb(1)$)
```
RobinJadoul: Do we need these to be bb? It already feels like a lot of bb throughout the document

spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
...
+witness space $witnessSpace$ to this set.
+We use $instanceSpace := functionSpace times inputSpace = {instance: witnessSpace mapsto BB}$ 
+to denote the set of functions with the public input "baked in"; 
+elements in this set are henceforth referred to as _function instances_, or simply _instances_.
+We then define $relation subset.eq instanceSpace$ 
+as the set of all _solvable instances_, 
+i.e., all instances $instance in instanceSpace$
+for which there exists a witness $witness in witnessSpace$ such that 
+$instance\(witness) = one$.
+Lastly, we introduce the commitment function $c: instanceSpace mapsto commitmentSpace$.
```
RobinJadoul: commitments may need to be randomized, so with an explicit randomness parameter maybe `c(x; r)`

spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
...
+To simplify notation, we use $commit(instance) = c(instance)$.
+
+We now assume the existence of _proving system_ $(prove, verify)$ with
```
RobinJadoul:
```suggestion
We now assume the existence of _proof system_ $(prove, verify)$ with 
```

spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
+#import "/book.typ": book-page, et, aside
+
+#show: book-page("recursion.typ")
+
+// Spaces and instances
+#let (functionSpace, function) = ($cal(F)$, $bb(f)$)
+#let (inputSpace, input) = ($II$, $bb(i)$)
+#let (instanceSpace, instance) = ($XX$, $bb(x)$)
+#let (witnessSpace, witness) = ($WW$, $bb(w)$)
+#let (proofSpace, proof) = ($bb(Pi)$, $bb(pi)$)
```
RobinJadoul: maybe just `pi` instead of bb?

spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
...
+To simplify notation, we use $commit(instance) = c(instance)$.
+
+We now assume the existence of _proving system_ $(prove, verify)$ with 
+prover $prove: instanceSpace times witnessSpace mapsto proofSpace$ and 
+verifier $verify: commitmentSpace times proofSpace mapsto BB$ such that
+$
+forall (instance, witness) in relation times witnessSpace 
+&: prob[verify\(commit(instance), prove\(instance; witness)) = one | instance(witness) = one] = 1 \
+forall instance in instanceSpace without relation,  forall proof in proofSpace 
+&: prob[verify\(commit(instance), proof) = one] < epsilon
```
RobinJadoul: what probability space is this over? For the prover I'm assuming internal randomness, but the verifier should presumably be deterministic

spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
...
+for a given instance $instance$ succeeds:
+$
+  &prove\(verify(comm(instance), dot); proof) = proof', text("and")
+  &verify(comm(verify(comm(instance), dot)), proof') = one.
+$
+This new proof $proof'$ thus attests to _the existence of a proof $proof$ that 
+satisfies the verifier on the given instance $instance$_.
+
+This concept, colloquially known as _proof recursion_, can be applied repeatedly.
+The technique is specifically beneficial for _succint_ proving systems where proof size
```
RobinJadoul:
```suggestion
The technique is specifically beneficial for _succinct_ proving systems where proof size
```

spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
...
+typically shrinks (and verification time therefore reduces) as the level of recursion increases.
```
RobinJadoul:
```suggestion
typically shrinks (and verification time reduces) as the level of recursion increases.
```

It is not a guaranteed fact that smaller proofs are faster to verify, though typically they would be

spec/recursion.typ
RobinJadoul: I think this may be a side-effect of formalizing things as NIZKs, since the common way to express this as an interactive proocol can rely on the random challenges from the verifier for both of these.
For a NIZK, the soundness becomes computationally secure, presumably, so it becomes dependent on the prover somewhere.

spec/recursion.typ
RobinJadoul: Or maybe we can be content with a deterministic commitment scheme and make explicit that we don't care about/expect it to be hiding.

spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
...
+
+A solution to this, is to leverage the following variation to the verification algorithm:
+$
+  verify': commitmentSpace^2 times {0, 1} times proofSpace: (commitment_0, commitment_1, b, proof) mapsto
+  cases(
+    verify(commitment_0, proof) &text("if") b=0,
+    verify(commitment_1(commitment_0, commitment_1, dot), proof) &text("if") b=1
+  )
+$
+where it is assumed that $comm(function(x_1, x_2, dot))$ can be easily
```
RobinJadoul: stick to just $f$ (or use function everywhere and make it regular instead of bb)

spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
...
+constructed from $comm(function), comm(x_1)$, and $comm(x_2)$.
```
RobinJadoul: Is this formulation the case in practice for us, and how does it interact with the absorption remark? That is, we let the verifier take an instance as argument, which has already absorbed its input, so that part of the formalization seems slightly wrong here. If we keep instance as `(f, i)` without the reinterpretation as `f(i; .)` it might work, but then the way we commit should also change a bit to have a commitment to `f` and potentially plain `i` (maybe a commitment to `i` works, but I'm not 100% certain).

spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
...
+The introduction of this split allows the verification algorithm to be split in two halves,
+with each VM performing the computations it is fastest at.
+The two halves cannot work independently, however.
+In the process of verifying proofs of the current proof system (`DEEP-FRI` + `LogUp`),
+results of binary arithmetic are used to verify field arithmetical constraints 
+--- e.g., field challenges extracted from binary hash outputs --- 
+and vice-versa --- e.g., hashing merkle leafs containing field elements during FRI-query proof verification.
+This implies that some form of communication between both VMs is required.
+
+This architecture enables the required communications by introducing a
```
RobinJadoul: "This" feels like it refers to what already was described, but the record is not an inherent part of introducing a split, so maybe

```suggestion
Our architecture enables the required communications by introducing a 
```

spec/recursion.typ
```diff
@@ -0,0 +1,224 @@
...
+]
+
+In theory, any division of tasks between the two VMs would work.
+Yet, it is expected that some division will be more performant than others.
+Below, we provide a division that, in theory, is expected to achieve solid performance:
+
+*Record $record$.*
+The record contains all challenges the prover derived using Fiat-Shamir.
+
+*Tasks for $verify'_b\(commitment_0, commitment_1, b, proof, record)$:*
```
RobinJadoul: This is missing a few of the ad-hoc checks, which are (I think) mostly related to memory and paging.
It would probably be a good idea to have an overview of those somewhere central as well.

spec/recursion.typ
erik-3milabs: > It already feels like a lot of bb throughout the document

I know! Once I stumbled across `bb` on Typst, I fell in love with it 🙈 Let's collaborate on reducing its use to an acceptable level 😉 

> Do we need these to be bb? 

Specifically `bb`? No. I did feel like it would be useful to distinguish these booleans false/true from normal 0/1, also because some programs use 1 as error state and 0 as success state; I didn't want there to be confusion on what 0/1 meant. I considered `bot` and `top`, but was not content with how they looked. 

Do you agree that distinguishing them from false/true is valuable? Do you perhaps know of a different useful symbol here?

spec/recursion.typ
erik-3milabs: I'm going to miss `bb(pi)` 🥹 
I really liked the symbol 😄 

You mean just the proof, or also the space, i.e.?
```suggestion
#let (proofSpace, proof) = ($bb(Pi)$, $pi$)
```
or 
```suggestion
#let (proofSpace, proof) = ($Pi$, $pi$)
```

spec/recursion.typ
erik-3milabs: Wdyt of the rephrase?

spec/recursion.typ
erik-3milabs: I'm not content with this section, at all. Indeed, stuff is missing.
I propose we have a full verifier description in `verifier.typ` (including all the ad-hoc checks you mention) and here only mention how to transform that monolithic verifier into the split/recursive verifier in this section.

wdyt of that split?

spec/recursion.typ
erik-3milabs: wdyt of simply rephrasing this in terms of completeness (= proof for solvable instances is accepted) and completeness (= any proof for unsolvable instances is rejected), without the mathematical formulae?

spec/recursion.typ
erik-3milabs: My view on this is as follows: 
- `f` is the program with any constants baked in. The commitment of `f` is the merkle commitment of the DECODE table.
- `i` are the program settings. The commitment of `i` is the merkle commit of the PAGE tables containing public memory.

So, I would say that `instance-comm( f(i; ·) ) ≡ (function-comm(f), input-comm(i))`

> Is this formulation [of commit composition] the case in practice for us?

I believe so, yes

> and how does it interact with the absorption remark?

Could you elaborate what you mean by this?

(I haven't refactored the text to work with split instances just yet; would like to reach understanding on the topic before performing the refactor)

spec/recursion.typ
RobinJadoul: I don't think it's strictly needed here to have the separation, as long as it's clearly defined what counts as a "successful" execution. But I'd say top/bot or `serif("success")`/`serif("fail")` or the like could probably work as well or better, if you don't want to just to with 0/1

spec/recursion.typ
RobinJadoul: Keeping the spaces bb is fine (I tend to go with cal, but bb(Pi) should be fine), I don't really like the lowercase one, and just pi for that seems like a fairly standard way to indicate a proof (or proof string)

spec/recursion.typ
RobinJadoul: Should work, I think

spec/recursion.typ
RobinJadoul: > So, I would say that instance-comm( f(i; ·) ) ≡ (function-comm(f), input-comm(i))

My surface reading without this clarification would be to read `f(i; ·)` as the function with now also `i` "baked in".
Maybe this needs some in-text clarification.

It can also help a bit to make the definition of `bb(X)` have a `≅` instead of plain `=` when reinterpreting it as purely functions taking a witness.

RobinJadoul: > Could you elaborate what you mean by this?

I mean kinda that I read `f(i; ·)` as absorbing `i` into `f` (as clarified in my previous comment). So you'd end up with some weirdness on how exactly the commitments work there, similar-ish to the circularity of the absorption remark.

spec/recursion.typ
RobinJadoul: Works for me

spec/recursion.typ
erik-3milabs: Refactored it to just 0/1.

spec/recursion.typ
erik-3milabs: done

spec/recursion.typ
erik-3milabs: I rewrote this section. Let me know what you think!

spec/recursion.typ
erik-3milabs: Let me know if the rephrasing works

spec/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+We thus have to demonstrate the commitments the prover provides are as expected.
+This is achieved by having the verification algorithm `COMMIT` (see @commit)
+to the public input it is provided.
+This act produces an imbalance in the LogUp-component of the proof-of-verification, 
+which must be balanced during verification in the _next_ recursion layer.
+In later recursions, the verifier must consistently `COMMIT` to its public input
+and use the _same_ public input to balance out the LogUp-component of the proof
+it is provided.
+This solution effectively kicks the can down the road; the final verifier has to
+provide the initial input to the program as input to verify the recursive proof.
```
RobinJadoul, line 225: 1. We may not have access to COMMIT in a future flock-fieldvm hybrid
2. Isn't this handled by $\mathbb{c}_1 = \bar{\mathbb{x}}$ already? Since the instance $\mathbb{x}$ includes the public input?

spec/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+
+*Prover:*
+$
+  proof &arrow.l prove(instance, witness)\
+  proof' &arrow.l prove(verify'_b || verify'_f, (commit(instance), (commit(verify'_b), commit(verify'_f)), 0, proof, record))\
+  proof^((i)) &arrow.l prove(verify'_b || verify'_f, (commit(instance), (commit(verify'_b), commit(verify'_f)), 1, proof^((i-1)), record))
+$
+
+*Final verification.*
+$verify'(commit(instance), (commit(verify'_b), commit(verify'_f)), 1, proof^((n))) =? one$
```
RobinJadoul, line 255:
```suggestion
$verify'(commit(instance), (commit(verify'_b), commit(verify'_f)), 1, proof^((n))) =^? one$
```

RobinJadoul, line 255: We may be able to get rid of the `b` argument by having the statement being proven the OR of both options:
I know a proof that is either accepted for the original statement OR for the recursion function.
I suppose this is essentially absorbing this bit into the proof or witness, but it cleans up the final verification and allows proofs without compressing to be used without any further change.

spec/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+#let (zero, one) = ($0$, $1$)
+#let iff = $arrow.double.l.r$
+#let implies = $arrow.double.r$
+#let prob = $PP$
+
+#show math.equation.where(block: false): box
+
+= Notation
+Let $BB := { zero, one }$ denote the boolean set and let 
+$functionSpace := {function: inputSpace times witnessSpace mapsto BB}$ denote
```
cdesaintguilhem: I always thought `mapsto` was for defining the actual mapping, and `to` was for the function signature; my mistake?

spec/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+
+= Resolving growing instance complexity
+While recursive proving leads to a decrease in proof size, this is naively traded off
+against an increase in instance complexity.
+Looking at a depth-two recursive proof,
+$
+  &prove\(verify(comm(verify(comm(instance), dot)), dot); proof') = proof'', text("and")
+  verify(comm(verify(comm(verify(comm(instance), dot)), dot)), proof'') = one
+$
+we see that the verifier first the verifier first has to derive the commitment
```
cdesaintguilhem: Duplicated "the verifier first"

spec/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+The technique is mostly useful in settings where the extra time spent by the prover
+is outweighed by the time saved by the verifier(s), 
+e.g., a computationally constrained verifier, or multiple verifiers.
+
+= Resolving growing instance complexity
+While recursive proving leads to a decrease in proof size, this is naively traded off
+against an increase in instance complexity.
+Looking at a depth-two recursive proof,
+$
+  &prove\(verify(comm(verify(comm(instance), dot)), dot); proof') = proof'', text("and")
```
cdesaintguilhem: I don't immediately see the usefulness of `&` and `\` in this equation.

spec/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+$comm(verify(comm(verify(comm(instance), dot)), dot))$
+from the given base instance $instance$ before verifying the proof.
+This increase in verifier computation is undesirable and should be avoided.
+
+A solution to this, is to leverage the following variation to the verification algorithm:
+$
+  verify': commitmentSpace^2 times {0, 1} times proofSpace: (commitment_0, commitment_1, b, proof) mapsto
+  cases(
+    verify(commitment_0, proof) &text("if") b=0,
+    verify(commitment_1((commitment_0, commitment_1); dot), proof) &text("if") b=1
```
cdesaintguilhem: I can't parse what `commitment_1((commitment_0, commitment_1); dot)` is meant to be here.

spec/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+#et("note shakiness of recursion")
+
+= Split processing
+#let record = $bb(r)$
+In practice, we find that the set of operations utilized for verification differs vastly from
+those typically performed by guest programs.
+Specifically, verification primarily involves hashing and (extension) field arithmetic, 
+where especially the second is absent in typical guest programs.
+
+Emulating field arithmetic on the a binary arithmetic-oriented VM, typically
```
cdesaintguilhem: See "the a".

spec/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+= Notation
+Let $BB := { zero, one }$ denote the boolean set and let 
+$functionSpace := {function: inputSpace times witnessSpace mapsto BB}$ denote 
+the set of functions mapping the (public) input space $inputSpace$ and (private) 
+witness space $witnessSpace$ to this set.
+Function-input pairs $(function, input) in functionSpace times inputSpace$ 
+are henceforth referred to as _function instances_, or simply _instances_.
+Where convenient, we may also denote this as $function(input; dot) in instanceSpace$
+with instance space $instanceSpace$.
+We then define $relation subset.eq instanceSpace$
```
cdesaintguilhem: This relation is actually a language.

spec/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+  to accelerate hash-verification) while incurring as little design overhead as possible.
+]
+
+In theory, any division of tasks between the two VMs would work.
+Moreover, it is unclear what division will lead to optimal performance.
+Yet, it is expected that divisions adhering to these high-level guidelines will 
+be a good first step towards a performant verifier:
++ have the field-VM perform all verification steps involving field arithmetic,
++ include all verifier-issued challenges required by these verification steps 
+  in the communication record $record$ as field-elements, so that the field-VM does not have to derive them,
```
cdesaintguilhem, line 202: "field element" is not hyphenated.

```suggestion
  in the communication record $record$ as field elements, so that the field-VM does not have to derive them,
```

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+= Recursive proving and split processing
+We lastly provide some notes on applying the recursive proving and split processing to
+the verification of a proof in the context of this VM.
+
+First, we note that in any scenario, the prover has to commit to the program $function$
+being exeucted and the public input $input$ that is provided.
+In a non-recursive proof, this is trivially done by committing to the `DECODE` table
+representing $function$, and the `PAGE` tables storing the public $input$.
+Since these commitments are deterministic, the verifier can locally reconstruct
+the commitments and verify any opening proofs against its own version of the commitment.
```
cdesaintguilhem, line 214: As an optimisation: if the verifier has the `DECODE` and relevant `PAGE` tables in the clear, the prover does not even need to provide the corresponding openings as part of the proof string.

cdesaintguilhem (commented):
Mostly minor comments!

spec/recursion.typ
erik-3milabs: replaced with `arrow.right`

spec/recursion.typ
erik-3milabs: I rewrote the paragraph, as discussed. Is it correct now?

spec/recursion.typ
erik-3milabs: indeed, they're no longer relevant. Removed them

spec/recursion.typ
erik-3milabs: cool! I didn't know that works! Nice 👌

erik-3milabs: I like the idea, but am not confident that it works well in practice. Specifically, for the ultimate verifier to confirm the proof, they would have to execute both v(x, proof) and v(v'(x, v', proof) to determine whether one of the two is true.

I think a better approach would be to absorb `b` into the proof. From a notation perspective, this still gets rid of the `b`, but now you have a bit which allows you to select the right verification configuration.

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,255 @@
...
+We thus have to demonstrate the commitments the prover provides are as expected.
+This is achieved by having the verification algorithm `COMMIT` (see @commit)
+to the public input it is provided.
+This act produces an imbalance in the LogUp-component of the proof-of-verification, 
+which must be balanced during verification in the _next_ recursion layer.
+In later recursions, the verifier must consistently `COMMIT` to its public input
+and use the _same_ public input to balance out the LogUp-component of the proof
+it is provided.
+This solution effectively kicks the can down the road; the final verifier has to
+provide the initial input to the program as input to verify the recursive proof.
```
erik-3milabs: Re 2.

Do you mean that having the guest `COMMIT` to $\mathbb{c}_1$ is excessive, since there must be a `PAGE` storing it, and all `PAGE`s are committed to by the VM?

In that case the verifier will have to add a constraint to that specific `PAGE` table, constraining that it contains the commitment on the required address :thinking: or were you thinking of a different solution?

RobinJadoul: There are multiple PAGEs that need to be present to correctly represent the ELF being loaded into memory, so those would be part of $\bar{\mathbb{x}}$ already. And the ELF is the public input part, potentially with the PAGEs for the commitments made by the guest program.
So in the end, we'd need these specific tables to be used already as precomputed tables (which could/would be part of a verification key that can be derived from the ELF once and not computed on every verification).

spec/recursion.typ
RobinJadoul: Sure, turning it from a hint into an explicit value inside the proof should work fine

spec/chapters/recursion.typ
RobinJadoul: The downside would be that you get an increased storage cost in the verifying key, where storing those tables on-chain would be very costly, compared to just merkle roots.

spec/recursion.typ
RobinJadoul: Maybe the theoretical part is actual nicer represented as an OR statement (not sure, but it avoids making assumptions on the structure of the proof), but we can have the bit inside the proof in practice, making it convenient in the implementation too.

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,260 @@
...
+In a non-recursive proof, this is trivially done by committing to the `DECODE` table
+representing $program$, and the `PAGE` tables storing the public $input$.
+Since these commitments are deterministic, the verifier can locally reconstruct
+the commitments and verify any opening proofs against its own version of the commitment.
+Note that in this case, the prover may exclude the opened value from any proof strings
+pertaining to the `DECODE` and `PAGE` tables, as these are already known to the verifier.
+
+When recursing on this process, the prover provides the verifier with this commitment.
+We thus have to demonstrate the commitments the prover provides are as expected.
+This is achieved by having the verificationan  algorithm `COMMIT` (see @commit)
```
RobinJadoul, line 220:
```suggestion
This is achieved by having the verification  algorithm `COMMIT` (see @commit)
```

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,260 @@
...
+- `COMMIT` to $comm(instance)$, $comm(verify'_b),$ and $comm(verify'_f)$ 
+- assert that $b in {0, 1}$,
+- verify proof:
+  - if $b=0$: execute $verify_b (comm(instance), proof)$
+  - if $b=1$:
+    + construct $comm(verify'(comm(instance), comm(verify'); dot))$ from $(comm(instance), (comm(verify'_b), comm(verify'_f)))$
+    + execute $verify_b (comm(verify'(comm(instance), comm(verify'); dot)), proof)$
+
+*$verify'_f\(comm(instance), (comm(verify'_b), comm(verify'_f)), b, proof, record)$:*
+- verify proof: $verify_b (comm(verify'(comm(instance), comm(verify'); dot)), proof)$
```
RobinJadoul, line 247:
```suggestion
- verify proof: $verify_f (comm(verify'(comm(instance), comm(verify'); dot)), proof)$
```

RobinJadoul (commented):
I'm wondering whether we should choose for the $f$ function in the proof system to be the ELF proof, or if it should purely be the VM itself, and moving the ELF into public inputs.

spec/recursion.typ
RobinJadoul: I think the offloading to the original v_f and v_b should work to kick the can down the road and have a full description elsewhere :+1:

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+#show math.equation.where(block: false): box
+
+#set list(marker: [---])
+
+= Proof recursion <proof-recursion>
+Let $BB := { zero, one }$ denote the boolean set and let 
+$programSpace := {program: inputSpace times witnessSpace to BB}$ denote 
+the set of functions mapping the (public) input space $inputSpace$ and (private) 
+witness space $witnessSpace$ to this set.
+We define instance space $instanceSpace := programSpace times inputSpace = {program: witnessSpace to BB}$;
```
RobinJadoul: Maybe 

```suggestion
We define instance space $instanceSpace := programSpace times inputSpace tilde.equiv {program: witnessSpace to BB}$;
```

as it's not an equality of sets?

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+program-input pairs $(program, input) in instanceSpace$ 
+are henceforth referred to as _function instances_, or simply _instances_.
+Where the individual components of the pair are irrelevant, an instance is 
+denoted as $instance in instanceSpace$.
+
+We define relation $relation subset instanceSpace times witnessSpace$ where $((program, input), witness) in relation$ if $program(input, witness) = 1$.
+This relation induces the language $language subset instanceSpace$ of _solvable instances_,
```
RobinJadoul:
```suggestion
We define relation $relation subset.eq instanceSpace times witnessSpace$ where $((program, input), witness) in relation$ if $program(input, witness) = 1$.
This relation induces the language $language subset.eq instanceSpace$ of _solvable instances_,
```

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+
+// Spaces and instances
+#let (programSpace, program) = ($cal(F)$, $f$)
+#let (inputSpace, input) = ($II$, $bb(i)$)
+#let (instanceSpace, instance, instance2) = ($XX$, $bb(x)$, $bb(y)$)
+#let (witnessSpace, witness) = ($WW$, $bb(w)$)
+#let (proofSpace, proof) = ($bb(Pi)$, $pi$)
+
+#let (commitmentSpace, commitment) = ($cal(C)$, $bb(c)$)
+#let commit(x) = $overline(#x)$
```
RobinJadoul, line 11:
```suggestion
#let commit(x) = $dash(#x)$
```

should work better for html export
(except there's some weird rendering issues where at least FF, and seemingly other browsers too, so maybe this should be a later effort to clean up browser rendering a bit, as chromium also seems generally quite bad atm)

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+they know a private input ($witness in witnessSpace$) such that the program terminates successfully 
+(i.e., $(program, input) in language$).
+To this end, the prover uses $prove(program(input; dot); witness)$ 
+to construct some proof $proof in proofSpace$ and sends this to the verifier.
+They then use $verify(comm(program(input; dot)), proof)$ to check that the proof is valid, 
+convincing them of the prover's claim.
+
+== Recursively proving proof verification
+Now observe that the verifier $verify$ is itself a function in 
+$verifierSpace := {hat(f): commitmentSpace times proofSpace to BB} subset.eq programSpace$.
```
RobinJadoul, line 72: I think that to use the proof as private input, we need to have $\hat{f}$ map from only $\mathcal{C}$, or let $\mathcal{V} \subseteq \mathbb{X}$
The former is probably the more interesting, though that kinda moves the meaning of $\hat{f}$ to "c commits to a satisfiable (/solvable?) instance"

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+Secondly, note that the size of the verification input 
+$(comm(instance), comm(verify'))$ is _constant_ in the depth of the recursion.
+In fact, the size of the verification instance is typically fully determined by 
+$comm(instance)$, since $comm(verify')$ can often be precomputed.
+
+It is important to note that we have not proven soundness of this construction.
+For example, there may exist proofs that _attest to the existence of itself_ in a finite number of recursion steps.
+Such a proof would be accepted by $verify'$ even if $instance in.not language$.
+In practice, one might be able to prevent this problem by including a recursion-level counter in the proof.
+The existence of other soundness gaps are not ruled out by the authors.
```
RobinJadoul, line 137: Feels a bit weird to phrase it like this, maybe something about not providing a full proof of soundness (yet)?

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+that would ideally be sent between the two verification halves.
+
+To achieve communication between two programs running on different VMs, 
+we thus introduce the concept of a prover-hinted _communication record_ $record$ 
+provided as input to both subprocesses.
+All values that are to be communicated from one part to another, are stated on this record.
+For each value on the record, the "sending" half _verifies_ that it is as expected, 
+whilst the "receiving" half _assumes_ its correctness and resumes verification under this assumption.
+One can now conclude that the proof satisfies the instance
+when both algorithm-halves produce the same record for this input.
```
RobinJadoul, line 190: Do they really "produce" a record now? Maybe "accept"?

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+  different proof system (e.g., moving to Flock 
+  #footnote(link(
+    "https://eprint.iacr.org/2026/1329", 
+    "Flock: Fast Proving for Batch Boolean Computations. src: https://eprint.iacr.org/2026/1329"
+  )) 
+  to accelerate hash-verification) while incurring as little design overhead as possible.
+]
+
+More formally, we define
+$v_0, v_1: instanceSpace times proofSpace to recordSpace$ as a valid _split_ of
```
RobinJadoul, line 208: commitment instead of instance?

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+
+More formally, we define
+$v_0, v_1: instanceSpace times proofSpace to recordSpace$ as a valid _split_ of 
+verifier $v in verifierSpace$ if
+$
+  forall (instance, proof) in instanceSpace times proofSpace: v(comm(instance), proof) = 1 iff v_0(comm(instance), proof) = v_1(comm(instance), proof),
+$
+where $recordSpace$ denotes the communication record space.
+That is, the two halves agree if and only if the instance-proof verifies succesfully.
+By applying the Kronecker delta function $delta$ #footnote("https://en.wikipedia.org/wiki/Kronecker_delta"),
```
RobinJadoul, line 215: Not sure the footnote is really needed.

spec/chapters/recursion.typ
RobinJadoul: It feels a bit weird for one half to "produce" the output of the other half as well.
Introducing the record as extra parameter directly instead of deriving $\tilde{v_b}$ should keep the argument intact, and merely define the validity of a split differently.
Something like `∀(x, π): v(x, π) = 1 <=> ∃r. v_0(x, π, r) = 1 AND v_1(x, π, r) = 1` should probably work

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+That is, $tilde(v)_0, tilde(v)_1$ indicate whether the provided record $record in recordSpace$
+would indeed be produced by $v_0$ respectively $v_1$ when provided $instance$ and $proof$.
+With this transformation in place, one can express 
+$
+	v(comm(instance), proof) 
+	= tilde(v)_0(comm(instance), [proof, d(comm(instance), proof)]) 
+		dot tilde(v)_1(comm(instance), [proof, d(comm(instance), proof)]).
+$
+where $d(comm(instance), proof)$ denotes the record deriviation function.
+We introduce the function product $(f || g)(input, witness) := f(input, witness) dot g(input, witness)$ for $f, g in programSpace$.
```
RobinJadoul: Notation feels slightly weird, maybe $f \wedge g$?

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+== Recursive instance verification
+In @growing-instance-complexity, it is assumed that one can efficiently construct
+the commitment $comm(y([comm(x), comm(y)], dot))$ from $(comm(x), comm(y))$.
+We show that this assumption holds in practice.
+
+First, note that for an instance $program(input, dot)$, the program 
+$program$ and input $input$ are committed to separately.
+Recall that a (split) program is encoded as one or more 
+`DECODE` (@decode) and/or `FIELD-DECODE` (@field-decode) tables, while the 
+input $input$ --- and private input $witness$ and record $record$ for that matter
```
RobinJadoul, line 346: Technically, public input is committed private input, but committed values are now available through COMMIT's PAGEs

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+    evaluates to the column entries on specific inputs,
+  + the polynomial is _encoded_, creating a codeword of polynomial evaluations 
+    over a different, larger domain,
+  + the codewords are _batched_, turning a list of codewords into a codeword of lists, and
+  + the batched codeword is _merkle-committed_, yielding the tree root.
+])
+
+Aside from this proof-system level commitment contraption, we can additionally leverage
+the `COMMIT` chip to commit to individual bytes in the input.
+Any values sent by the guest program to `std::out` are recorded in this chip's
```
RobinJadoul:
```suggestion
Any values sent by the guest program to `stdout` are recorded in this chip's
```

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+table, with the table set up such that it causes an imbalance in the interaction logic
+of the proof system during proving.
+During verification, it can be verified that this imbalance is caused by the 
+`COMMIT`ments of this public information.
+
+Leveraging the `COMMIT` chip is mostly useful in situations where revealing full `PAGE`
+tables is unnecessarily expensive or cumbersome, e.g., when committing to small 
+amounts of data or to data scattered across several `PAGE` tables.
+One potential limitation is that `COMMIT`ting in this way is at the discretion of the VM's
+guest program.
```
RobinJadoul, line 372: I think this model of COMMIT is outdated since epoch/L2G system for committing? Though I suppose a COMMIT-domain PAGE table is not strictly different from just providing the same interactions from the verifier.

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+
+== Summary
+
+*Communication record $record$:*\
+The communication record contains all challenges derived from the proof by
+means of the Fiat-Shamir transformation.
+
+*`RiscV-VM` subalgorithm*
+$tilde(v)_b ([comm(instance), comm(instance2)], [[proof, b], record])$:
+- `COMMIT`s to $comm(instance), comm(instance2)$ by sending both halves to `std::out`.
```
RobinJadoul:
```suggestion
- `COMMIT`s to $comm(instance), comm(instance2)$ by sending both halves to `stdout`.
```

spec/chapters/recursion.typ
```diff
@@ -0,0 +1,404 @@
...
+- derives challenges from $proof$ by means of Fiat-Shamir, and assert that they 
+  match those located in $record$ at the expected location.
+- Performs the binary arithmetic steps required to verify that $proof$ attests
+  to $instance$ (when $b=0$) or $instance2$ (when $b != 0$#footnote[The RiscV-VM's `PAGE` tables only supports byte data. All non-zero data is treated as $b=1$.]),
```
RobinJadoul, line 395: We can assert the bit-ness of `b` as well, alternatively

RobinJadoul (commented):
Not paying attention to all possible typography yet, trying to get the last high-level stuff out of the way
I'll try to do another pass later this week, but this is a first go at it

spec/chapters/recursion.typ
erik-3milabs: Let's discuss this in person

spec/chapters/recursion.typ
erik-3milabs: sure. wdyt of
```suggestion
The provision of a full soundness proof is deferred.
```
?

spec/chapters/recursion.typ
erik-3milabs: A)
```suggestion
when both algorithm-halves accept the same record for this input.
```

B) let's discuss this in person.

spec/chapters/recursion.typ
erik-3milabs: like this, you mean, right?
```suggestion
$v_0, v_1: commitmentSpace times proofSpace to recordSpace$ as a valid _split_ of 
```

spec/chapters/recursion.typ
erik-3milabs: I included it, as I've encountered a wild variety of notation for the delta function and wanted to point the reader to the exact definition I'm using here. We could alternatively derive it in the document and keep it self-contained. wdyt?

spec/chapters/recursion.typ
erik-3milabs: why would this model be outdated? I don't see how the epoch/L2G system changed this.

spec/chapters/recursion.typ
erik-3milabs: You mean by including
```
x = EQ[b, 0];
y = EQ[b, 1];
z = x + y;
z = EQ[z; 1];
```
in the guest program?

spec/book.typ
erik-3milabs, line 1: hmmm, this should be removed, no?

spec/chapters/recursion.typ
erik-3milabs: I'm leaving it out of the commitment-batch for now, but would be happy to switch once we've established how to fix the html support issue.

spec/chapters/recursion.typ
erik-3milabs: A) _some_ committed values (namely those COMMITted by the guest program) are available through the `COMMIT`'s PAGEs, right?

B) how would you propose the text be modified? I'm confused by the remark.

PR #944, opened 2026-08-21
spec: Experimental switch from shiroa to typst html bundle export

RobinJadoul:
This moves most of the pain of working with shiroa into something more under our control, and additionally gets significant compilation speed improvements.
Minimal required typst version becomes 0.15.0 because of this.


spec/assets/style.css
```diff
@@ -0,0 +1,558 @@
...
+  grid-area: nav;
+  background: var(--sidebar-bg);
+  border-right: var(--border-size) solid var(--border);
+  padding: 20px;
+  overflow-y: auto;
+  min-height: calc(100vh - var(--header-height) - var(--border-size));
+  font-size: 0.9em;
+  position: sticky;
+  top: calc(var(--header-height) + var(--border-size));
+  max-height: calc(100vh - var(--header-height) - var(--border-size));
```
erik-3milabs:
```suggestion
  height: calc(100vh - var(--header-height) - var(--border-size));
  font-size: 0.9em;
  position: sticky;
  top: calc(var(--header-height) + var(--border-size));
```

spec/chapters/bitwise.typ
```diff
@@ -12,8 +11,6 @@
 #let chip = load_chip("src/bitwise.toml", config)
 
 #let bitwise = raw(chip.name)
-
-#show: book-page(chip.name)
```
erik-3milabs, line 16: ❤️

erik-3milabs (commented):
1. I believe the font is too small; I can only read comfortably at 110% zoom.
2. the color scheme is 3MI's; should be Aligned/Lambdaclass's (?)
4. sidebar:
    a. the sidebar minwidth is too small; text wraps when page is made too narrow (but not narrow enough to minimize sidebar)
    b. when sidebar is minimized, the word "Navigation" is too close to the hamburger ≡
2. lightmode/darkmode:
    a. I cannot find a manual toggle; it defaults to the browser setting; I typically prefer darkmode, but not here. :shrug:
    b. in darkmode, highlighted areas (e.g. Lemma 5, sec 33.3) is barely noticable
    c. in darkmode, highlight around `raw` sections (e.g., tags, ranges, var types, etc.) is not noticable.
6. Re: tables:
    a. Range col of virtual overview is too small for DVRM and KECCAK
    b. the "About Ecall" section has some weird itemize misalignment issue 
    c. the tag column is too crammed on 10+ assumption/constraint tables
    d. the "range" column is too crammed on ~10 constraint tables
7. Re: file organization: 
    a. I'd be inclined to move the .css / .js files to a separate folder. wdyt?
    b. perhaps this is also a good moment to clean this folder up and move all section .typ files to a separate folder?

That being said: thanks for taking the time to make this happen. It's great!

spec/chapters/bitwise.typ
erik-3milabs: Also, :clap: for the fact that this now runs on my pc without crashing :heart:

spec/.gitignore
```diff
@@ -1,4 +1,3 @@
-dist/*
-interaction_count.json
-ebook.pdf
+bundle/*
+spec.pdf
```
erik-3milabs, line 2: Is there a reason to just block `spec.pdf`? Why not just `*.pdf`?

spec/bundle.typ
erik-3milabs, line 3: This doesn't seem to work 100%; executing `typst watch bundle.typ` doesn't trigger this assert (where I'd expect it does).

spec/assets/style.css
```diff
@@ -0,0 +1,651 @@
...
+
+@container style(--theme: dark) {
+  #theme-toggle {
+    background-color: var(--bg-subtle);
+  }
+
+  #theme-toggle input ~ span::before {
+    content: "☀︎︎";
+  }
+}
```
erik-3milabs, line 422: I would expect the background-color of the button to either
1. always match the background of the header, or
2. always match the header background color you'd get if you hit the toggle button

i.e., something like

```suggestion
#theme-toggle {
  border: 1px solid var(--text);
  border-radius: 2em;
  width: 2em;
  height: 2em;
  display: flex;
  align-items: center;
  justify-content: center;
  color: var(--gold);
  background-color: var(--text-muted);
}

#theme-toggle input ~ span::before {
  content: "⏾";
}

#theme-toggle input {
  display: none;
}

@container style(--theme: dark) {
  #theme-toggle input ~ span::before {
    content: "☀︎︎";
  }
}
```
or
```suggestion
#theme-toggle {
  border: 1px solid var(--text);
  border-radius: 2em;
  width: 2em;
  height: 2em;
  display: flex;
  align-items: center;
  justify-content: center;
  color: var(--gold);
  background-color: var(--bg-muted);
}

#theme-toggle input ~ span::before {
  content: "⏾";
}

#theme-toggle input {
  display: none;
}

@container style(--theme: dark) {
  #theme-toggle input ~ span::before {
    content: "☀︎︎";
  }
}
```

(I don't know whether/how that debate was settled in the accessibility community...)

What is your reason for having the button's background color stay roughly constant?

erik-3milabs (approved):
This PR looks and feels really solid 👍

Given this, I don't think it is worth me spending more time on reviewing all the tiny tricks you performed to make things look and feel this way.

I'm leaving three minor comments of things that caught my eye as I perused this change list.

spec/.gitignore
RobinJadoul: A habit of avoiding `*.pdf` because sometimes you want to include figures as PDF

spec/bundle.typ
RobinJadoul: Yeah, I noticed that too, at some point, but I think it's an effect of having a panic inside context, which doesn't really propagate.
I'm not sure how to solve it better, tbh.

spec/assets/style.css
RobinJadoul: I found the "use the other background color" to be too jarring, and "use the normal background color" to not really stand out enough.
The current choice looked okay to my eyes, with no explicit accessibility reasoning behind it

PR #971, opened 2026-09-08
spec: Field VM

RobinJadoul:


spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+This chapter describes, in line with the split between a binary and a field VM from @recursion,
+the ISA and an arithmetization of a dedicated field VM.
+The ISA is centered around a single, versatile instruction, that can handle both
+extension field arithmetic and program flow.
+
+= ISA
+
+The central instruction of the ISA is a constraint for a fused multiply-add over the extension field:
+`FMA o == a * b + c`.
+Here all of `o`, `a`, `b` and `c` are arguments following the addressing scheme described below.
```
erik-3milabs, line 14: Reading this section, I've noticed that this `o` is easy to misinterpret. Also, `o` and just o are hardly distinguishable. Would it be an idea to rename this as `d` (for "destination")? I also considered `r` for "result", but that also overlaps with "register" so that might not be the greatest...

erik-3milabs: > "The central instruction of the ISA is a constraint"

We could (and maybe should?) spend a full paragraph unpacking the meaning of this single phrase. Starting with the purpose of this VM: verifying a proof. In other words, this VM doesn't need to really produce anything (like traditional programs), but rather just needs to check that a string has an acceptable format, by performing loads of assertions. (We do have some typical instructions, but those are to writeable registers only, and meant to reduce proof size.)

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+The field VM uses only read-only memory, which is implemented as a committed table,
```
erik-3milabs, line 15:
```suggestion
The field VM only uses read-only memory, which is implemented as a committed table,
```
or
```suggestion
The field VM uses read-only memory only, which is implemented as a committed table,
```
or something else that bypasses the double "only"

erik-3milabs, line 15: this does not mention the existence of writeable registers.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+
+The VM has a state consisting of $N$ general purpose extension field registers,
+a base-field `PC` register, and a bit-register `ZERO`.
+The number of registers was chosen as a tradeoff between the versatility of having more mutable state,
+and the extra cost in committed columns and decoding logic that grows with $N$.
+#rj[Register index for `ZERO` = $0$ and `PC` = $1$ and gp register `Ri` = $i + 2$]
+
+Each argument to the FMA constraint has either of the two following forms:
+- `imm_0 * reg + imm_1`
+- `MEM[imm_0 * reg + imm_1]`
```
erik-3milabs:
```suggestion
- $#`imm_0` dot #`reg` + #`imm_1`$
- $#`MEM`[#`imm_0` dot #`reg` + #`imm_1`]$
```
or
```suggestion
- $#`imm_0` times #`reg` + #`imm_1`$
- $#`MEM`[#`imm_0` times #`reg` + #`imm_1`]$
```

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+
+== Arguments and addressing
+
+The VM has a state consisting of $N$ general purpose extension field registers,
+a base-field `PC` register, and a bit-register `ZERO`.
+The number of registers was chosen as a tradeoff between the versatility of having more mutable state,
+and the extra cost in committed columns and decoding logic that grows with $N$.
+#rj[Register index for `ZERO` = $0$ and `PC` = $1$ and gp register `Ri` = $i + 2$]
+
+Each argument to the FMA constraint has either of the two following forms:
```
erik-3milabs:
```suggestion
Each argument to the `FMA` constraint has either of the two following forms:
```

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+- `imm_0 * reg + imm_1`
+- `MEM[imm_0 * reg + imm_1]`
+where each immediate is a base field element, encoded in the instruction for a specific argument,
+and `MEM[...]` reads memory.
```
erik-3milabs, line 30: I think we should be more precise here than "MEM reads memory". Something like "MEM[x] indicates the value is obtained from memory at address `x`"

Also, starting in section 35.1.3, the "MEM" is completely dropped and the square brackets remain. Either way is fine, but I think we should be more consistent..

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+The `o` argument to the instruction obtains its register values from the _future_ state.
```
erik-3milabs:
```suggestion
The `o` argument to the instruction obtains its register value from the _future_ state.
```
singular, right?

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+That is, the state from which the next instruction will get its input register values.
```
erik-3milabs: Hmmm, this definition of "future state" is really hard to parse, as you're introducing concepts in reverse order:
- first we introduce the concept of "future" state, 
- only then to (retroactively) define what a "state" even is ("a collection of values from which an instruction gets its input register values")
And even then, the definition of "state" is rather abstract.

I prefer to be presented with information in linear order. Is retroactively defining things a literary thing?

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+The `ZERO` register of the future state contains a bit (as a base field element) that indicates whether or not
```
erik-3milabs: These sentences are joined without connector word (implicitly suggesting it is an enumeration of likeminded concepts), but we're actually doing two different things:
1. `o[i]` **gets** its register value from the next state
2. `o[i] = 0` **informs** whether the ZERO flag in the next state.

I find that confusing. I would be inclined to reason where register values come from, i.e.
1. the o arg gets its register value from the future state
2. the ZERO flag is set when the o arg in the previous state is zero.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+
+= ISA
+
+The central instruction of the ISA is a constraint for a fused multiply-add over the extension field:
+`FMA o == a * b + c`.
+Here all of `o`, `a`, `b` and `c` are arguments following the addressing scheme described below.
+The field VM uses only read-only memory, which is implemented as a committed table,
+together with the multiplicities with which each element is accessed.
+
+== Arguments and addressing
```
erik-3milabs, line 18: I'm not a fan of the ordering in this subsection:
1. introduce gp registers, PC and ZERO (in inverse order of their indexing?)
2. decompose FMA argument,
3. talking about the FMA's o-argument (for which I now have to recall what that again meant)
4. talk about ZERO
5. talk about PC.

Why not introduce them as
1. ZERO, PC, N gp registers
2. talk about ZERO
3. talk about PC
4. talk about registers
--- at this point we know everything there is to know about registers ---
6. FMA argument decomposition
7. talk about how FMA is mapped over three consecutive states (i-1, i, and i+1), and how it all gets cobbled together.
?

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+The `o` argument to the instruction obtains its register values from the _future_ state.
+That is, the state from which the next instruction will get its input register values.
+The `ZERO` register of the future state contains a bit (as a base field element) that indicates whether or not
+the `o` value was zero for the current instruction.
+The `PC` register contains the program counter (as a base field element) and indicates which instruction is to be executed.
+Every other register can hold an arbitrary extension field element.
+
+== Register hints
+
+Each register (other than `ZERO` and `PC`) in the current state can be marked as _hinted_ in an instruction.
```
erik-3milabs:
```suggestion
Each general-purpose register in the current state can be marked as _hinted_ in an instruction.
```

erik-3milabs: > Each register [...] in the current state can be marked as _hinted_ in an instruction.

We're again mixing state and instruction here and it is confusing.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+This means that its value from the current instruction onward can get a new value
+that is independent from the previous value, except as constrained by the instruction.
+Additionally, the _output_ can be marked as hinted, meaning that the register used in the `o` argument
+whether wrapped in a `MEM[]` lookup or not, will change in the future state, and as such in the `o` argument too.
```
erik-3milabs:
```suggestion
Additionally, the _output_ can be marked as hinted, meaning that the register used in the `o` argument ---
whether wrapped in a `MEM[]` lookup or not --- will change in the future state, and as such in the `o` argument too.
```

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+That is, the state from which the next instruction will get its input register values.
+The `ZERO` register of the future state contains a bit (as a base field element) that indicates whether or not
+the `o` value was zero for the current instruction.
+The `PC` register contains the program counter (as a base field element) and indicates which instruction is to be executed.
+Every other register can hold an arbitrary extension field element.
+
+== Register hints
+
+Each register (other than `ZERO` and `PC`) in the current state can be marked as _hinted_ in an instruction.
+This means that its value from the current instruction onward can get a new value
```
erik-3milabs, line 41: > This means that its value [...] can get a new value ...

Its value can get a new value? A register can get a value, or a value can be modified. But a value getting a new value isn't thing, I'd say.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+that is independent from the previous value, except as constrained by the instruction.
```
erik-3milabs, line 42: I think the point we're trying to make here is that
- in typical code execution, only the value held by the output variable is be updated.
- this isn't typical execution; we only constrain that the output is an FMA of the input: we can rearrange this constraint such that any of the arguments is the output. Heck, there could even be multiple outputs.
- hence, we can designate any of the arguments to be updated.

but I don't think this brings that point across clearly.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+Additionally, the _output_ can be marked as hinted, meaning that the register used in the `o` argument
+whether wrapped in a `MEM[]` lookup or not, will change in the future state, and as such in the `o` argument too.
+Output hinting is commonly used to assign the result of a computation: `FMA X == X * X, hint out` would
+compute `X * X` and re-assign it to `X` in the future state.
```
erik-3milabs:
```suggestion
compute $#`X` dot #`X`$ and assign the result to `X` in the future state.
```

spec/chapters/field_vm.typ
erik-3milabs: This sentence is too complex.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+The output hint, in contradiction with the input hints, does allow `PC` to be hinted, so as to enable
```
erik-3milabs, line 47: > The output hint, in contradiction with the input hints [..]

At the start of this paragraph, you mention that "each register in the current state can be hinted". In our side-discussion, you indicate that this is what you meant by "input hinting", but that definition does does not become clear from context, imo.

Secondly, I think that "Unlike input hints, the output hint..." might be easier to read here (primarily because "in contradiction with" suggests (to me) that the two types of hints are at odds with one another. But perhaps that is me failing at English)

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+causal jumps and control flow in the program.
+Any register that is not hinted will have the same value in the future state as in the current state,
```
erik-3milabs, line 49: again, I'm getting the sense that we're meshing instructions and constraints, and it is tough to read.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+with the appropriate exceptions in behaviour for the `ZERO` and `PC` registers.
+
+#aside("Hint collisions")[
+One may observe that an output hint for state `i` and an input hint on state `i + 1` can affect
```
erik-3milabs: register hint, not input hint.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+an identical register in an identical state.
```
erik-3milabs, line 54: why use "identical" in favor of "same" here? (IIUC, two things can be identical and yet not be the same).
Isn't it possible for two identical states not to be the same?

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+While this is true in theory, it is not a problem in practice, as two successive states are,
+in almost all cases, operated on by two consecutive ---in the program text--- instructions,
+and as such, hinting collisions can be easily identified, and most actual programs
+should have no reason to have hinting collisions.
```
erik-3milabs, line 58: Please write shorter sentences. They are easier to read:

```suggestion
While this is true in theory, it is not a problem in practice.

In almost all cases, two successive states are operated on by two consecutive ---in the program text--- instructions.
As such, hinting collisions can be easily identified, and most actual programs should have no reason to have hinting collisions.
```

also
> --- in the program text---

? Do you mean context? Haven't heard of 'program text' before.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+Additionally, the _output_ can be marked as hinted, meaning that the register used in the `o` argument
+whether wrapped in a `MEM[]` lookup or not, will change in the future state, and as such in the `o` argument too.
+Output hinting is commonly used to assign the result of a computation: `FMA X == X * X, hint out` would
+compute `X * X` and re-assign it to `X` in the future state.
+The output hint, in contradiction with the input hints, does allow `PC` to be hinted, so as to enable
+causal jumps and control flow in the program.
+Any register that is not hinted will have the same value in the future state as in the current state,
+with the appropriate exceptions in behaviour for the `ZERO` and `PC` registers.
+
+#aside("Hint collisions")[
```
erik-3milabs, line 52: It might be worth mentioning here that hint collisions do _not_ allow the prover to cheat (and why); they're really just a nuisance that may make some programs unprovable, as two distinct constraints may require one register to hold two different values.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+- `()` indicate grouping to separate the arguments
+- `X, Y, Z, W` are placeholder register names
+- `hint` notation indicates which registers are hinted (default unhinted), `hint out` means hinting `o` as above
+
+We note that this may be insufficient for the execution/prover side of the program,
+as this provides no information on _which_ value exactly should be hinted,
+but leave this as an implementation detail to be decided upon based on practical experience.
+
+We label the instruction with an `FMA` mnemonic, even though that is the only possible "real" instruction,
+in order to allow program listings to include other mnemonics to indicate pseudoinstructions that
```
erik-3milabs:
```suggestion
We label the instruction with an `FMA` mnemonic --- even though that is the only possible "real" instruction ---
to allow program listings to include other mnemonics to indicate pseudoinstructions that
```

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+map more specialized semantics onto the FMA functionality.
+Next, we suggest some potential pseudoinstructions along with their translation.
+This list is meant as an example, rather than an exhaustive enumeration;
+implementers and practitioners are encouraged to discover and use their own,
+as experience may point out further useful abstractions.
+
+/ `ADD o, a, b`: Addition: `FMA o == (0 * X + 1) * a + b, hint out`
+/ `MUL o, a, b`: Multiplication: `FMA o == a * b + (0 * X), hint out`
+/ `INV o, a`: Extension field inversion. Note: `o` and `a` cannot use the same register here: `FMA o == (a + 1) * o - 1, hint <register of o>`
```
erik-3milabs, line 97: Once finalized, I think this overview can be refactored a bit to improve the readability. Right now, it looks messy.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+/ `J a`: Jump. Can be to a register, memory content, or absolute address, depending on the addressing mode of `a`, even relative to PC: `FMA PC == a, hint out`
```
erik-3milabs:
```suggestion
/ `J a`: Jump: `FMA PC == a, hint out`. The jump can be to a register, memory content, or absolute address, depending on the addressing mode of `a`, even relative to PC
```

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+/ `JZA imm`: Jump if ZERO, absolute target address: `FMA PC == (ZERO)*(-1*PC+(imm-1))+(1*PC+1), hint out`
+/ `JZR a`: Jump if ZERO, PC-relative target address: `FMA PC == (ZERO) * (a - 1) + (PC + 1), hint out`
+/ `JNZA imm`: Jump if not ZERO, absolute target address: `FMA PC == ZERO * (PC - imm) + (ZERO + imm), hint out`
+/ `JNZR a`: Jump if not ZERO, PC-relative target address: `FMA PC == (a - 1) * (-1 * ZERO + 1) + (PC + 1), hint out`
+
+Eventually, we hope that a set of common pseudoinstructions can be extracted from actual usage,
+and inform potential optimizations that remove unused capabilities (e.g. reducing the number of immediates involved).
```
erik-3milabs, line 105: turn into a todo?
Idk, feels weird to put "hope" in a spec.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+
+== Calling convention
+
+Since the VM makes use of read-only memory, traditional usage of a program stack does not work.
+We assume that each function invocation (unless other optimizations apply) will have an associated _frame_,
+pointed to by a _frame pointer_ `fp`, one of the general purpose registers.
+In this frame, the following data is stored:
+
+/ `MEM[fp - k]...MEM[fp - 1]`: `k` saved registers from the calling function
```
erik-3milabs, line 114: will those always be right behind? What about when a function calls a function, does some work and then again calls a function. When inside the second function call, wouldn't `MEM[fp-1]` hold additional information required by the _first_ function call? (assuming they were placed back-to-back)

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+Then, we compute all values $#`imm`_0 dot #`reg` + #`imm`_1$,
+where we need to multiplex out of `registers`, based on `argument_registers[i]`.
+We do this by constructing the Lagrange basis polynomials $f_(i)(x)$ such that $f_(i)(j) = 1$
+for $i = j in [0, N + 1]$ and $f_(i)(j) = 0$ for $i != j in [0, N + 1]$.#footnote[
+  Note that we allow ourselves to multiplex $N + 2$ registers here,
+  combining the $N$ general purpose registers, the PC and the `ZERO` register.
+]
+Since the degree of these $f_(i)(x)$ can grow too large to express in a single polynomial constraint,
+we perform a _"degree split"_:
+$ f_(i)(x) = f_(i, 0)(x) + x^(d - 1) (f_(i, 1)(x) + x^(d - 2) (f_(i, 2) + x^(d - 2) (f_(i, 3) + ...))), $
```
erik-3milabs, line 165: This nested notation _suggests_ additional multiplications (even though you can collapse the x^k dot x^j terms), which trips me up. Is there a reason for not expanding it here already?

Also, the transition from f_i(j) to f_i,k(j) isn't clearly defined: is the k just an index? then f^((k)) might be neater?

Also, where did the `x`s go on fi2 and fi3?

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+for a maximal constraint degree $d$.
+Here, $deg f_(i, 0) <= d - 2$ and $deg f_(i, k) <= d - 3$.
+We denote by $t + 1$ the number of non-zero $f_(i, k)$ for fixed $i$.
+This allows us to first compute the values of
+$#`argument_registers[i]`^(d - 1)$, $#`argument_registers[i]`^(2d - 3)$ and so on
+to `arg_register_powers` with constraints of degree $<= d$,
+and then compute $#`args_premem[i]` = #`argument_scalars[i]` dot sum_(j = 0)^(N + 1) #`registers[j]` dot f_(j)(#`argument_registers[i]`) + #`argument_offsets[i]`$.
```
erik-3milabs, line 172: consider putting this equation on a separate line

spec/src/field_vm.toml
```diff
@@ -0,0 +1,216 @@
...
+    ["idx", "argument_registers", "i"],
+    ["idx", "argument_registers", "i"]]]
+iter = ["i", 0, 3]
+
+# PARAM: If T > 1, we need to populate arg_reg_pows_computed[i][j], j > 0, too, with power j * (d - 2) + (d - 1)
+
+[[constraints.mux]]
+kind = "arith"
+# PARAM: $#`args_premem`_i = #`argument_offsets`_i + #`argument_scalars`_i dot limits(sum)_(j = 0)^(N + 1) #`all_regs`_j dot (limits(sum)_(k = 0)^T #`arg_register_powers`_(i,k) dot (limits(sum_(l = 0)^(d - 2) #`argument_registers`_i^l dot #`MUX`_(j,k,l)))$
+constraint = "$#`args_premem`_i = #`argument_offsets`_i + #`argument_scalars`_i dot limits(sum)_(j = 0)^6 #`all_regs`_j dot (limits(sum)_(k = 0)^1 #`arg_register_powers`_(i,k) dot (limits(sum)_(l = 0)^3 #`argument_registers`_i^l dot #`MUX`_(j,k,l)))$"
```
erik-3milabs, line 135: I recall sum_(i=0)^j 1 = j+1, i.e., the bounds are _inclusive_. In that case, I'm observing a degree 6 constraint here.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+The coefficients for all $f_(i, k)$ are pre-computed once, based on the choices of `N`, `d` and `t`,
+and used through the `MUX` constant columns.
+#rj[Analysis of relation between $d$, $N$, $t$; mention current choice of $(d, N, t) = (5, 5, 1)$]
+Observe that the handling for `argument_registers[0]` is separate, as this represents the output argument,
```
erik-3milabs, line 176: might be nice to ref the constraint +  explicitly mention that ' is used to reference entries from the "next" row (i.e., the one with timestamp + 1).

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+
+#render_constraint_table(chip, config, groups: "fma")
+
+Finally, we must ensure the consistency between consecutive rows of the table,
+and allow for hinting.
+We again make use of the multiplexing machinery from before.
+The constraints we want to enforce on a register index $r$ are as follows:
+- $!#`hint_input`'_r and !#`hint_output` => #`registers`'_r = #`registers`_r$, `r` could not have been hinted,
+  since it was not input-hinted in the next row, and there was no output hint, so the next `r` should remain the same.
+- $!#`hint_input`'_r and f_r(#`argument_registers`_0) = 0 => #`registers`'_r = #`registers_r`$
```
erik-3milabs: fix compile issues:
```suggestion
- $!#`hint_input`'_r and f_r (#`argument_registers`_0) = 0 => #`registers`'_r = #`registers`_r$
```

erik-3milabs: consider line-breaking after the mathematical description.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+  `r` was not input-hinted in the next row, and it was not the output register, so it once again stays the same.
+
+This is equivalent to the logical statement $!#`hint_input`'_r and not (#`hint_output` and f_r(#`argument_registers`_0) = 1) => #`registers`'_r = #`registers`_r$, but expressed in a way that polynomial constraints can more easily handle.
```
erik-3milabs:
```suggestion
This second constraint is logically equivalent to $!#`hint_input`'_r and not (#`hint_output` and f_r (#`argument_registers`_0) = 1) => #`registers`'_r = #`registers`_r$, but expressed in a way that polynomial constraints can more easily handle.
```

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,220 @@
...
+Naturally, the PC gets an exception since if it is not (output-)hinted, we need $#`pc`' = #`pc` + 1$,
+and the $#`ZERO`'$ register purely depends on $#`args`_0$ and not on `ZERO`.
```
erik-3milabs:
```suggestion
Naturally, the `PC` and `ZERO` registers are exceptions since we need $#`pc`' = #`pc` + 1$ if it is not (output-)hinted,
and $#`ZERO`'$ purely depends on $#`args`_0$ and not on `ZERO`.
```

spec/src/field_vm.toml
```diff
@@ -0,0 +1,216 @@
...
+# PARAM: ["-", ["idx", "args_premem", 0], ["idx", "argument_offsets", 0], ["*", ["idx", "argument_scalars", 0], ["sum", ["=", "j", 0], ["+", "N", 1], ["*", ["idx", ["next", "all_regs"], "j"], ["sum", ["=", "k", 0], "T", ["*", ["idx", ["idx", "arg_register_powers", "k"], 0], ["sum", ["=", "l", 0], ["-", "d", 2], ["*", ["^", ["idx", "argument_registers", 0], "l"], ["idx", ["idx", ["idx", "MUX", "l"], "k"], "j"]]]]]]]]]
+poly = ["-", ["idx", "args_premem", 0], ["idx", "argument_offsets", 0], ["*", ["idx", "argument_scalars", 0], ["sum", ["=", "j", 0], 6, ["*", ["idx", ["next", "all_regs"], "j"], ["sum", ["=", "k", 0], 1, ["*", ["idx", ["idx", "arg_register_powers", 0], "k"], ["sum", ["=", "l", 0], 3, ["*", ["^", ["idx", "argument_registers", 0], "l"], ["idx", ["idx", ["idx", "MUX", "j"], "k"], "l"]]]]]]]]]
+
+[[constraint_groups]]
+name = "memory"
+
+[[constraints.memory]]
+kind = "arith"
+constraint = "$!#`mem_flags[i]` => #`args[i]` = #`args_premem[i]`$"
```
erik-3milabs, line 152: general note: why do we use ! to indicate the not-gate in this spec, when there is ¬?

spec/expr.typ
```diff
@@ -238,6 +242,10 @@
       assert(e.len() == 3, message: "Invalid type cast: " + repr(e))
       cwrap($#rec(PREC.cast, e.at(1)) colon.double #type_to_math(e.at(2))$, pp < PREC.cast)
     },
+    "next": (pp, rec, e) =>  {
+      assert(e.len() == 2 and type(e.at(1)) == str, message: "Invalid transition variable: " + repr(e))
+      cwrap($#rec(PREC.next, e.at(1))'$, pp < PREC.next)
```
erik-3milabs, line 247: why `cwrap` instead of `mwrap`? (I see now that "cast" also uses cwrap here, which is also confusing)

spec/expr.typ
```diff
@@ -57,16 +58,17 @@
...
-  "mul": 4,  // *
-  "div": 5,  // /
-  "mod": 6,  // mod
-  "sum": 7,  // Σ
-  "not": 8,  // not
-  "sub": 9,  // -
-  "add": 10,  // +  
-  "eq": 11,   // = and :=
-  "MAX": 12, // <the void outside every expression>
+  "next": 3, // var'
```
erik-3milabs, line 61: I'm noticing that `[idx, [next, var], i]` is expressed as $(var')_i$, where I would expect that $var'_i$ would also be acceptable. In fact, the added parentheses make reading the formula look bloated.

1. do you agree that the added parentheses are undesirable?
2. should we perhaps move the idx code to 3 (after next)? I'm not expecting we'll be indexing on a pow or a neg any time soon.

erik-3milabs (commented):
Summary:
- instruction and constraints are similar, but not the same. Sketching the right mental model is informative.
- an instruction spans multiple states, which is a new concept in this spec. The reader is not helped to understand this. Sketching the right mental model is informative.
- the concepts of input hinting, output hinting and register hinting and their relation is unclear.
- have another look at the lower and upper bounds of sum. There seem to be some issues with that.
- minor notes/rephrase proposals

spec/chapters/field_vm.typ
RobinJadoul: The idea is that the frame is big enough to contain the entire range from `fp - k` up to `fp + S`, for some frame size S the callee needs, just that we don't point to the start of the frame, because the callee shouldn't care

spec/chapters/field_vm.typ
RobinJadoul: > Is there a reason for not expanding it here already?

I wanted to keep the mathematical description clear, more than the exact way in which the constraints will use it. And expanding it would mean needing more complex exponents of the form `x^((d - 1) + (k - 2) * (d - 2))`, which felt less readable to me.

> is the k just an index? then f^((k)) might be neater?

The `f_(i,k)` are just different polynomials, using both `i` and `k` as indices for naming. I'd find the suggested alternative notation more confusing, I think, as it may look more directly related to `f`, like a k-fold repetition or the k'th derivative.

spec/src/field_vm.toml
RobinJadoul: It's only syntactically a degree 6 constraint.
By construction of MUX, all `MUX[j][k][5]` are zero, except for `MUX[j][0][5]`, which has `arg_register_powers[i][0] = 1` as constant and not contributing to the degree.

spec/src/field_vm.toml
RobinJadoul: I'm not sure; I initially wrote `not` there, but we'd used `!` in the past, so I went for consistency.

spec/chapters/field_vm.typ
RobinJadoul: I think state and instruction are in fact different concepts.
An instruction applies to a state, but doesn't define it, and it's the instruction that carries the information about hinting.

spec/chapters/field_vm.typ
RobinJadoul: On the notation, at first `MEM` is used for the semantics of the instructions. Later, the `[]`-only variant is only notation when encoding instructions.

spec/chapters/field_vm.typ
RobinJadoul: It's a common math thing. First you define the central thing, and then you flesh out the details, avoiding needing to give the details first without a good way to motivate them.
See e.g. the `where` statement in haskell for evidence on it being common enough to influence a (mathematical) programming language ;)

spec/chapters/field_vm.typ
RobinJadoul:
```suggestion
- $#`imm`_0 dot #`reg` + #`imm`_1$
- $#`MEM[`#`imm`_0 dot #`reg` + #`imm`_1#`]`$
```

spec/chapters/field_vm.typ
RobinJadoul: I believe I tried this, but it looked worse due to margin overruns

spec/chapters/field_vm.typ
RobinJadoul:
```suggestion
- $!#`hint_input`'_r and f_(r)(#`argument_registers`_0) = 0 => #`registers`'_r = #`registers`_r$
```

spec/chapters/field_vm.typ
RobinJadoul: It's not only the second constraint, it's the two together

```suggestion
Together, these constraints are logically equivalent to $!#`hint_input`'_r and not (#`hint_output` and f_(r)(#`argument_registers`_0) = 1) => #`registers`'_r = #`registers`_r$, but expressed in a way that polynomial constraints can more easily handle.
```

spec/expr.typ
RobinJadoul: Uhh, copy-paste problems, I'm assuming

spec/chapters/field_vm.typ
RobinJadoul: Program text = the program as a list of instructions (can be interpreted as DECODE table or whatever format is compiled to that, here); it's there to distinguish it from consecutive instructions in execution order

spec/chapters/field_vm.typ
RobinJadoul: While that's true for the case where `hint out = hint in = 1`, I'm not sure the `hint out = 0, hint in = 1` case is as clear-cut, since it changes the value of the out register, and hence can differ from what one expects when reading the first instruction in isolation.
Mentioning this here _could_ lead to less care during reviews/audits and still enable soundness bugs, or backdoors.

spec/src/field_vm.toml
```diff
@@ -0,0 +1,295 @@
...
+[[variables.input]]
+name = "mem_flags"
+type = ["Bit", 4]
+desc = "Whether the argument (`d`, `a`, `b`, `c`, in that order) comes from a memory lookup"
+pad = 0
+
+[[variables.input]]
+name = "hint_input"
+type = ["Bit", 5] # PARAM: ["Bit", N]
+desc = "Whether each input register is hinted this cycle"
```
erik-3milabs, line 37:
```suggestion
desc = "Whether each general purpose register is input hinted this cycle"
```
?

spec/chapters/field_vm.typ
erik-3milabs, line 1: consider adding

```typst
#show math.equation.where(block: false): box
```

as I find equations line breaking, which is rather annoying imo.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+a `ZERO` bit-register, the base field `PC` register and $N$ general purpose extension field registers.
+The number of registers was chosen as a tradeoff between the versatility of having more mutable state,
+and the extra cost in committed columns and decoding logic that grows with $N$.
+We index the registers from $0$ to $N + 1$ in the order above, so `ZERO` gets index $0$,
+`PC` gets index $1$ and then follow $N$ general purpose registers with indices $2...N+1$.
+
+The `ZERO` register indicates whether the previous instruction had a zero result,
+i.e. $next("ZERO") <=> #`d` = 0$.
+The `PC` register stores the program counter: the address of the current instruction,
+and --- except when branches are taken, through register hinting --- is incremented by one
```
erik-3milabs, line 62:
```suggestion
and --- except when branches are taken, through register hinting --- is incremented by $1$
```
?

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+Each argument to the `FMA` constraint has either of the two following forms:
+- $#`imm`_0 dot #`reg` + #`imm`_1$
+- $#`MEM[`#`imm`_0 dot #`reg` + #`imm`_1#`]`$
+where each immediate is a base field element, encoded in the instruction for a specific argument.
+The `d` argument to the instruction obtains its register value from the future state.
+
+== Register hinting
+
+Each general-purpose register in the current state can be marked as _hinted_ by the acting instruction.
+This means that from the current state onwards, the register can take a value
```
erik-3milabs, line 75: why are you choosing "onwards" here? It suggests that if for one instruction you set a register to hintable, and it remains that way for the rest of the execution trace. Perhaps worth clarifying that this is a way for the transition constraint not to hold for that particular instruction.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+Performing the hint only in the future state ensures that the original value of `reg` remains
+available throughout the computation.
+Additionally, output hinting allows for `PC` to be hinted,#footnote[
+  Note that we disallow this in input hinting, as it would allow for instructions
+  that can effectively hijack program execution.
+] enabling causal jumps and control flow in the program.
+
+In contrast, input hinting does not look like any traditional model of execution,
+instead allowing to update one or more values in the state, as long as the resulting state still satisfies
+the FMA constraint.
```
erik-3milabs, line 95:
```suggestion
the `FMA` constraint.
```

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+This can, e.g., be used to compute field inverses and square roots, which have a degree 2 constraint
```
erik-3milabs, line 96:
```suggestion
This can, e.g., be used to compute field inverses and square roots, which have a degree-2 constraint
```

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+
+#table(columns: (auto, 2fr, 1fr),
+       stroke: 0pt,
+       inset: (right: .5em),
+       table.header[*Pseudoinstr.*][*Translation*][*Comment*], table.hline(stroke: 1.5pt))[
+  `ADD d, a, b`][`FMA d == (0 * X + 1) * a + b, hint out`][Addition][
+  `MUL d, a, b`][`FMA d == a * b + (0 * X), hint out`][Multiplication][
+  `INV d, a`][`FMA d == (a + 1) * d - 1, hint <register of d>`][Extension field inversion. Note: `d` and `a` cannot use the same register here, and `d` should not be input-hinted in the next instruction.][
+  `J a`][`FMA PC == a, hint out`][Jump. Can be to a register, memory content, or absolute address, depending on the addressing mode of `a`, even relative to PC][
+  `JZA imm`][`FMA PC == (ZERO)*(-1*PC+(imm-1))+(1*PC+1), hint out`][Jump if ZERO, absolute target address][
```
erik-3milabs, line 161: A) shouldn't this -1*PC + imm jumps to imm-1? This way suggests that it is imm_0 * PC + imm_1 - imm_2.
B) what is the reason for leaving the open bracket of the first cell of each line on the previous row? Just looks odd.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+
+Eventually, usage may inform a set of common pseudoinstructions,
+along with informing potential optimizations that remove unused capabilities
+(e.g. reducing the number of immediates involved).
+
+== Calling convention<field-VM:sec:calling>
+
+Since the VM makes use of read-only memory, traditional usage of a program stack does not work.
+We assume that each function invocation (unless other optimizations apply) will have an associated _frame_,
+pointed to by a _frame pointer_ `fp`, one of the general purpose registers.
```
erik-3milabs, line 174: A) this comma is .... hard to read.
B) you're suggesting that a register _is_ the frame pointer. But isn't it technically that there is a register _containing_  the frame pointer? Might be worth pointing out that we label both the value and the register storing it the frame pointer.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+```
+CALL target:
+  FMA [fp] == fp, hint out
+  FMA [fp - i] == STORED_REG_i
+  FMA [fp + 1] == (PC + 2)
+  FMA PC == target, hint out
+  FMA fp == [fp], hint out
+
+RET:
+  FMA PC == [fp + 1], hint out
```
erik-3milabs, line 195: might be worth pointing out what each step does + what should be hinted at each step with a `hint`.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+We do this by constructing the Lagrange basis polynomials $f_(i)(x)$ such that $f_(i)(j) = 1$
+for $i = j in [0, N + 1]$ and $f_(i)(j) = 0$ for $i != j in [0, N + 1]$.#footnote[
+  Note that we allow ourselves to multiplex $N + 2$ registers here,
+  combining the $N$ general purpose registers, the PC and the `ZERO` register.
+]
+Since the degree of these $f_(i)(x)$ can grow too large to express in a single polynomial constraint,
+we perform a _"degree split"_:
+$ f_(i)(x) = f_(i, 0)(x) + x^(d - 1) (f_(i, 1)(x) + x^(d - 2) (f_(i, 2) + x^(d - 2) (f_(i, 3) + ...))), $
+for a maximal constraint degree $d$.
+Here, $deg f_(i, 0) <= d - 2$ and $deg f_(i, k) <= d - 3$.
```
erik-3milabs, line 230: Please consider

```suggestion
Here, $deg(f_(i, 0)) <= d - 2$ and $deg(f_(i, k)) <= d - 3$.
```
the `deg` is hard to distinguish from text without.

spec/chapters/field_vm.typ
erik-3milabs: >I wanted to keep the mathematical description clear, more than the exact way in which the constraints will use it. And expanding it would mean needing more complex exponents of the form x^((d - 1) + (k - 2) * (d - 2)), which felt less readable to me.

would you consider including both? e.g.,


```suggestion
$
f_(i)(x) 
&= f_(i, 0)(x) + x^(d - 1) (f_(i, 1)(x) + x^(d - 2) (f_(i, 2) + x^(d - 2) (f_(i, 3) + ...)))\ 
&= f_(i, 0)(x) + x^(d - 1) f_(i, 1)(x) + x^(2d - 3) f_(i, 2) + x^(3d - 5) f_(i, 3) + ...,\ 
$
```
or potentially with the reverse order. That way, the formula also neatly lines up with the `argument_registers[i]` powers listed next.

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+and then compute
+$
+  #`args_premem[i]` &= #`argument_scalars[i]` dot sum_(j = 0)^(N + 1) #`registers[j]` dot f_(j)(#`argument_registers[i]`)\
+                    &+ #`argument_offsets[i]`.
+$
+The coefficients for all $f_(i, k)$ are pre-computed once, based on the choices of $N$, $d$ and $t$,
+and used through the `MUX` constant columns.
+In this way, $f_(i, 0)$ can have degree at most $d - 2$, as it gets multiplied with $#`imm`_0$ and the register value,
+and the other $f_(i, k)$ can have degree at most $d - 3$, as they also get multiplied with the appropriate power of $x$.
+This leads to a total degree of $op("deg") f_(i) = d - 1 + t dot (d - 2) - 1$ for a maximal number of registers $N + 2 <= op("deg") f_(i) + 1$.
```
erik-3milabs, line 244: is there a reason you're not simplifying this to (t+1)(d-2)?

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+
+#render_constraint_table(chip, config, groups: "transition")
+
+Finally, to decode the instruction at the current PC, we would like to compress the information coming from
+the decoding table to reduce its number of columns.
+Doing so would require the elements being combined into one column to be range checked on this side
+of the interaction, ideally without needing any extra interactions or committed columns.
+For `Bit` variables, this is no problem with the `IS_BIT` template from @isbit.
+For `argument_registers` however, which should be in the range $[0, N + 1]$,
+which upon first attempt cannot be checked with a constraint of degree $<= d$.
```
erik-3milabs, line 289: this double "which" is confusing

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+Naturally, the `PC` and `ZERO` registers are exceptions since we need $next("pc") = #`pc` + 1$ if it is not (output-)hinted,
+and $next("ZERO")$ purely depends on $#`args`_0$ and not on `ZERO`.
+
+#render_constraint_table(chip, config, groups: "transition")
+
+Finally, to decode the instruction at the current PC, we would like to compress the information coming from
+the decoding table to reduce its number of columns.
+Doing so would require the elements being combined into one column to be range checked on this side
+of the interaction, ideally without needing any extra interactions or committed columns.
+For `Bit` variables, this is no problem with the `IS_BIT` template from @isbit.
```
erik-3milabs, line 287: Does 1 not have different square roots in the extension field other than itself?

erik-3milabs, line 287: does 1 not have another square root in the extension field?

spec/chapters/field_vm.typ
```diff
@@ -0,0 +1,332 @@
...
+In order to do so, we write the constant column `X` as the extension field element, such that $(1, #`X`, #`X`^2)$
+is the canonical basis of the extension field over the base field.
+
+#render_constraint_table(chip, config, groups: "decode")
+
+== Boundary constraints<field-VM:sec:boundary>
+
+Besides enforcing the FMA constraints and the correct transitions between states, we also need to ensure that execution
+starts at the correct instruction and ends with a halting instruction.
+This means that the verifier must check that the first row of the table corresponds to a state at `PC = 1` and all other variables set to $0$;
```
erik-3milabs, line 315: Shouldn't we introduce a "first-row" indicator row to make this a constraints? Otherwise, how is the verifier going to do this? The prover can only open elements in the codeword, but that is blown-up + shifted compared to the original table.

spec/chapters/field_vm.typ
RobinJadoul: Nope, as long as we're in a field, we only have ±1 as square roots of unity.
Stronger, the property that each non-zero polynomial of degree $d$ has at most $d$ roots remains valid.

spec/chapters/field_vm.typ
RobinJadoul: That just leads to needing a boundary condition on the "first-row" indicator column, since you can't enforce it cleanly otherwise.
In practice, you can do a boundary condition by using a single-row enforcement domain, so the quotient polynomial has a linear denominator for that constraint.

spec/chapters/field_vm.typ
RobinJadoul: The onwards is there to avoid making it seem like it's only for this state, i.e. that the transition constraints do carry it forward and it's not a temporary exception.

spec/chapters/field_vm.typ
RobinJadoul: A) I'm not exactly sure what you mean here, the notation is meant to imply that in the translation of JZA, the immediate used is modified and not copied as-is.

RobinJadoul: B) Because otherwise I can't keep everything in the table. If I were to write 

```
#table(columns: (auto, auto, auto))[a][b][c]
[d][e][f]
```

it would render wrong
image

spec/chapters/field_vm.typ
RobinJadoul: It just makes it a bit less clear where it comes from

spec/chapters/field_vm.typ
RobinJadoul: Added a footnote to hopefully clarify a bit. We don't really have "transition constraints" yet at this point of the expose.

spec/chapters/field_vm.typ
RobinJadoul: > But isn't it technically that there is a register containing the frame pointer

Not really, it's common enough to refer to either as frame pointer, and on pure technicality, it's probably more correct to only call the register the frame pointer, as a "pointer" is a type of variable.

spec/meta.typ
```diff
@@ -105,6 +105,7 @@
 #let common-formatting(body) = {
   set footnote(numbering: "[1]")
   show raw.where(block: true): it => block(it, inset: 1em, width: 100%, radius: 5pt)
+  show math.equation.where(block: false): box // Don't line-break inline equations by default
```
erik-3milabs: curious

spec/chapters/field_vm.typ
erik-3milabs: > A) I'm not exactly sure what you mean here, the notation is meant to imply that in the translation of JZA, the immediate used is modified and not copied as-is.

Let me rephrase.

Denoting `-1*PC+(imm-1)` in `JZA` suggests that this argument has the form `imm_0 * X + imm_1 - imm_2`, where `imm_0 = -1, X = PC` and `imm_2 = 1`; it does not cleanly fit the `imm_0 * X + imm_1` format

Same holds for `JNZA`, whose notation suggests that the `b`-argument has either the form
1. `imm_0 * X + imm_1 - imm_2`, with `imm_1 = 0`, or
2. `imm_0 * X + imm_1 * imm_2`, with `imm_1 = -1`.

B) Ok, I see. Why do you prefer this way of writing over, say,
```typst
#table(
    columns: (auto, auto, auto)
    [a][b][c]
    [d][e][f]
)
```
?

spec/chapters/field_vm.typ
erik-3milabs: What do you think of rephrasing 
1. "... pointed to by a frame pointer ..." to "... pointed to by the frame pointer ...", and
2. "we assume here that fp is one of the general purpose registers" to "we assume one of the general purpose registers to assume the role of fp"
?

(github doesn't support suggestions on outdated comments :grumble:)

spec/chapters/field_vm.typ
```diff
@@ -225,9 +228,10 @@ for $i = j in [0, N + 1]$ and $f_(i)(j) = 0$ for $i != j in [0, N + 1]$.#footnot
 ]
 Since the degree of these $f_(i)(x)$ can grow too large to express in a single polynomial constraint,
 we perform a _"degree split"_:
-$ f_(i)(x) = f_(i, 0)(x) + x^(d - 1) (f_(i, 1)(x) + x^(d - 2) (f_(i, 2) + x^(d - 2) (f_(i, 3) + ...))), $
+$ f_(i)(x) &= f_(i, 0)(x) + x^(d - 1) (f_(i, 1)(x) + x^(d - 2) (f_(i, 2) + x^(d - 2) (f_(i, 3) + ...)))\
+           &= 1 dot f_(i,0)(x) + x^(d - 1) f_(i,1)(x) + ... + x^(d - 1 + (t - 1) dot (d - 2)) f(i, t)(x), $
```
erik-3milabs, line 232:
```suggestion
           &= 1 dot f_(i,0)(x) + x^(d - 1) f_(i,1)(x) + ... + x^(d - 1 + (t - 1) dot (d - 2)) f_(i, t)(x), $
```

spec/chapters/field_vm.typ
erik-3milabs: > We don't really have "transition constraints" yet at this point of the expose. [...] The onwards is there to avoid making it seem like it's only for this state, i.e. that the transition constraints do carry it forward and it's not a temporary exception.

Should we then not start this section with outlining that, in principal, register values of contiguous states are the same (referring to the state machine description at the start of 35.1.1).
That furthermore allows us to explain why hinting should be included in the first place: to permit exceptions to this rule. And then clarify that indicating a "hint" just means severing this "continuity requirement" for one state transition.

wdyt?

spec/chapters/field_vm.typ
RobinJadoul: A) That's why I added the parens around `(imm - 1)`, maybe JNZA could use `PC + (-imm)` instead to make make that the same shape (or mess around with -1 factors everywhere, but that seems worse to me)
B) Because that also doesn't work without having commas between everything, and that looked worse to me

PR #986, opened 2026-09-11
Spec: unified MEMMOVE accelerator

nicole-graus:
## Description
Specifies the accelerator implemented in #984.

- New `MEMMOVE` chapter covering `memcpy`, `memmove`, `memset` and `commit`. The destination domain and the read/write timestamp order are decoded from the entry point, never chosen by the caller: `is_set` from the syscall number, `is_commit` from which bus the first row accepted from, both carried inside `MEMMOVE_NEXT` so a sequence cannot change functionality midway.
- `COMMIT` is now one row per ecall: it keeps the syscall number and the `x254` update, and defers the byte loop over `COMMIT_DEFER`. `CNB` is retired.
- Adds the `ADDNW` template, which is what rules out a sequence closing into a ring.
- Allocates ECALL `-30` (copy) and `-32` (memset), and lists `-31` as reserved for `HINT` so it is not handed out twice.


spec/src/commit.toml
erik-3milabs, line 1: The commit chip can be integrated in the MEMMOVE chip entirely: use the `is_commit` flag to select the right ecall code (feel free to update the commit ecall nr if required).

Also, note that memmove/memcpy/memset all return a void*, i.e., an address of sort. The same write statement can be used to return the proper return value for the commit chip.

spec/src/memmove.toml
```diff
@@ -0,0 +1,455 @@
...
+name = "tail"
+type = "Bit"
+desc = "Whether this row moves a single byte rather than eight"
+pad = 1
+
+[[variables.auxiliary]]
+name = "count_lt8"
+type = "Bit"
+desc = "Whether $#`count` < 8$, i.e. whether this row may not move eight bytes"
+pad = 1
```
erik-3milabs, line 72: consider merging these two variables into, e.g., "single".

Note that count < 8 => tail = 0 is equivalent to tail = 1 => count >= 8. This means that we can transform the link between them into

`LT[single; count, 8]` with multiplicity `mu-single`. 

This would require introducing a `single = 1 => mu = 1` constraint to ensure there is no negative multiplicity here.

spec/src/memmove.toml
```diff
@@ -0,0 +1,455 @@
...
+poly = ["*", ["not", "tail"], "count_lt8"]
+ref = "memmove:c:wide_needs_eight"
+
+[[constraints.width]]
+kind = "interaction"
+tag = "ALU"
+input = ["count", ["cast", 257, "DWordWL"], ["opsel", "LT"]]
+output = ["arr", 1, 0]
+multiplicity = "first_ecall"
+ref = "memmove:c:bound"
```
erik-3milabs, line 270: why are we restricting ourselves to count < 257, again?

In fact, why would we even need to range check count in the first place? Given that it is read from memory, we can assume it is a properly formatted value, and we use the SUB chip to decrement it appropriately, which also requires the decremented value to be range checked.

Recommendation: drop the constraint all together.

spec/src/memmove.toml
```diff
@@ -0,0 +1,455 @@
...
+multiplicity = ["idx", "commit_lane", "i"]
+ref = "memmove:c:commit_value_out"
+
+[[constraints.commit]]
+kind = "interaction"
+tag = "memory"
+input = [2, ["arr", ["+", ["idx", "dst", 0], "i"], ["idx", "dst", 1]], 1, ["idx", "value", "i"]]
+iter = ["i", 0, 7]
+multiplicity = ["-", ["idx", "commit_lane", "i"]]
+ref = "memmove:c:commit_value_in"
```
erik-3milabs, line 326: what is the reason you haven't used MEMW here? Doing so should reduce this chip considerably, as we reduce 16 interactions to 2.

spec/src/memmove.toml
```diff
@@ -0,0 +1,455 @@
...
+cond = "first_ecall"
+ref = "memmove:c:read_src"
+
+[[constraints.read_input]]
+kind = "template"
+tag = "REG"
+input = [12, "count", "timestamp"]
+output = "count"
+cond = "first_ecall"
+ref = "memmove:c:read_count"
```
erik-3milabs, line 245: what definition of `memset` are you using here? The one I found doesn't have a src, yet here we are reading it from 11.

spec/src/memmove.toml
```diff
@@ -0,0 +1,455 @@
...
+kind = "arith"
+constraint = "$#`is_set` => #`dst`_0 = #`src`_0 + 8$"
+poly = ["*", "is_set", ["-", ["idx", "dst", 0], ["idx", "src", 0], 8]]
+ref = "memmove:c:set_gap_lo"
+
+[[constraints.copy]]
+kind = "arith"
+constraint = "$#`is_set` => #`dst`_1 = #`src`_1$"
+poly = ["*", "is_set", ["-", ["idx", "dst", 1], ["idx", "src", 1]]]
+ref = "memmove:c:set_gap_hi"
```
erik-3milabs, line 307: when `is_set=1`, you still read both `src` and `dst` from memory. This constraint will probably give trouble there.

spec/src/memmove.toml
```diff
@@ -0,0 +1,455 @@
...
+[[constraint_groups]]
+name = "incr_decr"
+
+[[constraints.incr_decr]]
+kind = "template"
+tag = "ADDNW"
+input = ["src", ["arr", "step", 0]]
+output = ["cast", "src_incr", "DWordWL"]
+cond = ["-", "μ", "end"]
+ref = "memmove:c:src_incr"
```
erik-3milabs, line 337: why use ADDNW here, rather than just ADD? I don't know of any specification that does not require this value to wrap. Even still, this VM has a guard page at page_idx=0, so if you write there, the proof won't complete anyway.

Also, why include the mu-end =>  ? I don't think that is actually necessary? Same holds for dst_incr

spec/src/memmove.toml
```diff
@@ -0,0 +1,455 @@
...
+multiplicity = "μ"
+ref = "memmove:c:range_src_incr"
+
+[[constraints.incr_decr]]
+kind = "template"
+tag = "ADDNW"
+input = ["dst", ["arr", "step", 0]]
+output = ["cast", "dst_incr", "DWordWL"]
+cond = ["-", "μ", "end"]
+ref = "memmove:c:dst_incr"
```
erik-3milabs, line 353: same comment on using ADDNW to src applies here.

spec/src/memmove.toml
erik-3milabs, line 337: Note that by inlining the ADD chip, we can even reconstruct `src_incr[3]` from `src_incr[2]`, `src[1]` and `carry[0]` since `carry[1] = 0`. This saves a column.

Same applies to `dst_incr`.

spec/chapters/memmove.typ
```diff
@@ -0,0 +1,173 @@
...
+
+The #memmove chip is comprised of #nr_variables variables that are expressed using #nr_columns columns and leverages #nr_interactions interaction(s):
+#render_chip_variable_table(chip, config)
+
+= Assumptions
+#render_chip_assumptions(chip, config)
+
+These concern the _first_ row of a sequence, where the values come from the register file or from `COMMIT`; every later row receives them over `MEMMOVE_NEXT`, where @memmove:c:range_src_incr, @memmove:c:range_dst_incr and @memmove:c:range_count_decr range-check three of the four on the sending side.
+`timestamp` is range-checked by neither side and holds only because it travels unchanged from the `ECALL` at the root.
+@memmove:a:dst is not discharged at all on a commitment sequence (@memmove:aside:index).
```
erik-3milabs, line 45: what does it mean to "discharge" an assumption?

spec/chapters/memmove.typ
```diff
@@ -0,0 +1,173 @@
...
+
+= Constraints
+In this VM, we assign syscall number -30 to the copy functionality of the #memmove accelerator, and -32 to `memset`.
+Since the number of bytes is not known in advance, this chip is recursive: each row moves one chunk and "calls" itself to move the remainder, so only the `first` row of a sequence accepts an entry.
+There are two entries --- an `ECALL` from the `CPU`, or the byte loop `COMMIT` defers (@commit) --- and `first_ecall` and `first_commit` split `first` between them.
+#render_constraint_table(chip, config, groups: "incoming")
+
+== Selecting the functionality
+@memmove:c:receive_ecall receives the system call number as $2^32 - 30 - 2 dot #`is_set`$, so `is_set` is decoded from the `ECALL` the guest executed rather than chosen.
+Note that the low limb of that tuple is a line in `is_set` and so reaches every system call number in the negative range: @memmove:c:range_is_set is what excludes them, and it therefore carries the whole decoding argument.
```
erik-3milabs, line 55: > ... is a line in ...

what does that mean?

spec/chapters/memmove.typ
```diff
@@ -0,0 +1,173 @@
...
+== Reading the operands
+The guest-side `memcpy` this chip accelerates has the following signature:
+
+```c
+void *memcpy(size_t count; void dest[restrict count], const void src[restrict count], size_t count);
+```
+
+That is to say, `A0` contains the address of the first byte to write, `A1` the address of the first byte to read, and `A2` the number of bytes to move; `memset` uses the same three registers for the same three roles.
+Each read writes back the value it read, so the operation leaves the registers untouched and the guest produces the return value.
+These are conditioned on `first_ecall`, since a deferred commitment sequence takes its operands from `COMMIT`.
```
erik-3milabs, line 71:
```suggestion
These are conditioned on `first_ecall`, since a deferred commitment sequence takes its operands from `COMMIT_DEFER`.
```
?

spec/chapters/memmove.typ
```diff
@@ -0,0 +1,173 @@
...
+= Padding
+To pad this chip, use the below data.
+#render_chip_padding_table(chip, config)
+
+This padding row is not all-zero: @memmove:c:count_decr is unconditional, so $#`tail` = 1$ makes $#`step` = 1$, which $#`count` = 1$ and $#`count_decr` = 0$ satisfy, and $#`tail` = 1$ also satisfies @memmove:c:wide_needs_eight.
+The low-limb carry of the two position updates is constrained on every row (@addnw:c:carry), which $#`src_incr` = #`dst_incr` = 1$ satisfies with a zero carry.
+
+= The Accelerated Memory Operations standard
+The Ethereum Foundation's Accelerated Memory Operations standard fixes what an accelerated `memcpy`, `memmove` and `memset` must provide.
+#footnote([Accelerated Memory Operations; eth-act/zkevm-standards, commit `e6a4cc0`. #link("https://github.com/eth-act/zkevm-standards/tree/e6a4cc0/standards/accelerated-memory-operations")[[src]]])
```
erik-3milabs, line 162: link is broken.

erik-3milabs (changes requested):
First set of comments. Primary change request: the COMMIT chip should be completely integrated into the MEMMOVE chip; no COMMIT chip should be needed after this PR is done.

Once this is done, I'd be happy to give this another review!

spec/src/commit.toml
nicole-graus: Yes, we can do that. We kept it because Robin suggested keeping it for the ecall number and the x254 update. We can integrate it, the total number of columns will shrink but the cost per-row of MEMMOVE goes up (COMMITS interactions will be paid by all the rows of MEMMOVE).

spec/src/memmove.toml
nicole-graus: We want to avoid one ecall to add a lot of rows and make an epoch unprovable. We kept that restriction from #874. Do you say it isn't a problem to have many rows for one cycle?

spec/src/memmove.toml
nicole-graus: you're right, I will change it.

spec/src/memmove.toml
nicole-graus: I didn't understand.
When `is_set=1`, the memset stub set the registers x10 and x11. `src` and `dst` are the values in registers x10 and x11, and those values  satisfy `dst = src + 8`, what do you say is the problem?

spec/src/memmove.toml
nicole-graus: Yes, I agree. src is where the stub writes the bytes, since the accelerator doesn't fill just copies those bytes. But since it's always `dst − 8` we can derive it instead of passing it, and then that constraint becomes a definition rather than a check.

spec/src/memmove.toml
nicole-graus: You're right. I will implement it

spec/src/memmove.toml
nicole-graus: We use ADDNW to make `src` strictly increase, so the chain can't close in a loop that balances while consuming no ecall.

I don't see a guard at page 0, address 0 is writable like any other address: `store_byte(0, x)` just inserts into the map, and a later `load_byte(0)` reads `x` back.

About mu-end, you're right. The padding rows already satisfy the constraint, so we can drop it.

spec/src/memmove.toml
nicole-graus: For `dst` it's a different reason than `src`. The executor rejects a copy whose range overflows (`dst.checked_add(n)`), so plain ADD would let the AIR accept a copy the VM refuses.

spec/src/memmove.toml
nicole-graus: True, I'll change it.

spec/chapters/memmove.typ
nicole-graus: We meant that an assumption is discharged when you can point to something that actually guarantees it. In that line we wanted to say that nothing guarantees it.
I agree it is confusing, I will rephrase it.

spec/chapters/memmove.typ
nicole-graus: A linear function in `is_set`. I will reword it

spec/src/memmove.toml
nicole-graus: Fixed it here [eb8c87c](https://github.com/yetanotherco/lambda_vm/pull/986/commits/eb8c87cf7736d9e1fca35ddd43b228fd30a4abc5)

spec/src/memmove.toml
nicole-graus: Fixed it here [c96a174](https://github.com/yetanotherco/lambda_vm/pull/986/commits/c96a17453123869a4b37998dc6c8d17a45b669a3).

spec/src/memmove.toml
nicole-graus: I was changing it in the code and realized that if we don't include `src` and we don't read it from x11, then the multiplicity should be `f_ncommit * (1 − is_set)` with a degree 2, since that read already has multiplicity `f_ncommit` and we would need to switch it off for memset. And a degree 2 would need an other column.

spec/src/memmove.toml
nicole-graus: Changed it in [e43cad6](https://github.com/yetanotherco/lambda_vm/pull/986/commits/e43cad61aed230740bb857e087ef4968e328823f). Note that `src_incr[3]` is used in the address we send on MEMMOVE_NEXT bus, so now we have to derive it there, making it a bit messier.

spec/chapters/memmove.typ
nicole-graus: Done here [c2bf432](https://github.com/yetanotherco/lambda_vm/pull/986/commits/c2bf432ba796314279734e8137179aec8ce234a1).

spec/chapters/memmove.typ
nicole-graus: Done here [013d433](https://github.com/yetanotherco/lambda_vm/pull/986/commits/013d43375bb16a842f0d6c2966b118c108444f95).

spec/chapters/memmove.typ
nicole-graus: Fixed in [5ce3ef4](https://github.com/yetanotherco/lambda_vm/pull/986/commits/5ce3ef4974eba10d776833d2b622e68e423b59fb).

PR #988, opened 2026-09-14
Multilinear whir

ColoCarletti:


MauroToscano:
Heads-up on a draft stacked on this branch: **#992**, `whir/rpx` → `multilinear-whir`.

It makes the WHIR path's hash a parameter and adds an RPX256 instantiation. **Keccak stays the default and
is byte-identical to this branch's head** — same proof, same serialized length, same `num_queries`; the
byte gate in it compares a fixed proof's digest at `307d7c00` and at every commit of the stack and they
match. `Commitment` stays `[u8; 32]`, so no proof struct changes shape and there is no format bump.

The reason for it is recursion rather than performance, and the PR says so plainly: RPX is slower than
keccak everywhere a host or a GPU does the hashing (measured on block 25368371: prove 39.9 s → 78.1 s), and
it is carried because a keccak-f costs ~73,700 trace cells in a field-native verifier against RPX's 325.

**Three commits in it are repairs to things that were already red on this branch, and they are separately
droppable — you may want them regardless of what happens to the rest:**

- `f57a71a5` — four call sites in `multilinear_bench_tests.rs` pass the wrong arity to
  `Traces::from_elf_and_logs` under `disk-spill`, so `make lint`'s third pass cannot build. Verified red at
  this branch's head before anything was stacked on it.
- `140e2694` — four clippy findings in `crypto/multilinear/src/gpu.rs` (a test module mid-file, an unused
  type parameter on `columns_at`, a complex return tuple, and a never-read RAII field where the lint is
  wrong and wants an `allow` with a reason). Likewise red at this branch's head. With these, `make lint` is
  five passes and all five are green — the `cuda` pass for the first time.
- `c9c452c6` — `hash-metrics` was in no lint pass and no test target, which is how its Merkle counters
  stayed keccak-only.

Also worth knowing while you review: `-p multilinear` is absent from `.github/workflows/pr_main.yaml` (the
test archive at line 349 lists `lambda-vm-prover`, `stark`, `crypto`, `ecsm`), so the 275 tests under
`crypto/multilinear/src` do not run in CI today. Orthogonal to both PRs, but it surprised us.

Happy to split, rebase, or drop any part of it — it is a draft and it is stacked on your work, so it moves
when you do.

PR #1002, opened 2026-09-21
fix(lfm,tests): the eleven pins the main-sync port left behind, each against its mover

MauroToscano:
## What this fixes

The first GitHub CI run ever made against a `whir/*` base (PR #999) came back with four red prover tests. They were not #999's: all four fail identically at this base, `c00342c1f`. A plain-libtest pass of the whole `lambda-vm-prover` lib suite on the FAST box then showed the true red set on the lineage tip is **eleven** tests (sharded nextest cancels a shard on its first failure, so a sharded run reports a lower bound, never a count).

Every one of them traces to one event: the main-sync merge `892c7d1bc` brought #977 (skip empty tables) onto the lineage, taking `FIXED_TABLE_COUNT` from 11 to 5. That has two arms:

- **the table set shrank** — only BITWISE, DECODE, KECCAK_RC, REGISTER, HALT are unconditional now; the six accelerator families are counted and elided when empty;
- **the epoch statement grew 49 bytes** — the six counts joined the absorbed statement and `is_final` became its last byte (215 → 264 B), one extra BLAKE3 compression.

The port updated everything that fails to *compile* and missed every pin that fails only at *runtime*.

## The commits

| commit | what |
|---|---|
| `8cedf2e25` | the re-bless: `STATEMENT_REPLAY_BLAKE3_ROWS` 9 → 10; `SUB_PROOFS` 25 → 16 and `CHALLENGES_AT_MIN_PRESET` 115 → 79 with the identity respelled both-sides-positive (`SUB_PROOFS - 24` on a `usize` now underflows instead of failing); the statement byte length 215 → 264; the constraint-leg census split into `always_on` (5) and `accelerators` (6); the registry regenerated — exactly the `StatementReplayV0` entry moved (roots 0,1,3,4,7,9,11 and `program_id`), the generated `roots[11]` equals byte-for-byte what the failing test computes; two docs and a rider corrected. Each re-blessed literal names `892c7d1bc` and its arm in the assertion. |
| `788f36a19` | the table-set gates print the tables they argued, by name |
| `865c689b0` | rustfmt fixup |
| `413b3a3b7` | the two table-set censuses pinned as ordered lists; `ALWAYS_ON_TABLES` shared and pinned against `FIXED_TABLE_COUNT` (a third independent confirmation of the cause) |
| `3218346c7` | three more the port left behind (see findings 1–3 below), and a precompile test that proved the wrong guest |
| `c8c7c036e` | the tenant socket census measures before it asserts, and names its hasher |
| `05f0500a2` | the socket assertion tied a counterfactual tenant to the live pin: replaced by the two checks it was reaching for (production is the algebraic tenant; an algebraic tenant exists at `BLOCK_HASHER`); no number moves |

**The honest headline:** of the eleven, exactly one moved a blessed byte, the registry regeneration, where only `StatementReplayV0` moved and its new `roots[11]` equals what the failing test computes. Everything else was a test describing the machine wrongly. The port did not break the prover; it broke the tests' account of it.

## Findings surfaced on the way (not fixed here where they are out of scope)

1. **`a_program_using_a_precompile_is_proved_and_verified` proved a software-keccak guest.** `keccak.elf` depends on `tiny-keccak`; its only syscall is `commit`. The accelerator guest `keccak_precompile.elf` is built by the Makefile wildcard and was proved nowhere on the multilinear path. The test now loads the accelerator guest and asserts KECCAK, KECCAK_RND and COMMIT are argued; the software guest keeps its own case, pinned as the suite's widest table set (21 tables, the only one reaching MEMW, LOAD, STORE, BRANCH, EQ).
2. **`all_instructions_64` argues 15 tables** and reaches no branch, load, store, `eq`, unaligned MEMW or COMMIT. Pinned as measured; widening the fixture lives in `executor/programs/asm`.
3. **`SOUNDNESS.md` claimed `T = 24` was "measured, not assumed"**, citing a test that asserts no count. The bound survives (increasing in T, T went down); the doc now says so.
4. **The LFM_HASH socket census** pinned a counterfactual tenant's socket to the current build's pin. Measured widths, by socket: Test 28, Rpx 316, Rpo 436, Poseidon 612 main columns (3 ext aux). Production is the algebraic tenant; the BLAKE3 tenant is lever 0's comparison arm and its 28-column socket is that build's own record, so `RECORDED` and the 0.774× ratio stand. The alternative modelling fork (`BLAKE3_TENANT_SOCKET = BLOCK_HASHER`, `RECORDED = (316, 3)`, ratio ≈0.757× by interpolation) is documented in the test for whoever wants to take it.
5. **`no_call_site_outside_the_pin_reaches_a_default_alias`** flagged `lfm/algebraic_commit.rs` for a test naming `GrindingDigest<RpxStarkHash>` and a deliberate BLAKE3 cross-hash control; `GrindingDigest` is allowed as hash-agnostic, the file is blessed for the concrete tag only.

## Gates

`make fmt` and `make lint` (ten steps, three under `cuda`) green at every commit, on the laptop. Full `lambda-vm-prover` lib suite, plain libtest, on the FAST box at `3218346c7`: 1451 passed, 82 ignored, 1 failed (the census test); that test re-run at `05f0500a2` on the box: green, 0.06 s, panel printed. This PR's CI shards are the full pass at the tip. Registry regeneration, the eight argued-table censuses and the socket-width panel were all measured on the box, CPU-only, with no campaign environment set.


MauroToscano:
Cherry-picked `90a38eccd` from #999 (the host KAT result-type pun under LP64) on top, as `b420ee950`: the Host KAT job is red at the lineage base itself and #999 carries the only fix, so without it the two PRs would block each other. #999 will see the commit as already applied. No other change.

MauroToscano:
Full `lambda-vm-prover` lib suite at `05f0500a2` on the FAST box, plain libtest (no fail-fast), CPU-only, no campaign environment: **1452 passed, 0 failed, 82 ignored** (2553 s). The first fully green pass of this suite on the lineage.

